PR #6, opened 2022-09-30
feat: proof debugger

paolodamico:
Introduces the proof debugger based on recent experiences at hackathons. This will help devs (particularly building on-chain) make sure the proofs are valid and detect common issues.

Closes https://linear.app/worldcoin/issue/WID-114/proof-and-encoding-tester

### DIFF
--- a/web/urls.ts
+++ b/web/urls.ts
@@ -21,5 +21,5 @@ export const urls = {
   kiosk: (action_id: string): `/kiosk/${string}` => `/kiosk/${action_id}`,
   team: (): "/team" => "/team",
 
-  debugger: () => "/debugger",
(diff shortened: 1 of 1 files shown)


web/urls.ts
```diff
@@ -20,4 +20,6 @@ export const urls = {
 
   kiosk: (action_id: string): `/kiosk/${string}` => `/kiosk/${action_id}`,
   team: (): "/team" => "/team",
+
+  debugger: () => "/debugger",
```
igorosip0v, line 24: @paolodamico One small thing, can we add explicit return type here? Improves the displaying in IDE with typescript: `() => "/debugger"` instead of `() => string`

igorosip0v (approved):

PR #24, opened 2022-10-11
refactor: update codeblocks styles on depoyment instructions

michaelfurutin:
image


igorosip0v (commented):
Seems good to me.

But one small detail -- is it possible to change color of code highlight?

PR #80, opened 2023-02-20
fix: updating external nullifier when relevant

paolodamico:


hasura/migrations/default/1676583921722_create_table_public_action/up.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE "public"."action" (
   "app_id" varchar(50) NOT NULL,
   "max_accounts_per_user" integer NOT NULL DEFAULT 1,
   "max_verifications" integer NOT NULL DEFAULT 1,
+  "creation_mode" varchar(50) NOT NULL DEFAULT 'developer_portal',
```
igorosip0v, line 12: @paolodamico  Maybe we should add some comment for this field with the union of values it should contain?

igorosip0v (approved):

PR #97, opened 2023-02-23
feat: OIDC I

paolodamico:
- Implements the main OIDC method that receives a ZKP from IDKit, verifies it and issues either an auth code (authorization code flow) or a signed JWT (implicit flow, non-OAuth flow).
- Adds auth code model to store the temporary codes.
- Introduces models to track types for models, and cleans up some legacy types.

### DIFF
--- a/web/pages/api/v1/oidc/authorize.ts
+++ b/web/pages/api/v1/oidc/authorize.ts
@@ -29,7 +29,6 @@ export default async function handler(
     "proof", // ZKP param
     "nullifier_hash", // ZKP param
     "merkle_root", // ZKP param
-    "nonce", // `signal` in ZKP world
     "credential_type",
     "app_id",
   ]) {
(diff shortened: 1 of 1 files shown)


web/pages/api/v1/oidc/auth.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (!req.body[attr]) {
+      return errorRequiredAttribute(attr, res);
+    }
+  }
+
+  const {
+    proof,
+    nullifier_hash,
+    merkle_root,
+    nonce,
```
maxpetretta, line 45: `nonce` is only required in implicit flow

web/pages/api/v1/oidc/auth.ts
```diff
@@ -0,0 +1,129 @@
+import {
```
maxpetretta, line 1: To be friendly w/ spec, let's rename to `authorize.ts`

maxpetretta (approved):

PR #103, opened 2023-02-23
feat: add return_url to action table

igorosip0v:
WIP

### DIFF
--- a/hasura/migrations/default/1677178918422_alter_table_public_action_add_column_return_url/up.sql
+++ b/hasura/migrations/default/1677178918422_alter_table_public_action_add_column_return_url/up.sql
@@ -1,0 +1,2 @@
+alter table "public"."action" add column "return_url" jsonb
+ not null default '[]';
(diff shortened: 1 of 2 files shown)


hasura/migrations/default/1677178918422_alter_table_public_action_add_column_return_url/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."action" add column "return_url" jsonb
```
maxpetretta, line 1: Can rename `redirect_uri` to match OIDC spec

maxpetretta (approved):

hasura/migrations/default/1677178918422_alter_table_public_action_add_column_return_url/up.sql
igorosip0v: @paolodamico @maxpetretta 
I've added a table `redirect`, it has `uri` JSONB field and `action_id` field which is connected to `action.id` as an object relation. `action_id` field is unique, so that the one action has one redirect row.

```json
{
  "data": {
    "action": [
      {
        "id": "action_2031281ef344cff39155ec85aeeb42f1",
        "redirect": {
          "id": "redirect_612bf3a0b66a0731db5ca4c4c590ea92",
          "uri": []
        }
      }
    ]
  }
}
```

maxpetretta:
Wait to include this w/ redirect uri validation

PR #107, opened 2023-02-24
feat: token endpoint

paolodamico:


web/pages/api/v1/oidc/token.ts
```diff
@@ -1,2 +1,135 @@
...
+    app_id,
+    nullifier_hash: code.nullifier_hash,
+    credential_type: code.credential_type,
+    ...jwk,
+  });
+
+  return res.status(200).json({
+    access_token: token,
+    token_type: "Bearer",
+    expires_in: 3600,
```
maxpetretta, line 131: These three lines are only required for the OAuth spec, it's unclear whether we need them, since we are not provisioning access to anything

maxpetretta (approved):

PR #157, opened 2023-03-04
fix: fixes all API tests

paolodamico:
PR also fixes a few edge cases on endpoints caught by tests.


igorosip0v (approved):
Tried locally, and everything seems good.

PR #163, opened 2023-03-07
feat: app profile saving

michaelfurutin:


web/src/scenes/app/Configuration.tsx
```diff
@@ -35,10 +72,23 @@ export const Configuration = memo(function Configuration() {
   const { updateAppName, updateAppDescription } = useApps();
   const [appName, setAppName] = useState<string>(initialValues.name);
   const [debouncedAppName] = useDebounce(appName, 1000);
```
igorosip0v, line 74: @michaelfurutin Let's remove debounced inputs and related logic which saves data on input. Just leave saving onBlur and saving using Enter button.

igorosip0v (changes requested):

web/src/hooks/useApps.ts
```diff
@@ -149,7 +149,7 @@ const useApps = () => {
         return;
       }
 
-      updateApp({
+      return updateApp({
```
igorosip0v, line 152: @michaelfurutin It will be better to update fetcher instead. Let's update this so that it can be used as
```typescript
const {trigger: toggleAppActivity} = useSWRMutation("app", updateAppFetcher, {...})
toggleAppActivity(true | false)

const {trigger: setAppName} = useSWRMutation("app", updateAppFetcher, {...})
setAppName(<name>: string)

```

PR #164, opened 2023-03-07
feat: add not found app screen

michaelfurutin:


web/src/scenes/app/index.tsx
```diff
@@ -20,12 +20,20 @@ export const App = memo(function App(props: { appId: string }) {
...
       <Layout>
-        {(isLoading || !currentApp) && (
+        {isLoading && (
           <div className="w-full h-full flex justify-center items-center">
             <Preloader className="w-20 h-20" />
           </div>
         )}
 
+        {!isLoading && !currentApp && (
+          <div className="grid">
```
igorosip0v, line 30: @michaelfurutin Let's make this centered.

igorosip0v (changes requested):

igorosip0v:
@michaelfurutin Also, let's make an apps selector in the `Layout` possible to open, so users can choose from the existing apps in the list.

web/src/scenes/app/index.tsx
michaelfurutin: done

PR #173, opened 2023-03-10
feat: connect team page to hasura

michaelfurutin:
https://user-images.githubusercontent.com/104859710/224374906-998bf1ce-8b83-4209-804c-fd71c4b642af.mp4

### DIFF
--- a/hasura/metadata/databases/default/tables/public_user.yaml
+++ b/hasura/metadata/databases/default/tables/public_user.yaml
@@ -34,7 +34,9 @@ select_permissions:
         - email
         - id
         - is_subscribed
+        - name
         - team_id
+        - world_id_nullifier
       filter:
         team:
           users:
--- a/web/src/scenes/team/MemberList/index.tsx
+++ b/web/src/scenes/team/MemberList/index.tsx
@@ -1,46 +1,35 @@
 import { useToggle } from "@/hooks/useToggle";
 import { Icon } from "@/components/Icon";
-import { memo, useCallback, useEffect } from "react";
-import { TeamMember, useTeamStore } from "../../../stores/teamStore";
-
+import { memo, useMemo, useState } from "react";
 import { Controls } from "./Controls";
 import { InviteMembersDialog } from "./InviteMembersDialog";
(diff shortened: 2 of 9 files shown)


web/src/scenes/team/MemberList/index.tsx
```diff
@@ -67,21 +59,22 @@ export const MemberList = memo(function MemberList() {
...
                 <Icon name="user-solid" className="w-4 h-4 bg-success" />
 
-                {member.verified && (
-                  <Icon
-                    name="badge-verification"
-                    className="absolute w-5 h-5 -bottom-1 -right-1"
-                    noMask
-                  />
-                )}
+                {/* FIXME: add verified flag to hasura */}
```
igorosip0v, line 70: @paolodamico we have is_subscribed field in user table. Is this field defines is user verified or not?

hasura/metadata/databases/default/tables/public_user.yaml
```diff
@@ -34,7 +34,9 @@ select_permissions:
         - email
         - id
         - is_subscribed
+        - name
         - team_id
+        - world_id_nullifier
```
igorosip0v, line 39: @michaelfurutin Why do we need to let user pick this field?

web/src/stores/teamStore.ts
```diff
@@ -1,3 +1,4 @@
+// FIXME: we don't need this store anymore, we can use useSWR directly
```
igorosip0v, line 1: @michaelfurutin remove it in case we don't use it anymore

igorosip0v (changes requested):

hasura/metadata/databases/default/tables/public_user.yaml
michaelfurutin: Maybe it is the presence of this hash that means the user is verified, but I am not sure.

igorosip0v: Ah, yeah Pretty likely 😅
@paolodamico Can you confirm this?

PR #185, opened 2023-03-13
fix: new action improves

denis-bushaev:
### Closes wid-380

- added filter for an action input value
- removed `app_id` input, and replaced it with `currentApp.id`
- rewording action, changed label font, added placeholder
- added generate `external_nullifier` on create


igorosip0v (changes requested):
@denis-ottofeller In case I've just created new action with action identifier "test" and then trying to do the same, nextjs throws me an error about duplicating external_nullifier. Can we catch this and highlight input red with "this action identifier already exists" text?

PR #190, opened 2023-03-13
feat: add redirect from actions on app not found

denis-ottofeller:
#### Closes wid-387

<details><summary>Result</summary>

https://user-images.githubusercontent.com/102026658/224775787-d09a2819-65db-4dbc-bf00-a998570b5759.mov
</details>


igorosip0v (commented):
@denis-ottofeller You redirecting to the default app. I guess we need to redirect the user to the non-existing app page, so the user can see that this app is not found.
@paolodamico what do you think?

igorosip0v (approved):

PR #285, opened 2023-04-25
chore: update README with dev instructions

maxpetretta:
Updates README with instructions for KMS, dev login, and more


0xPenryn:
i get the error `--detach cannot be combined with --abort-on-container-exit, --attach or --attach-dependencies` when running the docker compose command

PR #286, opened 2023-04-27
fix: improve key hooks stores

michaelfurutin:
* сorrected the value of the mutation key so that revalidation works
* changed the secret storage logic, now the displayed secrets are separated and stored by key id


web/src/scenes/team/Credentials/index.tsx
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck FIXME: temporary disable typescript checking
```
igorosip0v, line 1: Seems like we don't use this component. @michaelfurutin Can you confirm? 

@maxpetretta Should we just remove it?

michaelfurutin: yes, this component is not used

maxpetretta: Will remove on my end

PR #316, opened 2023-06-30
fix: pass missing invite id

igorosip0v:
- Add storing email invite id to the cookies.
  1. User clicks "join" in the email
  1. When the user redirects from email to `/login-with-invite?invite=VALUE`. The value will be stored as `invite_id` cookie from server side.
  2. login with the world id URL will be calculated on the server side and then passed to the component. 
  3. If we have loginUrl we redirect the user to this URL from the client.
  4. User scanning QR code. 
  5. Then the user redirected to the `redirect_uri` which is `/login-with-email` but with `client_id` query param.
  6. We'll try to get the `invite_id` cookie and pass it to the `/api/login-with-email` along with `client_id`
     - Problem was that we do not pass the `invite_id` on this step.
     

REVIEW: 

- Not sure how to get ironclad_id when we login user from the email invite.


igorosip0v:
Closing because this changes was merged in #320

PR #340, opened 2023-07-21
chore: replace all console.logs with logger

denis-ottofeller:
This PR replaces all SSR console logs with dd logger.


denis-ottofeller:
@paolodamico i have been replace all console.{error|log|warn|info} on the ssr.

igorosip0v (approved):

PR #344, opened 2023-07-22
fix: unique constraint on email

paolodamico:
Previously we were inserting blank emails as empty strings which was causing errors with the uniqueness constraint. This fixes it.


lukejmann (approved):
LGTM

PR #345, opened 2023-07-22
fix: clarify login/signup

paolodamico:


m1guelpf (approved):

web/src/scenes/login/login.tsx
```diff
@@ -183,7 +183,7 @@ export function Login({ loginUrl }: ILoginPageProps) {
 
                   <Link href={loginUrl ?? ""} className="contents">
                     <Button className="flex flex-1 justify-between px-6 py-4 text-16 font-semibold">
-                      Log in
+                      Log in or Sign up
```
m1guelpf, line 186: Thoughts on renaming this to sign in with Worldcoin?

PR #357, opened 2023-08-07
fix: properly escape backslash in SQL string

0xPenryn:


hasura/migrations/default/1678060027000_create_table_public_redirect/up.sql
```diff
@@ -26,11 +26,11 @@ BEGIN
   IF url IS NOT NULL AND url != '' THEN
     
     IF strpos(url, 'http://localhost') = 1 OR strpos(url, 'https://localhost') = 1 THEN
-      IF NOT (url ~* '^https?://localhost(:[0-9]+)?(/[^\s?]*)(\\?[^\s]*)?$') THEN
+      IF NOT (url ~* '^https?://localhost(:[0-9]+)?(/[^\\s?]*)(\\?[^\\s]*)?$') THEN
```
igorosip0v, line 29: @0xPenryn I don't think adjusting the existing migration is a good idea. I think you need to copy this function create or replace query with adjusted regexps and execute it in local hasura console with "new migration" checkbox to create new migration.

PR #361, opened 2023-08-09
feat: PKCE

m1guelpf:


web/src/pages/api/v1/oidc/token.ts
0xPenryn, line 1: we'll need to enable CORS on the token endpoint when using PKCE

m1guelpf: The token endpoint is not intended to ever be called from the front-end

0xPenryn: PKCE is specifically intended for scenarios where you're using a public client (i.e. a native app or SPA), rather than a confidential client (like a backend that can securely store the client_secret).

if you had a backend that you could use to call the token endpoint, you could just use the standard auth code authorization (client id and secret) instead

PR #365, opened 2023-08-11
refactor: add yup validation to api routes

michaelfurutin:
Added Yup validation to api routes


igorosip0v:
@paolodamico `schema.validate` is converting type if field defined as `yup.string()`, for example, `1` becomes `"1"` and `[1,2,3]` will become `"1,2,3"`. To avoid this we added `strict` like this `yup.string().strict()` it prevents type conversion.

Can you, please, validate the places we added `strict` for strings? May be we don't need it in some places.

PR #374, opened 2023-08-30
feat: add root age parameter

0xPenryn:
adds `max_age` as an optional parameter for the `/verify` endpoint to set the maximum allowed age of the merkle root in seconds. must be between 3600 (1 hour) and 604800 (7 days), enforced with yup.

sets the `max_age` to be 3600 (1 hour) for Sign In with Worldcoin.

unable to test locally, please assist with testing.


web/src/pages/api/v1/verify/[app_id].ts
```diff
@@ -39,6 +39,16 @@ const schema = yup.object({
...
+  max_age: yup
+    .number()
+    .integer()
+    .min(3600, "Maximum root age cannot be less than 3600 seconds (1 hour).")
+    .max(
+      604800,
+      "Maximum root age cannot be more than 604800 seconds (7 days)."
+    )
+    .strict()
+    .optional(),
```
Comment (andy-t-wang) [inline web/src/pages/api/v1/verify/[app_id].ts:51]: Why do we want this to be optional? Shouldn't everyone be using an implementation that has a root age restriction

Comment (0xPenryn) [reply web/src/pages/api/v1/verify/[app_id].ts:51]: roots are already limited to one week of validity at the on-chain/sequencer level, and we don't want to break existing integrations

web/src/backend/verify.ts
```diff
@@ -24,6 +24,12 @@ const KNOWN_ERROR_CODES = [
     detail:
       "The provided proof is invalid and it cannot be verified. Please check all inputs and try again.",
   },
+  {
+    rawMessage: "Root provided in semaphore proof is too old.",
```
andy-t-wang, line 28: might be missing a rawCode

andy-t-wang (approved):
Still might be worth getting another look from Paolo, who has more context on sequencer

web/src/backend/verify.ts
0xPenryn: rawCode is error signature from smart contracts, the root age is enforced at the sequencer level so it doesn't have a similar rawCode

and we don't actually use the rawCode for matching errors in the dev portal side

pdtfh (approved):

PR #380, opened 2023-09-08
add action for auto-deploy

lukejmann:
Sends an event message to [worldcoin/developer-portal-deployment](github.com/worldcoin/developer-portal-deployment), which then triggers a deployment for the corresponding branch.


pdtfh:
new pipeline will pull directly from master, closing

PR #381, opened 2023-09-15
feat: allow input terms and privacy policy for action

michaelfurutin:
- added inputs to the Sign in with Worldcoin config page
- added attributes to the registration endpoint
- corrected precheck api route handler

image


igorosip0v (changes requested):

web/src/scenes/sign-in-with-world-id/Urls/UrlInput/index.tsx
```diff
@@ -1,44 +1,47 @@
...
+    mode: "onChange",
+    shouldFocusError: false,
+    defaultValues: {
+      url: props.value ?? "",
     },
-    [submit]
-  );
+  });
+
+  const handleSave = handleSubmit((data, event) => {
```
igorosip0v, line 38: @michaelfurutin We don't use event in this function

igorosip0v (approved):

PR #383, opened 2023-09-18
fix: no auth on /token with PKCE

0xPenryn:
bypasses the check of the authorization header or client secret on the /token endpoint if the `code_verifier` param is present, which is only when using auth code flow with PKCE

### DIFF
--- a/web/src/pages/api/v1/oidc/openid-configuration.ts
+++ b/web/src/pages/api/v1/oidc/openid-configuration.ts
@@ -37,5 +37,9 @@ export default async function handleOidcConfig(
       "id_token token", // Implicit flow
       "code id_token", // Hybrid flow
     ],
+    token_endpoint_auth_methods_supported: [
+      "client_secret_basic",
+      "client_secret_post",
+    ],
   });
 }
--- a/web/tests/integration/oidc/token.test.ts
+++ b/web/tests/integration/oidc/token.test.ts
@@ -9,6 +9,7 @@ import { setClientSecret, testGetDefaultApp } from "../test-utils";
 import * as jose from "jose";
 import { publicJwk } from "tests/api/__mocks__/jwk";
 import { createHash } from "crypto";
+import { generateOIDCJWT } from "src/backend/jwts";
 
 jest.mock("src/backend/kms", () => require("tests/api/__mocks__/kms.mock.ts"));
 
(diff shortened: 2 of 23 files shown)


web/src/pages/api/v1/oidc/openid-configuration.ts
```diff
@@ -37,5 +37,9 @@ export default async function handleOidcConfig(
       "id_token token", // Implicit flow
       "code id_token", // Hybrid flow
     ],
+    token_endpoint_auth_methods_supported: [
```
pdtfh, line 40: I think this is missing `none` because we'll support public clients

web/tests/integration/oidc/token.test.ts
```diff
@@ -419,4 +420,160 @@ describe("/api/v1/oidc/token", () => {
...
+    );
+
+    expect(payload).toEqual(
+      expect.objectContaining({
+        aud: app_id,
+      })
+    );
+    // #endregion
+
+    // #region -> Can get app_id using auth token without passing it directly
```
pdtfh, line 480: this should be a different test case, and naming should be around using client_secret_post

web/tests/integration/oidc/token.test.ts
```diff
@@ -419,4 +420,160 @@ describe("/api/v1/oidc/token", () => {
       })
     );
   });
+
+  test("Can get app_id using auth token", async () => {
```
pdtfh, line 424:
```suggestion
  test("can get token using auth code with client_secret_basic auth", async () => {
```

web/tests/integration/oidc/token.test.ts
```diff
@@ -419,4 +420,160 @@ describe("/api/v1/oidc/token", () => {
...
+    );
+
+    expect(payload2).toEqual(
+      expect.objectContaining({
+        aud: app_id,
+      })
+    );
+    // #endregion
+
+    // #region -> Can get app_id with code_verifier & client_id
```
pdtfh, line 528: should be a different test case, and naming around public client support

pdtfh (changes requested):
@0xPenryn let's do this slightly different to maintain security. we will allow devs to configure their app as public in the Dev Portal (similar to how [Auth0 does it](https://auth0.com/docs/get-started/applications/confidential-and-public-applications/view-application-type)).

if the client is public, we enforce PKCE.
if the client is confidential, we enforce authentication (and PKCE when applicable)

andy-t-wang:
@0xPenryn Close this?

PR #384, opened 2023-09-20
feat: improve redirect adding

michaelfurutin:
- now an empty redirect is not saved in the database
- a message about adding a redirect is displayed

https://github.com/worldcoin/developer-portal/assets/104859710/4062307f-a93c-4ef3-972d-358a4595efba

### DIFF
--- a/web/src/scenes/sign-in-with-world-id/Redirects/index.tsx
+++ b/web/src/scenes/sign-in-with-world-id/Redirects/index.tsx
@@ -1,4 +1,4 @@
-import { memo } from "react";
+import { memo, useState } from "react";
 import { Button } from "src/components/Button";
 import useSignInAction from "src/hooks/useSignInAction";
 import { RedirectInput } from "./RedirectInput";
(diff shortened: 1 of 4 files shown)


web/src/scenes/sign-in-with-world-id/Redirects/index.tsx
```diff
@@ -38,11 +40,22 @@ export const Redirects = memo(function Redirects() {
             onDelete={() => deleteRedirect({ id: redirect.id })}
           />
         ))}
+        {addRedirectFormShown && (
+          <RedirectInput
+            value=""
+            onChange={(value) => {
+              addRedirect({ uri: value }).then(() => {
+                setAddRedirectFormShown(false);
```
igorosip0v, line 48: @michaelfurutin This causes a little blink when you add a new redirect.

michaelfurutin: Resolved with implementing optimistic update and hidding form before success response.

igorosip0v (approved):

PR #385, opened 2023-09-27
feat: PKCE token auth fix tests

igorosip0v:
- Fix conflicts with main
- Add tests for #383 
- Adjust token endpoint
  - Used parsedParams instead of req.body where it's possible. (Please, review).


0xPenryn (approved):
lgtm, and parsedParams changes are a good update as well

PR #391, opened 2023-10-09
fix: Make external_nullifier optional

m1guelpf:


web/src/pages/api/v1/precheck/[app_id].ts
```diff
@@ -110,12 +111,9 @@ const createActionQuery = gql`
 `;
 
 const schema = yup.object({
-  action: yup.string().strict(),
+  action: yup.string().strict().required("This attribute is required."),
```
Comment (0xPenryn) [inline web/src/pages/api/v1/precheck/[app_id].ts:114]: we should require that it includes either action or external nullifier to avoid breaking any existing usage

0xPenryn (changes requested):

0xPenryn (approved):
lgtm

0xPenryn (changes requested):
actually api tests aren't passing -- taking a look now

0xPenryn (approved):
actually lgtm now

PR #410, opened 2023-11-09
fix: incorrect login for the user with existing email

igorosip0v:
This PR fixes the issue when a previously existing user who has an email in Hasua tries to log in with auth0 and gets redirected to the signup.

Main changes: 
- Now we are checking email too in case where we get email-user in login callback.
- Also now sync name and email on login callback.
- Updated permissions to allow service role change user's name and email.


web/src/api/auth/login-callback.ts
```diff
@@ -54,11 +54,10 @@ export const auth0Login = withApiAuthRequired(
 
     if (isEmailUser(auth0User)) {
       try {
-        const userData = await FetchUserByAuth0IdSdk(client).FetchUserByAuth0Id(
-          {
-            auth0Id: auth0User.sub,
-          }
-        );
+        const userData = await FetchUserByAuth0IdSdk(client).FetchEmailUser({
```
pdtfh, line 57: this can lead to edge cases of e.g. returning two different users. best to have 2 GraphQL queries (can be executed in the same request), one for email, one for auth0Id, prioritize the auth0Id and only use email as fallback

web/src/api/auth/login-callback.ts
```diff
@@ -95,11 +94,14 @@ export const auth0Login = withApiAuthRequired(
     }
 
     if (user && !user.auth0Id) {
-      // TODO: Sync user's email & name
       try {
-        const userData = await addAuth0Sdk(client).AddAuth0({
+        const userData = await updateUserSdk(client).UpdateUser({
```
pdtfh, line 98: worth syncing name & email if it changed, even if the user already had an auth0Id

web/src/api/auth/graphql/update-user.graphql
```diff
@@ -0,0 +1,8 @@
+mutation UpdateUser($id: String!, $_set: user_set_input) {
+  update_user_by_pk(pk_columns: { id: $id }, _set: $_set) {
+    id
+    team_id
```
pdtfh, line 4: think we don't need this

pdtfh (changes requested):
We should add a migration to make the email column unique to avoid collisions

web/src/api/auth/graphql/update-user.graphql
igorosip0v: Why? It is still used for the jwt to make requests to Hasura from the user role. For this, we getting team_id from the session.
After this query, we setting the user to the session.

web/src/api/auth/login-callback.ts
igorosip0v: Updated

web/src/api/auth/login-callback.ts
igorosip0v: Updated

igorosip0v:
@pdtfh Email field is already unique
image

web/src/api/auth/login-callback.ts
```diff
@@ -54,13 +56,21 @@ export const auth0Login = withApiAuthRequired(
...
+        if (userData.userByAuth0Id.length > 0) {
+          user = userData.userByAuth0Id[0];
+        }
+
+        if (
+          userData.userByAuth0Id.length === 0 &&
+          userData.userByEmail.length > 0
+        ) {
+          user = userData.userByEmail[0];
+        }
```
pdtfh, line 73: the logic here is not following [the specs](https://www.notion.so/worldcoin/WDP17-Auth0-Dev-Portal-Login-d1cff5fc500746578deac731da5d52c1?pvs=4#813cdf8ad0284c5f8d7a54519102c996) exactly, we should check for auth0Id first (takes precedence), then nullifier hash, then email. with the current logic we're checking nullifier hash first for users from Sign in with Worldcoin

web/src/api/auth/login-callback.ts
```diff
@@ -113,6 +124,35 @@ export const auth0Login = withApiAuthRequired(
       }
     }
 
+    const shouldUpdateUserName =
+      auth0User.name && (!user?.name || user?.name !== auth0User.name);
```
pdtfh, line 128:
```suggestion
      auth0User.name && user?.name !== auth0User.name;
```
let's use Auth0 as the source of truth, same for email

pdtfh:
re email uniqueness you're right, the bug is that we're not saving the user's email from Auth0 when signing up

web/src/api/auth/login-callback.ts
igorosip0v: if we want to prioritize requests like you suggest, then we'll need to make two requests:
1. Fetch user by auth0
2. in case there is no data returned we making request by email or by nullifier. 
3. Then mutation to add auth0 
4. mutation to sync name + email.

pdtfh: we could do a single GraphQL query to fetch user by Auth0Id, nullifier hash & email (separately as you have it). we then set the `user` variable based on the first result from the GraphQL query. once we have the user object we check whether we need to update name, email and/or Auth0Id. so it's minimum 1 request, max 2. thoughts?

pdtfh (approved):

PR #413, opened 2023-11-14
fix: logged user display

igorosip0v:
1. Updates how user name displays
2. Fixes error when team name was not displayed right after signup


pdtfh (approved):

web/src/scenes/signup/signup.tsx
```diff
@@ -72,13 +74,14 @@ export function SignUp(props: { hasAuth0User: boolean }) {
       });
 
       if (response.ok) {
+        checkSession();
```
pdtfh, line 77: let's add a comment on why we have this check

PR #415, opened 2023-11-19
fix: precheck

igorosip0v:
Problem: 
- When passing null to `nullifier_hash` with `nullable` the default value not applies.
- After this we passing null to the gql query where we expect String type and this is throws an error. 

Fix: 
- Add additional `transform` to the schema to convert `null` to `""`

### DIFF
--- a/web/src/pages/api/v1/precheck/[app_id].ts
+++ b/web/src/pages/api/v1/precheck/[app_id].ts
@@ -153,7 +153,6 @@ export default async function handlePrecheck(
     value: req.body,
   });
 
-  console.log("parsedParams:", parsedParams);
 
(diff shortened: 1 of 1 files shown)


0xPenryn (approved):

web/src/pages/api/v1/precheck/[app_id].ts
```diff
@@ -147,6 +153,8 @@ export default async function handlePrecheck(
     value: req.body,
   });
 
+  console.log("parsedParams:", parsedParams);
```
Comment (pdtfh) [inline web/src/pages/api/v1/precheck/[app_id].ts:156]:
```suggestion
```

pdtfh (approved):

PR #421, opened 2023-12-06
fix datadog log formatting

lukejmann:


pdtfh (approved):
doing this as an intermediary step to improve server-side logging. as a next step we'll attempt a different approach where we rely on vanilla console.{x} for logging

if the above doesn't work we need to add proper documentation clarifying how logging should happen server side vs. client side

PR #422, opened 2023-12-07
feat: remove phone insertion & revoking

pdtfh:
This PR is a partial implementation of #405, it removes endpoints that are safe to delete as they were only called server side.
- PR fully breaks the connection with the consumer backend and Dev Portal (this is no longer needed with the introduction of World ID Lite)


igorosip0v (approved):
Should we close #405?
@pdtfh

PR #423, opened 2023-12-07
feat: removes unused dependencies, unused code

pdtfh:


web/src/components/Link/index.tsx
```diff
@@ -47,11 +47,7 @@ const Link = memo(
             href={href}
             className={className}
             target={props.target || "_blank"}
-            rel={
```
pdtfh, line 50: @igorosip0v worth being more explicit with the `external` prop to avoid misconfigurations and having a external link without noopener

PR #424, opened 2023-12-07
remove /register endpoint for OIDC

pdtfh:
This endpoint presents UX & security challenges, like devs losing access to their apps. We've made the decision to remove it. Registration should be done through the Developer Portal.


0xPenryn (approved):
lgtm

PR #427, opened 2023-12-08
fix: small improvement to anti-XSS regex

darighost:
The current version of the regex can be bypassed by simply using a URL with one of the letters capitalized, viz:

```javascript
> /javascript:/.test("javaScript:alert(1)")
false
```

Browsers, though, will execute this code successfully. I tested this by making a link (`<a href="javaScript:alert(1)">test</a>`), and the JS code ran successfully.

This PR fixes this by adding the `i` flag to make the regex case insensitive.

### DIFF
--- a/web/src/lib/utils.ts
+++ b/web/src/lib/utils.ts
@@ -62,7 +62,7 @@ export const validateEmail = (candidate: string): boolean => {
 
 export const isSSR = () => typeof window === "undefined";
 
-export const uriHasJS = (uri: string) => /javascript:/.test(uri);
+export const uriHasJS = (uri: string) => /javascript:/i.test(uri);
 
 export const isEmailUser = (user: Auth0User): user is Auth0EmailUser =>
   user.sub.startsWith("email|");


web/src/lib/utils.ts
```diff
@@ -62,7 +62,7 @@ export const validateEmail = (candidate: string): boolean => {
 
 export const isSSR = () => typeof window === "undefined";
 
-export const uriHasJS = (uri: string) => /javascript:/.test(uri);
+export const uriHasJS = (uri: string) => /javascript:/i.test(uri);
```
esbn1, line 65: just found some common [examples](https://gist.github.com/xsuperbug/1aff5c1d5ddbfefb035f33dd9c8e8a72) to be used for XSS. Does it makes sense to cover them too? Seems like attacker can use them to pass that check

esbn1: @darighost, @paolodamico confirmed [here](https://tfh.slack.com/archives/C051Q5M346Q/p1702031891176449?thread_ts=1702029353.476689&cid=C051Q5M346Q), that we should handle all cases. Do you want to tackle it by yourself or should I create new PR for that cases?

darighost:
The best way to deal with workarounds would be to change from checking for JS URLs, to instead making sure a URL starts with either http:// or https:// and disallowing anything else (unless there's some other URL protocol you want to allow, in which case, you whitelist that too).

If you agree with this approach, let me know and I'll modify this PR. 

>paolodamico confirmed [here](https://tfh.slack.com/archives/C051Q5M346Q/p1702031891176449?thread_ts=1702029353.476689&cid=C051Q5M346Q)

It looks like I need some kind of special access to join that Slack? Or maybe there's a way to get an invite?

pdtfh:
thx for the contribution @darighost, ensuring the url starts with http:// or https:// is a good approach. also worth adding the validation in the `/api/v1/oidc/authorize` as a sanity check. this redirect URI validation is also enforced at the DB level.

esbn1:
> The best way to deal with workarounds would be to change from checking for JS URLs, to instead making sure a URL starts with either http:// or https:// and disallowing anything else (unless there's some other URL protocol you want to allow, in which case, you whitelist that too).
> 
> If you agree with this approach, let me know and I'll modify this PR.
> 
> > paolodamico confirmed [here](https://tfh.slack.com/archives/C051Q5M346Q/p1702031891176449?thread_ts=1702029353.476689&cid=C051Q5M346Q)
> 
> It looks like I need some kind of special access to join that Slack? Or maybe there's a way to get an invite?

good idea, that should work

ph3t:
Closing in favor of #707

PR #429, opened 2023-12-10
feat: support verification_level on /verify endpoint

0xPenryn:
~~TODO: IDKit update with VerificationLevel exported~~

api/v1/verify has been updated to accept both `verification_level` and `credential_type`

api/v2/* has been initiated with two main breaking changes -- /verify only accepts verification_level and enforces using the signal_hash rather than raw signal

any future api changes should be made to /v2, and a timeline for deprecating /v1 should be determined


0xPenryn:
superseded by #432

PR #430, opened 2023-12-11
Add logging env

lukejmann:


lukejmann:
This sets the env tag for logs sent directly from the http logger

pdtfh (approved):

PR #435, opened 2024-01-05
Analytics Upgrade for Dev Portal 2.0

andy-t-wang:
Added Identifiers for posthog analytics

### DIFF
--- a/hasura/metadata/databases/default/tables/public_user.yaml
+++ b/hasura/metadata/databases/default/tables/public_user.yaml
@@ -15,6 +15,7 @@ insert_permissions:
         - ironclad_id
         - is_subscribed
         - name
+        - posthog_id
         - team_id
         - world_id_nullifier
 select_permissions:
@@ -27,6 +28,7 @@ select_permissions:
         - id
         - ironclad_id
         - name
+        - posthog_id
         - team_id
         - updated_at
         - world_id_nullifier
--- a/hasura/migrations/default/1703352066173_create_table_public_test/down.sql
+++ b/hasura/migrations/default/1703352066173_create_table_public_test/down.sql
@@ -1,0 +1,1 @@
+DROP TABLE "public"."test";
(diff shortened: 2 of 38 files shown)


pdtfh (changes requested):

web/src/services/posthogClient.ts
```diff
@@ -0,0 +1,43 @@
+import { PostHog } from "posthog-node";
```
pdtfh, line 1: let's add clear documentation that this is for server-side use only. perhaps add a directive to prevent accidental use from the client side.

web/src/scenes/actions/NewAction/index.tsx
```diff
@@ -92,6 +93,11 @@ export function NewAction() {
         if (result instanceof Error) {
           throw result;
         }
+        posthog.capture("action_created", {
```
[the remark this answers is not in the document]
pdtfh: if we're keeping for now, let's add a note in the event name to clarify it's client-side

web/src/services/posthogClient.ts
```diff
@@ -0,0 +1,43 @@
...
+export function captureEvent({
+  event,
+  distinctId,
+  properties,
+}: CaptureEventParams): void {
+  if (!posthogClient) {
+    posthogClient = initializePostHog();
+  }
+  if (posthogClient) {
+    posthogClient.capture({
```
pdtfh, line 35: important to add the geoip_disable prop to make sure we don't use any server location. also worth syncing with @igorosip0v for advice on `flushAt` (or check PostHog docs)

pdtfh (approved):

web/src/services/posthogClient.ts
igorosip0v: @andy-t-wang we can add some `flush` function with `await posthog.flush()` inside to use it on the server where we need to make sure that we sent all events before leaving endpoint, for example.

Alternatively, we can use `await posthog.shutdownAsync()` inside `captureEvent`. But this will stop the client (and send all hanging events). And everytime we will use `captureEvent` we will create a client and kill it at the end.

igorosip0v: Or we can set the `flushAt: 1` property in the `PostHog` options, it should flush every event, but maybe it will be more optimized if we use `flush` manually. Not sure here

igorosip0v (approved):

PR #437, opened 2024-01-09
fix: allow undefined signals

pdtfh:
`signal` is an optional parameter. This PR fixes the case of not passing a signal which previously resulted in an unhandled error. Test case added to cover.

### DIFF
--- a/web/tests/api/verify.test.ts
+++ b/web/tests/api/verify.test.ts
@@ -149,6 +149,44 @@ describe("/api/v1/verify", () => {
       })
     );
   });
+
+  test("can verify without a signal", async () => {
+    // signal defaults to empty string if not provided
+
+    const noSignalPayload: Record<string, any> = { ...validPayload };
+    delete noSignalPayload.signal;
+
+    const payloads = [
+      noSignalPayload,
+      { ...validPayload, signal: undefined },
+      { ...validPayload, signal: "" },
+    ];
+
+    for (const payload of payloads) {
+      const { req, res } = createMocks({
+        method: "POST",
+        body: payload,
+        query: { app_id: "app_staging_112233445566778" },
+      });
+
+      // mocks sequencer response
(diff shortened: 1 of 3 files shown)


web/tests/api/verify.test.ts
```diff
@@ -149,6 +149,44 @@ describe("/api/v1/verify", () => {
...
   });
+
+  test("can verify without a signal", async () => {
+    // signal defaults to empty string if not provided
+
+    const noSignalPayload: Record<string, any> = { ...validPayload };
+    delete noSignalPayload.signal;
+
+    const payloads = [
+      noSignalPayload,
```
andy-t-wang, line 160: this still has a signal since it's set to `semaphoreProofParamsMock` in line 8 which is defined as 

`export const semaphoreProofParamsMock = {
  proof:
    "0x13046d1af7e62f5b48f0abbd7bf90a0dbbed2eedcde362de280f0582069364ff180406ee4160f3649944d083c809a91f0e352dfcb6d4afba8bb35140c9dfc2171c3e646044e3af21aaf9d5c1674e941b1061e5512b1072f9ab7d880851df425c1c8f7a1d556092d201a7678244fddaef11287edd7aa5ea27e126e457227977b02156fdfc9d31e5b308b9728a9753333bc2c7e560087584fb5efbfae18e901818262f99affb31662d5bc0fdfdd1189bd54c662da2653538d4973c3f042e002ab20ad3756c84904a910c8663e775433923c18ef010843859b71e82af7e1856f14716acf42e25976ba0a7cb86ee58ee2ff8471a6563f540831bbb22d0a373087d1f",
  nullifier_hash:
    "0x0447c1b95a5a808a36d3966216404ff4d522f1e66ecddf9c22439393f00cf616",
  merkle_root:
    "0x0936d98c83151035b528d1631df5c3607a740bd296b4c79c627130a96645dcc7",
  verification_level: "orb",
  signal: "my_signal",
};`

andy-t-wang (approved):

PR #442, opened 2024-01-10
feat: app-database-form update with new fields

andy-t-wang:
Add new fields to the UI and allow updating. This stuff currently publishes directly which replicates the unverified app flow. Next feature will be the verification flow. These values are not displayed anywhere currently so it shouldn't present any risks.

### DIFF
--- a/web/src/hooks/useApps.ts
+++ b/web/src/hooks/useApps.ts
@@ -11,6 +11,12 @@ import useSWR from "swr";
 import useSWRMutation from "swr/mutation";
 import { shallow } from "zustand/shallow";
 
+enum DescriptionSubFields {
+  DescriptionOverview = "description_overview",
+  DescriptionHowItWorks = "description_how_it_works",
+  DescriptionConnect = "description_connect",
+}
+
 const appFields = `
 id
 verified_app_logo
@@ -77,12 +83,6 @@ const DeleteAppQuery = gql`
     }
   }
 `;
-
-const descriptionSubFields = [
-  "description_overview",
-  "description_how_it_works",
-  "description_connect",
-];
 
 const fetchApps = async () => {
   const response = await graphQLRequest<{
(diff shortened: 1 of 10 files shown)


andy-t-wang:
Also removed some individual mutators in use

Uploading Screen Recording 2024-01-10 at 6.54.55 PM.mov…

app that are no longer needed like updatedescription and name

andy-t-wang:
@paolodamico

web/src/hooks/useApps.ts
```diff
@@ -301,19 +312,46 @@ const useApps = () => {
...
+    return {};
+  };
+
+  const encodeDescription = (
+    description_overview: string,
+    description_how_it_works: string = "",
+    description_connect: string = ""
+  ) => {
+    return JSON.stringify({
+      [descriptionSubFields[0]]: description_overview,
```
igorosip0v, line 337: @andy-t-wang I suggest to make a TS enum here instead of descriptionSubFields object.

web/src/scenes/actions/common/Form/FieldInput/index.tsx
```diff
@@ -7,26 +7,45 @@ interface FieldInputProps extends InputHTMLAttributes<HTMLInputElement> {
...
+  maxChar?: number; // New prop for maximum characters
 }
 
 export const FieldInput = memo(function FieldInput(props: FieldInputProps) {
-  const { className, register, errors, isDirty, ...otherProps } = props;
+  const {
+    className,
+    register,
+    errors,
+    value = "",
```
igorosip0v, line 19: @andy-t-wang If I'm not mistaken you can get `control` from the `useForm` and pass it to the component. And then use `useWatch` inside the FieldInput component.

andy-t-wang: passing in watch from the call now

PR #443, opened 2024-01-11
Feat: Add Pending App Metadata table to Hasura

andy-t-wang:
This table will be used to push the information for apps that need to be reviewed. Included a new permission group "Reviewer" which can read all values, update only status and message, and delete any rows upon confirmed review


hasura/metadata/databases/default/tables/public_pending_app_metadata.yaml
```diff
@@ -0,0 +1,103 @@
...
+object_relationships:
+  - name: pending_app_metadata_team
+    using:
+      foreign_key_constraint_on: team_id
+insert_permissions:
+  - role: user
+    permission:
+      check: {}
+      columns:
+        - id
```
igorosip0v, line 13: @andy-t-wang As far as I know we shouldn't give permission to user to edit `id`, `created_at`, `updated_at` fields, it will be created automatically. We should only allow to insert fields that the user role will send directly from the client side. 
Same with update

hasura/migrations/default/1704946584423_create_table_pending_app_metadata/up.sql
```diff
@@ -0,0 +1,24 @@
+
+CREATE TABLE "public"."pending_app_metadata" ("id" varchar NOT NULL, "name" varchar NOT NULL, "logo_img_url" text NOT NULL, "showcase_img_urls" jsonb NOT NULL, "hero_image_url" text NOT NULL, "description" text NOT NULL, "category" varchar NOT NULL, "publisher" varchar NOT NULL, "link_to_integration" varchar NOT NULL, "review_message" text NOT NULL, "review_status" varchar NOT NULL, "is_developer_allow_listing" bool NOT NULL, "world_app_description" varchar NOT NULL, "team_id" varchar not null, "created_at" timestamptz NOT NULL DEFAULT now(), "updated_at" timestamptz NOT NULL DEFAULT now(), PRIMARY KEY ("id") );
```
igorosip0v, line 2: "id" varchar NOT NULL DEFAULT gen_random_friendly_id ('something'),

igorosip0v (changes requested):

andy-t-wang:
Note added permissions for reviewer to view basic team and app data as well

hasura/metadata/databases/default/tables/public_pending_app_metadata.yaml
```diff
@@ -0,0 +1,115 @@
...
+        - description
+        - category
+        - publisher
+        - link_to_integration
+        - review_message
+        - review_status
+        - is_developer_allow_listing
+        - world_app_description
+        - team_id
+        - created_at
```
igorosip0v, line 74: @andy-t-wang Are we using id, created_at and updated_at on the frontend side?

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/down.sql
```diff
@@ -0,0 +1,6 @@
+
+DROP TABLE "public"."pending_app_metadata";
```
igorosip0v, line 2: @andy-t-wang In your `up` migration you create the table first and `validate_metadata_uris` at the end. The order of actions in the `down` migration should be exactly vice versa.

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
```diff
@@ -0,0 +1,89 @@
...
+  IF url IS NOT NULL AND url != '' THEN
+    IF NOT (url ~* '^https://([[:alnum:]_-]+\.)+[[:alnum:]_-]+(/[[:alnum:]_\-./?%&=]*)?$') THEN
+      RAISE EXCEPTION USING ERRCODE= '22000', MESSAGE= 'Invalid URL format. URLs must use HTTPS protocol.';
+    END IF;
+  END IF;
+END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION validate_metadata_urls()
```
igorosip0v, line 67: @andy-t-wang Not sure, correct me if I'm wrong. But you calling this function on insert and checking fields that can be optional, for example source code URL, as far as I know. In this case `validate_single_url` will throw exception, right? and it will be impossible to make insert.

igorosip0v (changes requested):

hasura/metadata/databases/default/tables/public_pending_app_metadata.yaml
andy-t-wang: We will want to show how long things have been waiting for review. So I will leave updated_at exposed and the rest will be removed from permissions

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
andy-t-wang: Those two fields allow null and the sql checks ` IF url IS NOT NULL AND url != '' THEN:` so should be good. Just tested it to with an insert and it worked!

igorosip0v (approved):

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
```diff
@@ -0,0 +1,89 @@
...
+RETURNS TRIGGER AS $$
+DECLARE
+  url text;
+BEGIN
+  -- Validate each URL field
+  PERFORM validate_single_url(NEW.logo_img_url);
```
pdtfh, line 73: should we constraint images to a specific host?

PR #444, opened 2024-01-11
feat: Added new Fields to App Table 

andy-t-wang:
Synced with Ian and Michael today and they wanted some more fields in the app table. Added them as well as a trigger to check user inputted urls are correct. No option to currently include them on the form. Will add that with the pending_app_table front end change


hasura/migrations/default/1705005696536_squashed/up.sql
```diff
@@ -0,0 +1,30 @@
+alter table "public"."app"
+add column "showcase_img_urls" jsonb null;
```
pdtfh, line 2: not a fan of having this as a JSON, will require different handling than the rest of the images, can we add this later?

hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -19,6 +19,11 @@ computed_fields:
       function:
         name: get_app_is_verified
         schema: public
+  - name: show_details_page
+    definition:
```
pdtfh, line 23: what is this doing?

andy-t-wang: A field that is used in the app to decide to open the details page or just go directly to the link.

hasura/migrations/default/1705005696536_squashed/up.sql
andy-t-wang: ok I will make this text [] instead. And we can have the user store the paths but I can just add a validation on the db that the inputs can only be the fixed paths we predetermined

andy-t-wang:
We are changing the structure to a single metadata table to manage this. See updated specs WDP25

PR #445, opened 2024-01-12
Feat: app_metadata_table hasura changes

andy-t-wang:
Specs changed to have only one table. Starting with app_metadata_table since it's not breaking.


andy-t-wang:
API key and users basically have the same permissions. Service can only select since it's not doing any inserting or updating from what I can see

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
```diff
@@ -1,26 +1,28 @@
-CREATE TABLE "public"."pending_app_metadata" (
```
pdtfh, line 1: let's rename this migration as well

pdtfh: should we also drop the migration `1704828243607_alter_table_app_verifications`?

andy-t-wang: Yeah removing that migration will just break the current main branch since it's using those fields now so I was planning do it in a separate PR with the front end updates to use the new metadata table

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
```diff
@@ -1,26 +1,28 @@
...
   "description" text NOT NULL,
-  "category" varchar NOT NULL,
-  "publisher" varchar NOT NULL,
-  "link_to_integration" varchar NOT NULL,
-  "review_message" text NOT NULL,
-  "review_status" varchar NOT NULL,
-  "is_developer_allow_listing" bool NOT NULL,
-  "world_app_description" varchar NOT NULL,
-  "team_id" varchar NOT NULL,
+  "world_app_description" varchar NULL,
```
pdtfh, line 9: generally not a good idea to allow NULL in varchar fields because you then have two values representing empty/nil, empty string should be enough. important to make sure the default is an empty string

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
```diff
@@ -70,20 +67,79 @@ DECLARE
...
+  END IF;
+
+  IF NEW.showcase_img_urls IS NOT NULL THEN
+    IF array_length(NEW.showcase_img_urls, 1) > 3 THEN
+      RAISE EXCEPTION 'showcase_img_urls can have a maximum of 3 images';
+    END IF;
+
+    FOREACH item IN ARRAY NEW.showcase_img_urls
+    LOOP
+      IF item !~* '^(showcase_img_urls_[1-3]\.(png|jpg))$' THEN
```
pdtfh, line 106:
```suggestion
      IF item !~* '^(showcase_img_[1-3]\.(png|jpg))$' THEN
```

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -0,0 +1,193 @@
...
+  name: app_metadata
+  schema: public
+object_relationships:
+  - name: app
+    using:
+      foreign_key_constraint_on: app_id
+insert_permissions:
+  - role: api_key
+    permission:
+      check: {}
```
pdtfh, line 11: these permissions need constraints, both `user` and `api_key` can only be allowed to create records for apps they have permissions to

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - source_code_url
+        - verified_at
+        - reviewed_by
+        - is_reviewer_app_store_approved
+        - is_reviewer_world_app_approved
+        - review_message
+        - status
+        - created_at
+        - updated_at
+      filter: {}
```
pdtfh, line 93: let's constraint this to status = awaiting_review

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - is_developer_allow_listing
+        - world_app_description
+        - app_website_url
+        - source_code_url
+      filter:
+        app:
+          team:
+            id:
+              _eq: x-hasura-Team-Id
+      check: {}
```
pdtfh, line 157: this needs a restriction on only being set for unverified rows (same for user)

pdtfh (changes requested):
some permissions need updates but looks good otherwise

hasura/migrations/default/1704993161810_create_table_pending_app_metadata/up.sql
```diff
@@ -70,20 +67,79 @@ DECLARE
...
+
+CREATE UNIQUE INDEX "unique_verified_app_id" ON "public"."app_metadata" (app_id)
+WHERE status = 'verified';
+
+CREATE OR REPLACE FUNCTION "enforce_app_id_row_limit"()
+RETURNS TRIGGER AS $$
+DECLARE
+  app_id_count integer;
+BEGIN
+  SELECT COUNT(*)
```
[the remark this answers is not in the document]
andy-t-wang: From what I can see it looks like the opposite is true for postgres https://blog.jooq.org/whats-faster-count-or-count1/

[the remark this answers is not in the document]
andy-t-wang: Although very miniscule

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - verified_at
+        - reviewed_by
+        - is_reviewer_app_store_approved
+        - is_reviewer_world_app_approved
+        - review_message
+        - status
+        - created_at
+        - updated_at
+      filter: {}
+  - role: service
```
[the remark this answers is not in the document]
andy-t-wang: precheck pulls stuff like the app name and img_url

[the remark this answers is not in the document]
pdtfh: ah good call, also worth constraining to unverified / verified?

hasura/migrations/default/1704993161810_create_table_app_metadata/up.sql
```diff
@@ -0,0 +1,145 @@
+CREATE TABLE "public"."app_metadata" (
+  "id" varchar NOT NULL UNIQUE DEFAULT gen_random_friendly_id('update'),
```
pdtfh, line 2:
```suggestion
  "id" varchar NOT NULL UNIQUE DEFAULT gen_random_friendly_id('meta'),
```

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -0,0 +1,203 @@
...
+        - category
+        - integration_url
+        - is_developer_allow_listing
+        - world_app_description
+        - app_website_url
+        - source_code_url
+      filter:
+        _and:
+          - status:
+              _eq: unverified
```
pdtfh, line 191: for a subsequent improvement but this will also need permissions to change the status back to unverified if it's awaiting_review (to pull back from review)

pdtfh (approved):

hasura/metadata/databases/default/tables/public_app_metadata.yaml
andy-t-wang: Ok will make this part of a new commit when I start adding the review status. Probably like 2 merges away

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -0,0 +1,218 @@
...
+          _eq: verified
+  - role: api_key
+    permission:
+      filter:
+        app:
+          team:
+            id:
+              _eq: X-Hasura-Team-Id
+  - role: user
+    permission:
```
pdtfh, line 212: we don't want to allow deletion of verified rows, dev can submit an update request, make the app inactive or delete it

hasura/metadata/databases/default/tables/public_app_metadata.yaml
pdtfh: this check should be added pre-merge

PR #446, opened 2024-01-14
feat: Migration to Metadata table

andy-t-wang:
- Remove the migration that added the metadata fields to the app table, adds the new fields to metadata
- Integration test db initialization and precheck test were updated
- Most of the files changed are just updating from app -> app.metadata 
- Precheck was updated to the new schema but the fields returned were kept the same to prevent breaking
- UseApps
  - Logic updated to update metadata table except for status
  - Update logic uses an upsert protected by a trigger that ensures only one non verified row per app ID can exist
  - Fetch apps and update both have failsafes in the event the insert app metadata half of the insert fails to prevent orphan apps
  - All screens will show the unverified changes since this is the dev portal, I feel like this makes the most sense as this should focus on staging changes

Unfortunately as a result of this the diff is a bit large (Sorry Paolo :sweat_smile:) if you have some thoughts on how to split this in the future without breaking main that would be great! 

Note: 
1. Saving changes to pending
2. Removing changes awaiting review
3. Some fields and the review message + status were not included in this diff those will follow


hasura/migrations/default/1704993161810_create_table_app_metadata/up.sql
```diff
@@ -17,6 +17,7 @@ CREATE TABLE "public"."app_metadata" (
   "is_reviewer_app_store_approved" bool NOT NULL DEFAULT false,
   "is_reviewer_world_app_approved" bool NOT NULL DEFAULT false,
   "review_message" text NOT NULL DEFAULT '',
+  "unique_verification_status_row" varchar NOT NULL UNIQUE,
```
pdtfh, line 20: let's document the rationale for this attribute

hasura/migrations/default/1705087600730_alter_table_app_drop_metadata/up.sql
```diff
@@ -0,0 +1,11 @@
+alter table "public"."app"
```
pdtfh, line 1: destructive in production. we first need to create a migration that moves all existing values to `app_matadata`, then in a separate PR we incorporate this migration

web/src/pages/api/v1/precheck/[app_id].ts
```diff
@@ -61,10 +70,17 @@ const appPrecheckQuery = gql`
     ) {
       id
       is_staging
-      is_verified
-      name
-      verified_app_logo
       engine
+      app_metadata(where: { status: { _neq: "verified" } }) {
+        name
+        logo_img_url
```
Comment (pdtfh) [inline web/src/pages/api/v1/precheck/[app_id].ts:76]: we will never need this attribute if the app is not verified

web/src/pages/api/v1/precheck/[app_id].ts
```diff
@@ -189,7 +205,28 @@ export default async function handlePrecheck(
       req
     );
   }
+  const app_metadata: AppMetadataPayload | undefined = Array.isArray(
```
Comment (pdtfh) [inline web/src/pages/api/v1/precheck/[app_id].ts:208]: I may be wrong but I believe `rawAppValues.app_metadata` should always be an array?

hasura/migrations/default/1704993161810_create_table_app_metadata/up.sql
```diff
@@ -134,6 +134,23 @@ END;
 $$ LANGUAGE plpgsql;
 
 CREATE TRIGGER "trigger_enforce_app_id_row_limit"
+AFTER INSERT OR UPDATE ON "public"."app_metadata"
+FOR EACH ROW
+EXECUTE FUNCTION enforce_app_id_row_limit();
+
+CREATE FUNCTION set_unique_verification_status_row()
```
pdtfh, line 141: how about using a trigger like you're suggesting but without having an additional field. we can reject changes that would result in the trigger constraint being violated. this way we can also get more useful error messages

hasura/migrations/default/1704993161810_create_table_app_metadata/up.sql
```diff
@@ -42,12 +43,11 @@ EXECUTE PROCEDURE "public"."set_current_timestamp_updated_at"();
 COMMENT ON TRIGGER "set_public_app_metadata_updated_at" ON "public"."app_metadata"
 IS 'trigger to set value of column "updated_at" to current timestamp on row update';
 
-
 CREATE OR REPLACE FUNCTION validate_single_url(url text)
 RETURNS void as $$
 BEGIN
   IF url IS NOT NULL AND url != '' THEN
-    IF NOT (url ~* '^https://([[:alnum:]_-]+\.)+[[:alnum:]_-]+(/[[:alnum:]_\-./?%&=]*)?$') THEN
+    IF NOT (url ~* '^https://([[:alnum:]_-]+\.)+[[:alnum:]_-]+(/[[:alnum:]_\-./?%&#=]*)?$') THEN
```
pdtfh, line 50: let's add a comment here that for these particular URLs we allow hash params

hasura/migrations/default/1704993161810_create_table_app_metadata/up.sql
```diff
@@ -42,12 +43,11 @@ EXECUTE PROCEDURE "public"."set_current_timestamp_updated_at"();
 COMMENT ON TRIGGER "set_public_app_metadata_updated_at" ON "public"."app_metadata"
 IS 'trigger to set value of column "updated_at" to current timestamp on row update';
 
-
```
pdtfh, line 45: these changes need to be now in a new migration

hasura/migrations/default/1705187600730_alter_table_app_drop_metadata/up.sql
```diff
@@ -0,0 +1,11 @@
+alter table "public"."app"
```
pdtfh, line 1: let's move this to a separate PR as we'll need to merge this PR before the initial production deployment

hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -13,30 +13,25 @@ array_relationships:
...
-        name: get_app_is_verified
-        schema: public
-  - name: verified_app_logo
-    definition:
-      function:
-        name: get_verified_app_logo
-        schema: public
+  - name: app_metadata
+    using:
+      manual_configuration:
```
pdtfh, line 18: why does this need `manual_configuration`? seems like it should work similar to how actions are linked above?

hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -13,30 +13,25 @@ array_relationships:
...
+          name: app_metadata
+          schema: public
 insert_permissions:
   - role: api_key
     permission:
       check: {}
       set:
         team_id: x-hasura-Team-Id
       columns:
-        - description_internal
```
pdtfh, line 34: let's keep for now

hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -72,55 +62,31 @@ select_permissions:
       columns:
         - id
         - is_staging
-        - name
         - team_id
-      computed_fields:
-        - is_verified
-        - verified_app_logo
```
pdtfh, line 79: you also need to drop these computed fields with a migration

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -67,6 +65,9 @@ select_permissions:
         - app_website_url
         - source_code_url
         - verified_at
+        - is_reviewer_app_store_approved
+        - is_reviewer_world_app_approved
+        - unique_verification_status_row
```
pdtfh, line 70: drop

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -127,7 +131,10 @@ select_permissions:
         - app_website_url
         - source_code_url
         - verified_at
+        - is_reviewer_app_store_approved
+        - is_reviewer_world_app_approved
         - review_message
+        - unique_verification_status_row
```
pdtfh, line 137: drop

web/src/pages/api/v1/precheck/[app_id].ts
```diff
@@ -189,7 +204,25 @@ export default async function handlePrecheck(
       req
     );
   }
-
+  const app_metadata =
+    rawAppValues.app_metadata.length > 0
```
Comment (pdtfh) [inline web/src/pages/api/v1/precheck/[app_id].ts:208]: is this check even needed, if length is 0, then `rawAppValues.app_metadata[0]` will be `undefined` anyways

web/src/pages/kiosk/[action_id].tsx
```diff
@@ -61,9 +68,17 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
   if (data?.action[0].action === "") {
     return { props: { error_code: "no_sign_in" } };
   }
-
+  const processedAction: ActionKioskType = {
+    ...data.action[0],
+    app: {
+      ...data.action[0].app,
+      app_metadata: data.action[0].app.app_metadata?.[0] || null,
```
Comment (pdtfh) [inline web/src/pages/kiosk/[action_id].tsx:75]: let's do a cleaner approach here where the backend already handles the logic for what to display, i.e. we only pass a single `app_metadata.name` to the client, a single `app_metadata.is_verified`. if we pass both, the client needs to decide what to display, which will be prone to errors

Comment (pdtfh) [reply web/src/pages/kiosk/[action_id].tsx:75]: similar to how we have it in precheck

web/src/scenes/app/Configuration.tsx
```diff
@@ -29,37 +29,37 @@ const saveSchema = yup.object().shape({
...
 
 type ConfigurationFormValues = yup.Asserts<typeof saveSchema>;
 
 export const Configuration = memo(function Configuration() {
   const currentApp = useAppStore((store) => store.currentApp);
-  const { updateAppData, parseDescription, encodeDescription } = useApps();
+  const { updateAppMetadata, parseDescription, encodeDescription } = useApps();
 
-  const descriptionInternal = parseDescription(currentApp);
+  const descriptionInternal = parseDescription(currentApp?.app_metadata);
```
pdtfh, line 53: this is a different description, let's change to make sure we're updating the public description only right now

web/src/scenes/app/Configuration.tsx
```diff
@@ -78,12 +78,13 @@ export const Configuration = memo(function Configuration() {
 
       const updatedData = {
         ...rest,
-        description_internal: descriptionsJSON,
+        description: descriptionsJSON,
       };
-      await updateAppData(updatedData);
+      console.log(updatedData);
```
pdtfh, line 83: stale console.log

hasura/metadata/databases/default/tables/public_app.yaml
andy-t-wang: This keeps suggesting an object relationship when we need an array one so I made it manually

hasura/metadata/databases/default/tables/public_app.yaml
andy-t-wang: Dropped the functions but the fields are just removed in the metadata

web/src/scenes/app/AppHeader.tsx
andy-t-wang, line 1: Note we are showing unverified since it's the internal portal for devs. Part of the form update in a later pr will add a toggle to show the verified and unverified

hasura/metadata/databases/default/tables/public_app.yaml
pdtfh: if you follow the same pattern as for actions and reload metadata?

web/src/pages/kiosk/[action_id].tsx
```diff
@@ -61,9 +69,20 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
   if (data?.action[0].action === "") {
     return { props: { error_code: "no_sign_in" } };
   }
+  const { app_metadata, verified_app_metadata, ...app_data } =
+    data.action[0].app;
+  const app_metadata_item = verified_app_metadata?.[0] ?? app_metadata?.[0];
```
Comment (pdtfh) [inline web/src/pages/kiosk/[action_id].tsx:74]: not fully correct, we should not pass image URLs that don't come from a verified app

Comment (pdtfh) [inline web/src/pages/kiosk/[action_id].tsx:74]: for both clarity and data consistency, let's pass `logo_img_url = ""` for unverified apps

web/src/scenes/app/Configuration.tsx
pdtfh: still needs to be removed

hasura/migrations/default/1705344968980_modify-app-metadata-functions/up.sql
```diff
@@ -0,0 +1,39 @@
+ALTER TABLE "public"."app_metadata"
+ADD COLUMN unique_verification_status_row varchar NOT NULL UNIQUE;
+
+CREATE OR REPLACE FUNCTION validate_single_url(url text)
+RETURNS void as $$
+BEGIN
+  IF url IS NOT NULL AND url != '' THEN
+--   Allow hash params now
```
pdtfh, line 8:
```suggestion
--   Hash parameters are allowed for these URLs
```

hasura/metadata/databases/default/tables/public_app.yaml
andy-t-wang: I tried to make it in the yaml directly but it then didn't show the relationship (still worked) in the console and thought it would be confusing to anyone else looking at it so I changed it back.

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -196,13 +202,7 @@ update_permissions:
                 users:
                   id:
                     _eq: X-Hasura-User-Id
-      filter: {}
 delete_permissions:
-  - role: reviewer
```
pdtfh, line 201: FYI, it is possible that an app ends up without any app_metadata, code should account for that

web/src/lib/types.ts
```diff
@@ -67,19 +67,30 @@ export interface IPendingProofResponse {
...
+};
+
+export type ActionKioskType = Omit<ActionKioskQueryType, "app"> & {
+  app: Omit<
+    ActionKioskQueryType["app"],
+    "app_metadata" | "verified_app_metadata"
+  > & {
+    app_metadata: Pick<
+      AppMetadataModel,
+      "name" | "logo_img_url" | "status"
```
pdtfh, line 91: logo_image_url may not be present (in case of unverified apps)

web/src/components/Layout/NewAppDialog/index.tsx
```diff
@@ -99,7 +105,7 @@ export const NewAppDialog = memo(function NewAppDialog(
               className="w-full font-rubik"
               placeholder="For internal reference. Visible only to you and your team."
               type="text"
-              {...register("description_internal")}
```
pdtfh, line 102: the placeholder above is wrong, there may also be other indications in the UI that are wrong

web/src/hooks/useApps.ts
```diff
@@ -19,140 +19,299 @@ enum DescriptionSubFields {
...
 
-  if (response.data?.app.length) {
-    return response.data.app;
+  const apps = response.data?.app || [];
+  // Even verified apps will have an unverified row. Thus the only case this will be true is if we have an orphaned app row
+  const appsMissingMetadata = apps.filter((app) => !app.app_metadata);
+
+  // Checks for orphaned app rows and tries to add a metadata row
+  if (appsMissingMetadata.length > 0 && !attemptedToHandleMissingMetadata) {
+    await _handleMissingMetadata(appsMissingMetadata);
```
pdtfh, line 169: doing this when fetching apps is mixing concerns, perhaps best to do this when doing updates?

web/src/hooks/useApps.ts
```diff
@@ -19,140 +19,299 @@ enum DescriptionSubFields {
...
+    return fetchApps(true);
   }
-  return [];
+  // for each app we need to call _parseAppModel to make sure the app_metadata is not an array
+  apps.forEach((app, index, this_arr) => {
```
pdtfh, line 173: this could be simplified with an Array.map

web/src/hooks/useApps.ts
```diff
@@ -19,140 +19,299 @@ enum DescriptionSubFields {
...
+    this_arr[index] = _parseAppModel(app);
+  });
+  console.log("apps", apps);
```
pdtfh, line 176: stale console.log

web/src/hooks/useApps.ts
```diff
@@ -165,44 +324,90 @@ const deleteAppFetcher = async (
...
   const response = await graphQLRequest<{
-    app: Pick<AppModel, "id">;
+    delete_app_by_pk: string;
   }>({
     query: DeleteAppQuery,
     variables: { id },
   });
-  if (response.data?.app) {
-    return response.data?.app;
+  console.log(response);
```
pdtfh, line 332: another console.log

web/src/hooks/useApps.ts
```diff
@@ -1,10 +1,10 @@
-import { AppModel } from "@/lib/models";
```
pdtfh, line 1: @igorosip0v worth taking a look when you get a chance, I'm out of the loop with frontend

web/src/hooks/useApps.ts
```diff
@@ -165,44 +324,90 @@ const deleteAppFetcher = async (
...
+  }
+  const appResponse = await graphQLRequest<{
     insert_app_one: AppModel;
   }>({
     query: InsertAppQuery,
     variables: {
-      object: {
-        name,
-        description_internal,
+      appObject: {
```
pdtfh, line 354: wondering if we can insert both in a single query

web/src/hooks/useApps.ts
```diff
@@ -165,44 +324,90 @@ const deleteAppFetcher = async (
...
+  if (
+    appMetadataResponse.data?.insert_app_metadata_one &&
+    appResponse.data?.insert_app_one
+  ) {
+    const newApp = {
+      ...appResponse.data?.insert_app_one,
+      app_metadata: appMetadataResponse.data?.insert_app_metadata_one,
+      verified_app_metadata: undefined,
+    };
+    console.log("newApp", newApp);
```
pdtfh, line 380:
```suggestion
```

hasura/migrations/default/1705344968980_modify-app-metadata-functions/up.sql
```diff
@@ -0,0 +1,39 @@
...
+
+DROP TRIGGER IF EXISTS "trigger_enforce_app_id_row_limit" ON "public"."app_metadata";
+
+DROP FUNCTION IF EXISTS "enforce_app_id_row_limit";
+
+CREATE FUNCTION set_unique_verification_status_row()
+RETURNS TRIGGER AS $$
+BEGIN
+  IF NEW.status = 'verified' THEN
+    NEW.unique_verification_status_row := NEW.app_id || '_verified';
```
pdtfh, line 24: suggest changing to `can_row_be_edited` for clarity or something along those lines

pdtfh (approved):

hasura/metadata/databases/default/tables/public_app_metadata.yaml
andy-t-wang: yeah we account for this in the update and fetch

web/src/lib/types.ts
andy-t-wang: yeah but will be empty string so this won't break anything and is good for consistency

hasura/migrations/default/1705344968980_modify-app-metadata-functions/up.sql
andy-t-wang: changed to is_row_verified to more accurately reflect it's value

web/src/hooks/useApps.ts
andy-t-wang: cannot we need the id from the first call and apollo doesn't let you chain that way

PR #449, opened 2024-01-15
fix: migrate database before dropping columns and rollback app-table updates

andy-t-wang:
Migrating verified_at, name, app_id, logo before dropping those columns. 

Also remove the migration to add new fields to app table. Fields were never used so will not be breaking 

Full permissions should be looked at in #446. This is just rolling back the changes and migrating the database


hasura/migrations/default/1704993161810_create_table_app_metadata/up.sql
andy-t-wang, line 1: Temporarily remove this trigger to migrate the image values. We can write a script to upload all the images to the corresponding columns and then update the db with the image upload PR. Will add this trigger back then.

hasura/migrations/default/1705087600730_migrate_dropped_app_rows_to_app_metadata/down.sql
```diff
@@ -0,0 +1 @@
+-- Not sure there is a down migration for this?
```
igorosip0v, line 1: @andy-t-wang As I can see you are filling empty table, if so, then you can create down migration to remove all rows in the table, just in case.

igorosip0v (approved):

hasura/migrations/default/1705087600730_migrate_dropped_app_rows_to_app_metadata/down.sql
andy-t-wang: resolved

pdtfh (approved):

PR #451, opened 2024-01-16
Fix: create edit rows for already verified apps

andy-t-wang:
If some rows already are verified we need to create their verified rows so that users can edit.

### DIFF
--- a/hasura/migrations/default/1705367428341_create_edit_rows_for_verified_apps/up.sql
+++ b/hasura/migrations/default/1705367428341_create_edit_rows_for_verified_apps/up.sql
@@ -1,0 +1,11 @@
+INSERT INTO
+    "public"."app_metadata" (app_id, name, logo_img_url, verification_status)
+SELECT
+    app_id,
+    name,
+    logo_img_url,
+    'unverified'
+FROM
+    "public"."app_metadata"
+WHERE
+    verification_status = 'verified';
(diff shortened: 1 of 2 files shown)


hasura/migrations/default/1705344968980_modify-app-metadata-functions/up.sql
```diff
@@ -29,9 +29,9 @@ CREATE FUNCTION set_is_row_verified()
 RETURNS TRIGGER AS $$
 BEGIN
   IF NEW.verification_status = 'verified' THEN
-    NEW.is_row_verified := false;
-  ELSE
     NEW.is_row_verified := true;
+  ELSE
+    NEW.is_row_verified := false;
```
pdtfh, line 34: let's edit the function in the new migration, this migration is likely already applied

hasura/migrations/default/1705367428341_create_edit_rows_for_verified_apps/up.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO
```
pdtfh, line 1: do we need this backfilling? would be better if the frontend already handles this by letting users add an unverified row to make changes. bc when an app_metadata record moves from is_awaiting_review to verified by a reviewer, this row won't get created anyways

andy-t-wang: Ok the front end currently handles it expecting the row to be there at all times and when approved by the reviewer they create another unverified row

andy-t-wang: App rows without an edit row are supposed to be an edge case

andy-t-wang:
For rows where there's only verified. Status will default to unverified on save because it will trigger an insert since we only have a verified row. 

We will add functionality for handling edit once verified in the submit for review PR this is just to prevent staging from having blank pages if the row is verified

pdtfh (approved):

PR #455, opened 2024-01-16
feat: Submit review workflow rebase: WID-654

andy-t-wang:
Allow the client to submit apps for review and remove them from review as well as display the status.

Permissions were updated so users can insert apps with unverified status -> needed for the upsert.
Update: User cannot update verified rows and can only set status to unverified or awaiting_review.
WID-654

https://github.com/worldcoin/developer-portal/assets/41224501/13802f43-9d23-4aab-925f-af0cd0ce807d


0xPenryn (approved):
lgtm

PR #453, opened 2024-01-16
Developer Portal | M2M Users & Teams

igorosip0v:
Closes [WID-541](https://linear.app/worldcoin/issue/WID-541/developer-portal-or-m2m-users-and-teams) Feature

PR #456, opened 2024-01-16
fix: Hasura permissions

igorosip0v:
Noticed fixes while adjusting code for Hasura changes

### DIFF
--- a/hasura/metadata/databases/default/tables/public_api_key.yaml
+++ b/hasura/metadata/databases/default/tables/public_api_key.yaml
@@ -52,6 +52,11 @@ select_permissions:
                   _eq: X-Hasura-User-Id
               - team_id:
                   _eq: X-Hasura-Team-Id
+              - _or:
+                  - role:
+                      _eq: OWNER
+                  - role:
+                      _eq: ADMIN
 update_permissions:
   - role: service
     permission:
--- a/hasura/metadata/databases/default/tables/public_api_key.yaml
+++ b/hasura/metadata/databases/default/tables/public_api_key.yaml
@@ -9,16 +9,14 @@ insert_permissions:
   - role: user
     permission:
       check:
-        _and:
-          - team:
-              memberships:
-                user_id:
+        team:
+          memberships:
+            _and:
+              - user_id:
                   _eq: X-Hasura-User-Id
-          - team_id:
-              _eq: X-Hasura-Team-Id
-          - team:
-              memberships:
-                role:
+              - team_id:
+                  _eq: X-Hasura-Team-Id
+              - role:
                   _eq: OWNER
       set:
         team_id: x-hasura-Team-Id
@@ -47,13 +45,13 @@ select_permissions:
         - team_id
         - updated_at
       filter:
-        _and:
-          - team_id:
-              _eq: X-Hasura-Team-Id
-          - team:
-              memberships:
-                user_id:
+        team:
+          memberships:
+            _and:
+              - user_id:
                   _eq: X-Hasura-User-Id
+              - team_id:
+                  _eq: X-Hasura-Team-Id
 update_permissions:
   - role: service
     permission:
@@ -83,16 +81,14 @@ delete_permissions:
   - role: user
     permission:
       filter:
-        _and:
-          - team:
-              memberships:
-                user_id:
+        team:
+          memberships:
+            _and:
+              - user_id:
                   _eq: X-Hasura-User-Id
-          - team_id:
-              _eq: X-Hasura-Team-Id
-          - team:
-              memberships:
-                role:
+              - team_id:
+                  _eq: X-Hasura-Team-Id
+              - role:
                   _eq: OWNER
 event_triggers:
   - name: generate_api_key
--- a/hasura/metadata/databases/default/tables/public_membership.yaml
+++ b/hasura/metadata/databases/default/tables/public_membership.yaml
@@ -31,13 +31,11 @@ select_permissions:
   - role: user
     permission:
       columns:
+        - id
         - role
       filter:
-        _and:
-          - user_id:
-              _eq: X-Hasura-User-Id
-          - team_id:
-              _eq: X-Hasura-User-Id
+        team_id:
+          _eq: X-Hasura-Team-Id
 update_permissions:
   - role: service
     permission:
@@ -66,5 +64,3 @@ delete_permissions:
         _and:
           - team_id:
               _eq: X-Hasura-Team-Id
-          - role:
-              _eq: OWNER
--- a/hasura/metadata/databases/default/tables/public_membership.yaml
+++ b/hasura/metadata/databases/default/tables/public_membership.yaml
@@ -33,8 +33,11 @@ select_permissions:
       columns:
         - role
       filter:
-        team_id:
-          _eq: X-Hasura-Team-Id
+        _and:
+          - user_id:
+              _eq: X-Hasura-User-Id
+          - team_id:
+              _eq: X-Hasura-User-Id
 update_permissions:
   - role: service
     permission:
--- a/hasura/metadata/databases/default/tables/public_membership.yaml
+++ b/hasura/metadata/databases/default/tables/public_membership.yaml
@@ -34,8 +34,11 @@ select_permissions:
         - id
         - role
       filter:
-        team_id:
-          _eq: X-Hasura-Team-Id
+        _or:
+          - team_id:
+              _eq: X-Hasura-Team-Id
+          - user_id:
+              _eq: X-Hasura-User-Id
 update_permissions:
   - role: service
     permission:
--- a/hasura/metadata/databases/default/tables/public_membership.yaml
+++ b/hasura/metadata/databases/default/tables/public_membership.yaml
@@ -30,7 +30,8 @@ select_permissions:
       filter: {}
   - role: user
     permission:
-      columns: []
+      columns:
+        - role
       filter:
         team_id:
           _eq: X-Hasura-Team-Id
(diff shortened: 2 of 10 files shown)


hasura/metadata/databases/default/tables/public_membership.yaml
```diff
@@ -56,3 +57,11 @@ delete_permissions:
...
     permission:
       filter: {}
+  - role: user
+    permission:
+      filter:
+        _and:
+          - team_id:
+              _eq: X-Hasura-Team-Id
+          - role:
+              _eq: OWNER
```
andy-t-wang, line 67: This means rows where team id matches and has a member that is an owner? Need the user-id here in an and with the role?

andy-t-wang (changes requested):
Also looks like some permissions still have the permission nesting issues from #453 where it's not checking the user's role specifically but just that the role exists

hasura/metadata/databases/default/tables/public_membership.yaml
igorosip0v: Any OWNER in this team, as I can see it.

hasura/metadata/databases/default/tables/public_user.yaml
```diff
@@ -67,11 +70,3 @@ update_permissions:
         id:
           _eq: X-Hasura-User-Id
       check: null
-delete_permissions:
```
pdtfh, line 70: ok for now but let's please note internally we need to add/confirm full support for: deleting your own user account; orphaned users who don't have a team (frontend handling mainly).

hasura/metadata/databases/default/tables/public_user.yaml
```diff
@@ -48,8 +50,9 @@ select_permissions:
         - is_subscribed
         - name
       filter:
-        id:
-          _eq: X-Hasura-User-Id
+        memberships:
```
pdtfh, line 53: is there a way we don't expose auth0Id and is_subscribed other than to yourself?

hasura/metadata/databases/default/tables/public_user.yaml
```diff
@@ -23,6 +23,7 @@ insert_permissions:
         - ironclad_id
         - is_subscribed
         - name
+        - team_id
```
pdtfh, line 26: is this for temporary compatibility?

hasura/metadata/databases/default/tables/public_membership.yaml
andy-t-wang: But does this check if the user submitting the delete request is the Owner? 
Think it's missing a 
```
- user_id:
     _eq: X-Hasura-User-Id
```

andy-t-wang: For some reason I keep reading this as it filters for rows where you are on the team and the role is owner. So instead of it saying you are on the team and the owner so you can delete any role on the team.
It's saying you are on the team and the role you are deleting is an owner role

andy-t-wang:
Looks good to me besides that delete permission public_membership.yaml.

hasura/metadata/databases/default/tables/public_user.yaml
igorosip0v: @pdtfh First we need to discuss how it will look like on the UI side. Because now we will have two different processes:
1. Removing user from the team (Deleting `membership` table row)
2. Deleting user from the Dev Portal completely (Deleting `user` table row)

hasura/metadata/databases/default/tables/public_api_key.yaml
```diff
@@ -47,13 +45,13 @@ select_permissions:
...
-              _eq: X-Hasura-Team-Id
-          - team:
-              memberships:
-                user_id:
+        team:
+          memberships:
+            _and:
+              - user_id:
                   _eq: X-Hasura-User-Id
+              - team_id:
```
pdtfh, line 53: let's only let ADMIN & OWNER read API keys

hasura/metadata/databases/default/tables/public_membership.yaml
```diff
@@ -56,3 +60,11 @@ delete_permissions:
   - role: service
     permission:
       filter: {}
+  - role: user
```
pdtfh, line 63: can be in a separate PR, but we should let ADMINs change all the way to ADMIN (including downgrading)

pdtfh: same for deletion below

pdtfh (approved):

hasura/metadata/databases/default/tables/public_user.yaml
pdtfh: 👍 
1. Should be from the list of teammates.
2. Should only be done by each person, can be done from profile settings. We also need to handle the case of you logging in and not belonging to any team. You should be able to create a new team or delete your account. (CC @andy-t-wang to make sure it's covered in design).

hasura/metadata/databases/default/tables/public_membership.yaml
pdtfh: I think you're right @andy-t-wang, this seems to be looking at the role in the row, not the role of the user

igorosip0v: > It's saying you are on the team and the role you are deleting is an owner role
Just checked, seems you are right. But the problem is that if we will add check for `X-Hasura-User-Id` it will allow remove only the rows with you as member.

In this case I think we need to leave only 
```
- team_id
  _eq: X-Hasura-Team-Id
 ```

And then just show Remove button only for the ADMIN and OWNER roles from the code. This is already implemented in #457.

Comment (igorosip0v) [reply hasura/metadata/databases/default/tables/public_membership.yaml]: @pdtfh @andy-t-wang Please check [this](https://github.com/worldcoin/developer-portal/pull/456/commits/a61ac17852622c5b5cd639cd8a69483e1de49589) commit

File: hasura/metadata/databases/default/tables/public_user.yaml (hunk 3 above)
Comment (igorosip0v) [reply hasura/metadata/databases/default/tables/public_user.yaml:53]: Not sure about this with the single `user` hasura role.

File: hasura/metadata/databases/default/tables/public_team.yaml [hunk 7]
```diff
@@ -56,41 +56,36 @@ select_permissions:
...
-              team_id:
-                _eq: X-Hasura-Team-Id
-          - memberships:
-              user_id:
+        memberships:
+          _and:
+            - user_id:
                 _eq: X-Hasura-User-Id
+            - team_id:
+                _eq: X-Hasura-Team-Id
```
igorosip0v, line 64: @andy-t-wang @pdtfh Just found out that we need to remove this team_id check from select row check in team table for user role. Otherwise, we can't pick a list of teams for teams selector.

I think it's ok, user should be able select all the teams related with his x-hasura-user-id. Or am I missing something? 

I will push this change when I finish team selector

hasura/metadata/databases/default/tables/public_user.yaml
pdtfh: ok let's track this in a Linear issue to revisit

hasura/metadata/databases/default/tables/public_team.yaml
pdtfh: yes, aligned on this

hasura/metadata/databases/default/tables/public_membership.yaml
pdtfh: I don't think this is the best approach, permissions as they currently are would allow any teammate to delete a membership record. for update permissions, I believe we're still letting any team member update any record that belongs to an OWNER, regardless of the permission level of the user executing.

In addition, do we plan to validate the `x-hasura-team-id` in the `api/v1/graphql` endpoint ? because in other models we rely on permissions to validate this, but here you could have an incorrect value and bypass permissions

hasura/metadata/databases/default/tables/public_membership.yaml
```diff
@@ -30,10 +30,14 @@ select_permissions:
       filter: {}
   - role: user
     permission:
-      columns: []
+      columns:
+        - role
```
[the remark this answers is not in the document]
igorosip0v: Yes. Yes

hasura/metadata/databases/default/tables/public_user.yaml
igorosip0v: [WID-657](https://linear.app/worldcoin/issue/WID-657/make-auth0id-and-is-subscribed-fields-visible-only-for-yourself)

hasura/metadata/databases/default/tables/public_api_key.yaml
igorosip0v: Done

hasura/metadata/databases/default/tables/public_membership.yaml
igorosip0v: > I don't think this is the best approach, permissions as they currently are would allow any teammate to delete a membership record. 
How MEMBER can do this if he has nothing to click to invoke deletion? 

As additional measures I can add checks for roles: `{_or: [{role: {_eq: MEMBER}, {role: {_eq: ADMIN}}]}`. This way there is no possibility to remove OWNER.

> In addition, do we plan to validate the x-hasura-team-id in the api/v1/graphql endpoint ?
Can you clarify?

hasura/metadata/databases/default/tables/public_user.yaml
igorosip0v: > Should only be done by each person

Updated. Add check `{id: {_eq: X-Hasura-User-Id}}`

PR #461, opened 2024-01-17
fix: 500 error on session expiry

0xPenryn:
Fixes the 500 error users would receive when their login sessions expired by ensuring Hasura user data is included. Tested locally.

**Walkthrough of the Issue:**
Let's say a logged-in user has the page `/app/app_6efd0...` open in their browser. While they step away, the `appSession` cookie expires. They re-open the tab, and `nextjs-auth0`'s `withPageAuthRequired()` helper is called, realizes the user no longer has a session cookie, and redirects them to `/api/auth/login?returnTo=%2Fapp%2Fapp_6efd0...`. They go through the login flow (usually without interaction, as the session on Auth0's page is valid for a longer time), and are redirected back to the page they originally requested.

The issue lies in that `returnTo` query parameter -- it would override the returnTo value that we set in `pages/api/auth/[auth0].ts`:

```typescript
export default handleAuth({
  login: (req: NextApiRequest, res: NextApiResponse) => {
    const invite_id = req.query.invite_id;

    return handleLogin(req, res, {
      returnTo: invite_id
        ? `/api/auth/login-callback?invite_id=${invite_id}`
        : "/api/auth/login-callback",
    });
  },
// ...
``` 

`login-callback` handles coordination of data between Hasura and Auth0, including the crucial step of adding the user's `id` and `team_id` from Hasura into the current session, so that they can be used to do things like retrieve a list of that user's applications. 

When that `returnTo` value was overridden and `login-callback` was bypassed, the user's Hasura `id` and `team_id` were not added to the session, causing 500 errors when pages attempted to query applications by `user_id`, without a `user_id`.

**The Fix:**
Manually configure every authenticated page's `withPageAuthRequired` helper to return to `/api/auth/login-callback` after the user has logged back in,  rather than the current page.

**Drawbacks:**
This does come with a downside -- if a user was at `/app/app_6efd0.../actions`, they will always be redirected to the first app for that user in Hasura after logging back in. 

This could be remedied by merging `nextjs-auth0`'s `handleCallback` and the `login-callback` handler into a single function handler for `/api/auth/callback` instead of manually setting `returnTo` as done in this PR, however I elected against this approach as it could potentially impact team invitation functionality.

### DIFF
--- a/web/src/pages/app/[app_id]/actions.tsx
+++ b/web/src/pages/app/[app_id]/actions.tsx
@@ -4,4 +4,6 @@ import { withPageAuthRequired } from "@auth0/nextjs-auth0";
 
 export default Actions;
 
-export const getServerSideProps: GetServerSideProps = withPageAuthRequired();
+export const getServerSideProps: GetServerSideProps = withPageAuthRequired({
+    returnTo: "/api/auth/login-callback",
+});
--- a/web/src/pages/app/[app_id]/actions.tsx
+++ b/web/src/pages/app/[app_id]/actions.tsx
@@ -5,5 +5,5 @@ import { withPageAuthRequired } from "@auth0/nextjs-auth0";
 export default Actions;
 
 export const getServerSideProps: GetServerSideProps = withPageAuthRequired({
-    returnTo: "/api/auth/login-callback",
+  returnTo: "/api/auth/login-callback",
 });
--- a/web/src/pages/app/[app_id]/debugger.tsx
+++ b/web/src/pages/app/[app_id]/debugger.tsx
@@ -5,5 +5,5 @@ import { withPageAuthRequired } from "@auth0/nextjs-auth0";
 export default Debugger;
 
 export const getServerSideProps: GetServerSideProps = withPageAuthRequired({
-    returnTo: "/api/auth/login-callback",
+  returnTo: "/api/auth/login-callback",
 });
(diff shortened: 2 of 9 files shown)


pdtfh:
@igorosip0v can you help review this? you have the best context. as a side note, we should sync expiration times between Auth0 and the Developer Portal.

igorosip0v:
@0xPenryn 
>  longer has a session cookie, and redirects them to /api/auth/login?returnTo=%2Fapp%2Fapp_6efd0....
First where we get after this redirect is the login function where we handle `invite_id`. In this function we getting `req` that we passing further to the `handleLogin` function from `@auth0/nextjs-auth0`.

> The issue lies in that returnTo query parameter -- it would override the returnTo value that we set in pages/api/auth/[auth0].ts
I assume `handleLogin` checks `req.query.returnTo` and if we already have it not apply value from the 3rd param. 

Maybe we can do something like this: 

```ts
  login: (req: NextApiRequest, res: NextApiResponse) => {
    const invite_id = req.query.invite_id;
    delete req.query.returnTo

    return handleLogin(req, res, {
      returnTo: invite_id
        ? `/api/auth/login-callback?invite_id=${invite_id}`
        : "/api/auth/login-callback",
    });
  },
```

I think we don't need automatical returnTo from `@auth0/nextjs-auth0`, so we can delete/overwrite it somehow.

0xPenryn:
@igorosip0v should we make your suggested change in #474 and close this PR?

PR #462, opened 2024-01-18
feat: Img backend

andy-t-wang:
Backend updates for image upload workflow. Front end needs clean up but is currently a draft pr in img-upload-dev if you want to see how that connects. #458 (most of it is implemented just needs some clean up and testing)

Also this currently accepts PNG only since we need to make a separate post policy if we want to exactly enforce content type. I figure this is fine since PNG accommodates transparent bg, but let me know if you want to also accept jpeg. 

This pr ended up being quite large. 
1. Migrated to hasura custom actions
2. Update Precheck
3. New Tests

Might be worth holding off on adding delete images until we have permissions merged
complete WID-658
Merge #463 first

### Endpoints Overview
1. delete_images: Used to delete all unverified images when the user deletes the app
4. get_images: Gets a single image that the user specifies. Used after the upload so the user can verify the image was uploaded correctly
5. get_unverified: Checks the unverified images available on the app using the app_id and returns them. Used when the user loads an app with existing images
6. upload_image: Used when the user uploads an image.


hasura/metadata/databases/default/tables/public_app_metadata.yaml
andy-t-wang, line 1: Service needs to fetch unverified app images

web/next.config.js
```diff
@@ -4,7 +4,8 @@
 // @ts-ignore
 const nextSafe = require("next-safe");
 const isDev = process.env.NODE_ENV !== "production";
-
+const s3BucketUrl = process.env.AWS_S3_BUCKET_URL;
```
pdtfh, line 7: S3 Bucket URL should be constructed from bucket name, `ASSETS_S3_BUCKET_NAME`

web/next.config.js
```diff
@@ -4,7 +4,8 @@
...
+const cdnUrl = process.env.AWS_CDN_URL;
```
pdtfh, line 8:
```suggestion
const cdnUrl = process.env.ASSETS_CDN_URL;
```

web/next.config.js
```diff
@@ -44,7 +48,7 @@ const nextConfig = {
 
   reactStrictMode: true,
   images: {
-    domains: ["world-id-public.s3.amazonaws.com"],
```
pdtfh, line 47: let's add a note that `world-id-public.s3.amazonaws.com` is deprecated and should be removed soon

web/next.config.js
```diff
@@ -17,8 +18,10 @@ const nextConfig = {
             mergeDefaultDirectives: true,
             "img-src": [
               "'self'",
+              "blob:", // Used to enforce image width and height
```
pdtfh, line 21: review with security team to understand any potential concerns

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,106 @@
...
+  async (req: NextApiRequest, res: NextApiResponse<ImageGetResponse>) => {
+    try {
+      if (!req.method || req.method !== "GET") {
+        return errorNotAllowed(req.method, res, req);
+      }
+      const { app_id, image_type } = req.query;
+
+      const { isValid, parsedParams, handleError } =
+        await validateRequestSchema({
+          value: { app_id, image_type },
```
pdtfh, line 45: couldn't you pass directly req.query here and avoid the additional destructure above?

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,106 @@
...
+          schema,
+        });
+
+      if (!isValid || !parsedParams) {
+        return handleError(req, res);
+      }
+      const session = (await getSession(req, res)) as Session;
+      const auth0Team = session?.user.hasura.team_id;
```
pdtfh, line 53: I don't think this is compatible with #453, we'll be pulling the team_id from the URL path, and API routes from a header, correct @igorosip0v ?

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      const client = await getAPIServiceGraphqlClient();
+      const { team: userTeam } = await checkUserInAppDocumentSDK(
+        client
+      ).CheckUserInApp({
+        team_id: auth0Team,
+        app_id: app_id as string,
+      });
+      if (!userTeam[0].apps.some((app) => app.id === app_id)) {
+        return errorHasuraQuery({
```
pdtfh, line 63: if we're using these errors, then this API routes should be exposed through Hasura custom actions (which I think may be a good idea) instead of as direct API routes. thoughts @igorosip0v ?

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,106 @@
...
+        });
+
+      if (!isValid || !parsedParams) {
+        return handleError(req, res);
+      }
+      const session = (await getSession(req, res)) as Session;
+      const auth0Team = session?.user.hasura.team_id;
+
+      const client = await getAPIServiceGraphqlClient();
+      const { team: userTeam } = await checkUserInAppDocumentSDK(
```
pdtfh, line 56: reminder for @pdtfh, re-review permissions conditions after rest of feedback is addressed

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+      const bucketName = process.env.AWS_BUCKET_NAME;
+      const objectKey = `unverified/${app_id}/${image_type}.png`;
+
+      const command = new GetObjectCommand({
+        Bucket: bucketName,
+        Key: objectKey,
+      });
+      const signedUrl = await getSignedUrl(s3Client, command, {
+        expiresIn: 7200, // The URL will expire in 2 hours
```
pdtfh, line 87: rationale? this should be short use right? maybe 15 mins?

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,106 @@
...
+      });
+
+      res.status(200).json({
+        url: signedUrl,
+        message: "Success",
```
pdtfh, line 92:
```suggestion
        success: true,
```
following convention on other endpoints

web/src/api/images/get_uploaded_image.ts
andy-t-wang: I put req.query there but still destructured it after since I use both the values in the code later. Probably slightly cleaner now

web/src/api/images/get_uploaded_image.ts
igorosip0v: Yes, but while we don't have these changes in this PR we can leave as it is. Or one thing that we can do here is accept team id from header/body/param and pass `session?.user.hasura.team_id` for now, but change to `router.query.team_id` later

web/src/api/images/get_uploaded_image.ts
igorosip0v: Mostly it will be just a wrapper, but this will give us the possibility to generate hooks to gql query and then use it on the frontend side, so we can do this.

pdtfh: as aligned on stand up, let's convert these endpoints to GraphQL custom actions

web/src/api/images/delete_all_images.ts
```diff
@@ -0,0 +1,121 @@
...
+  success?: boolean;
+  message?: string;
+};
+
+const schema = yup.object({
+  app_id: yup.string().strict().required(),
+});
+
+export type DeleteAllImagesBody = yup.InferType<typeof schema>;
+// This endpoint takes in an App ID and deletes all the images in the folder. Used if the user deletes the app.
```
pdtfh, line 29: 1. how does a user then delete a single image? 
2. this should be an internal call, triggered from a delete hook from Hasura. also just in case, might be best to expire the images and let's add a lifecycle rule to the bucket (infra) to delete all expired objects after 2-3 days.

web/src/api/images/upload_image.ts
```diff
@@ -0,0 +1,113 @@
...
+      "showcase_img_3",
+    ])
+    .required(),
+});
+
+export type ImageUploadBody = yup.InferType<typeof schema>;
+// This endpoint takes an AppID and checks the user is on the team. If so it returns a signed URL to upload an image to S3.
+export const handleImageUpload = withApiAuthRequired(
+  async (req: NextApiRequest, res: NextApiResponse<ImageUploadResponse>) => {
+    try {
```
pdtfh, line 40: FYI this would normally require CORS but given you're changing it to Hasura custom actions we should be fine

web/src/api/images/upload_image.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+      const s3Client = new S3Client({
+        region: process.env.ASSETS_S3_REGION,
+      });
+      if (!process.env.ASSETS_S3_BUCKET_NAME) {
+        throw new Error("AWS Bucket Name must be set.");
+      }
+      const bucketName = process.env.ASSETS_S3_BUCKET_NAME;
+      const objectKey = `unverified/${app_id}/${image_type}.png`;
+      const contentType = "image/png";
```
pdtfh, line 86: worth adding support for jpg. if images don't have background generally this is a more effective format

pdtfh (changes requested):
good progress, main change is converting to custom GraphQL actions, comments inline

web/src/api/images/delete_all_images.ts
andy-t-wang: I was mainly using this to clean up old resources if the user deleted the app. For verified images they will overwrite the old image to replace it and can't just delete an image. For unverified, the bucket path is protected and the hasura is updated to not have an img_url

andy-t-wang: Figure not clearing the image would accommodate for the use case where they click the x on the image but then don't save the changes

andy-t-wang: Aligned in slack pushing this to another version

web/src/api/images/get_uploaded_image.ts
andy-t-wang: ok I am now using the session variables with the hasura actions is that secure?

web/src/api/images/get_unverified_images.ts
```diff
@@ -0,0 +1,147 @@
...
+  };
+  success?: boolean;
+};
+
+const schema = yup.object({
+  app_id: yup.string().strict().required(),
+});
+
+export type ImageGetAllUnverifiedImagesBody = yup.InferType<typeof schema>;
+// This endpoint takes in an App ID and returns all available unverified images for that app.
```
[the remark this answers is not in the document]
andy-t-wang: Actually when you upload an image but don't save yet this won't be able to pull it. We also don't know what extension to manually get. Feel like at that point it might be easier to separate the logic. Would require us to then have get_all_unverified generate signed URLs even if an image doesn't exist there.

web/next.config.js
```diff
@@ -4,7 +4,8 @@
 // @ts-ignore
 const nextSafe = require("next-safe");
 const isDev = process.env.NODE_ENV !== "production";
-
+const s3BucketUrl = `https://s3.${process.env.ASSETS_S3_REGION}.amazonaws.com/${process.env.ASSETS_S3_BUCKET_NAME}`;
+const cdnUrl = process.env.ASSETS_CDN_URL;
```
andy-t-wang, line 8: @pdtfh It is unclear to me if we need to have region name in prod, but on my local dev I did need it.

web/src/api/images/delete_all_images.ts
```diff
@@ -0,0 +1,159 @@
...
+  ListObjectsCommand,
+  PutObjectTaggingCommand,
+} from "@aws-sdk/client-s3";
+import { logger } from "@/lib/logger";
+
+export type DeleteAllImagesResponse = {
+  success?: boolean;
+};
+
+/**
```
andy-t-wang, line 21: Think this makes sense to implement the actual trigger in a seperate pr from the image upload process. Just leaving the endpoint here for now.

andy-t-wang:
Note: Since your last review I updated to hasura custom actions and added a img validation trigger back to unverified apps

hasura/migrations/default/1705775389481_turn_on_img_trigger/up.sql
```diff
@@ -0,0 +1,43 @@
...
+      END LOOP;
+    END IF;
+
+  END IF; -- End of verification_status check
+
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER "validate_img_uris" BEFORE INSERT
```
pdtfh, line 41: likely also want to drop validate_all_img_urls_format?

web/next.config.js
andy-t-wang: resolved with security

web/src/api/images/delete_all_images.ts
```diff
@@ -0,0 +1,159 @@
...
+    if (objectKeys && objectKeys.length > 0) {
+      const expirePromises = objectKeys.map((key) =>
+        s3Client.send(
+          new PutObjectTaggingCommand({
+            Bucket: bucketName,
+            Key: key,
+            Tagging: {
+              TagSet: [
+                {
+                  Key: "ExpireAfter",
```
pdtfh, line 135:
```suggestion
                  Key: "expired",
```

web/src/api/images/delete_all_images.ts
```diff
@@ -0,0 +1,159 @@
...
+                  Value: "3Days",
```
pdtfh, line 136:
```suggestion
                  Value: "true",
```

web/src/api/images/get_unverified_images.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (body?.action.name !== "get_all_unverified_images") {
+      return errorHasuraQuery({
+        res,
+        req,
+        detail: "Invalid action.",
+        code: "invalid_action",
+      });
+    }
+
+    if (body.session_variables["x-hasura-role"] === "admin") {
```
pdtfh, line 48: ok admin

web/src/api/images/get_unverified_images.ts
```diff
@@ -0,0 +1,176 @@
...
+    ).GetUnverifiedImages({
+      team_id: teamId,
+      app_id: app_id as string,
+      user_id: userId,
+    });
+    if (appInfo.length === 0 || appInfo[0].app_metadata.length === 0) {
+      return errorHasuraQuery({
+        res,
+        req,
+        detail: "User does not have access to this app.",
```
pdtfh, line 97: In general is best to return more generic errors for permissions, e.g. "App not found"

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (body?.action.name !== "get_uploaded_image") {
+      return errorHasuraQuery({
+        res,
+        req,
+        detail: "Invalid action.",
+        code: "invalid_action",
+      });
+    }
+
+    if (body.session_variables["x-hasura-role"] === "admin") {
```
pdtfh, line 49: same

web/src/api/images/get_uploaded_image.ts
```diff
@@ -0,0 +1,142 @@
...
+import { getAPIServiceGraphqlClient } from "src/backend/graphql";
+import { getSdk as checkUserInAppDocumentSDK } from "@/api/images/graphql/checkUserInApp.generated";
+import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
+import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
+import { logger } from "@/lib/logger";
+
+type RequestQueryParams = {
+  app_id: string;
+  image_type: string;
+  content_type_ending: string;
```
pdtfh, line 18: wouldn't we want to get this automatically from Hasura? seems like it would not be great UX otherwise

web/src/api/images/upload_image_url.ts
```diff
@@ -0,0 +1,141 @@
...
+    });
+    if (!process.env.ASSETS_S3_BUCKET_NAME) {
+      throw new Error("AWS Bucket Name must be set.");
+    }
+    const bucketName = process.env.ASSETS_S3_BUCKET_NAME;
+    // Standardize JPEG to jpg
+    const objectKey = `unverified/${app_id}/${image_type}.${
+      content_type_ending === "jpeg" ? "jpg" : content_type_ending
+    }`;
+    const contentType = `image/${content_type_ending}`;
```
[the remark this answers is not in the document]
andy-t-wang: yes it is the current interaction is using jpeg from the front end. Will update to have type checking

hasura/metadata/actions.yaml
```diff
@@ -1,8 +1,58 @@
...
+          value_from_env: INTERNAL_ENDPOINTS_SECRET
+      request_transform:
+        method: GET
+        query_params:
+          app_id: "{{$body.input.app_id}}"
+        request_headers:
+          add_headers: {}
+          remove_headers:
+            - content-type
+        template_engine: Kriti
```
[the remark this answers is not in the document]
andy-t-wang: Was the one automatically being used when I made the action in the console

hasura/metadata/actions.yaml
```diff
@@ -1,8 +1,58 @@
...
+          image_type: "{{$body.input.image_type}}"
+        request_headers:
+          add_headers: {}
+          remove_headers:
+            - content-type
+        template_engine: Kriti
+        version: 2
+    permissions:
+      - role: api_key
+      - role: user
```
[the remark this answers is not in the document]
andy-t-wang: does not appear to be an option

hasura/metadata/actions.yaml
```diff
@@ -1,8 +1,58 @@
 actions:
+  - name: delete_unverified_images
+    definition:
+      kind: synchronous
+      handler: "{{NEXT_API_URL}}/images/delete_images"
+      headers:
```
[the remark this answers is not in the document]
andy-t-wang: Yeah originally didn't include it since it's not implemented, but added user and api_key

andy-t-wang:
Can't find the comment anymore but within next.config we do need to use the hostname domain part not the full url. The image won't load if we do 
```
 {
        protocol: "https",
        hostname: "s3.us-east-2.amazonaws.com/developer-portal-2.0",
      },
```

web/next.config.js
andy-t-wang: This does not work with the presigned url domain returned by presignedpost

web/src/api/images/graphql/checkUserInApp.graphql
```diff
@@ -0,0 +1,10 @@
+query CheckUserInApp($team_id: String!, $app_id: String!, $user_id: String!) {
```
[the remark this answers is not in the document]
andy-t-wang: combined into a single query

web/src/api/images/upload_image_url.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (!app_id || !image_type || !content_type_ending) {
+      return errorHasuraQuery({
+        res,
+        req,
+        detail: "app_id, image_type, and content_type_ending must be set.",
+        code: "required",
+      });
+    }
+    // Check that content_type_ending is png or jpeg
+    if (!["png", "jpeg"].includes(content_type_ending)) {
```
andy-t-wang, line 74: We enforce jpeg on the front end by extracting form the file.type

andy-t-wang:
Notes: Before Merging
1. Will update next.config to use a path name
2. Remove web-app-router .env

web/src/api/images/upload_image_url.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (
+      userTeam.length === 0 ||
+      !userMembership ||
+      userMembership.role === "MEMBER"
+    ) {
+      return errorHasuraQuery({
+        res,
+        req,
+        detail: "App not found.",
+        code: "no_access",
```
pdtfh, line 104:
```suggestion
        code: "not_found",
```

web/src/api/images/delete_all_images.ts
```diff
@@ -0,0 +1,156 @@
...
+    const userMembership = appInfo[0].team.memberships.find(
+      (membership) => membership.user_id === userId
+    );
+    // There should only be one app with the matching app_id and team_id
+    // Only Owner is allowed to delete
+    if (
+      appInfo.length === 0 ||
+      appInfo[0].app_metadata.length === 0 ||
+      !userMembership ||
+      userMembership.role !== "OWNER"
```
pdtfh, line 93: shouldn't admin also be able to do this?

pdtfh, line 93: another minor improvement, you could already account for roles in the query

pdtfh (approved):

web/src/api/images/delete_all_images.ts
andy-t-wang: This deletes all images and should only be used if the app is being deleted, so admin users. We will make a separate endpoint to delete individual images.

andy-t-wang: This query is used by getunverified images too. I added a roles check to the other query though

PR #468, opened 2024-01-22
M2m test

igorosip0v:


hasura/seeds/default/1676722071560_teamSeed.sql
```diff
@@ -1,2 +1,3 @@
 SET check_function_bodies = false;
 INSERT INTO public.team (id, name, created_at, updated_at) VALUES ('team_d7cde14f17eda7e0ededba7ded6b4467', 'Testing Co', '2023-02-18 10:44:27.212563+00', '2023-02-18 10:44:27.212563+00');
+INSERT INTO public.team (id, name, created_at, updated_at) VALUES ('team_d7cde14f17eda7e0ededba7ded6b4468', 'Testing Co', '2024-01-01 10:44:27.212563+00', '2024-01-01 10:44:27.212563+00');
```
pdtfh, line 3:
```suggestion
INSERT INTO public.team (id, name, created_at, updated_at) VALUES ('team_2222214f17eda7e0ededba7ded6b4222', 'Testing Co #2', '2024-01-01 10:44:27.212563+00', '2024-01-01 10:44:27.212563+00');
```
minor thing to make it clearer

hasura/seeds/default/1676722088705_userSeed.sql
```diff
@@ -2,3 +2,6 @@ SET check_function_bodies = false;
...
 INSERT INTO public."user" (id, created_at, updated_at, email, team_id, world_id_nullifier, is_subscribed, ironclad_id) VALUES ('usr_a78f59e547fa5bd3d76bc1a1817c6d90', '2023-02-19 10:44:58.085661+00', '2023-02-19 10:44:58.085661+00', 'test-admin@worldcoin.org', 'team_d7cde14f17eda7e0ededba7ded6b4467', '0x1234', false, 'ironclad_1234');
 INSERT INTO public."user" (id, created_at, updated_at, email, team_id, world_id_nullifier, is_subscribed, ironclad_id) VALUES ('usr_a78f59e547fa5bd3d76bc1a1817c6d91', '2023-02-20 10:44:58.085661+00', '2023-02-20 10:44:58.085661+00', 'test-member@worldcoin.org', 'team_d7cde14f17eda7e0ededba7ded6b4467', '0x12345', false, 'ironclad_12345');
+INSERT INTO public."user" (id, created_at, updated_at, email, team_id, world_id_nullifier, is_subscribed, ironclad_id) VALUES ('usr_a78f59e547fa5bd3d76bc1a1817c6d92', '2023-02-18 10:44:58.085661+00', '2023-02-18 10:44:58.085661+00', 'test1@worldcoin.org', 'team_d7cde14f17eda7e0ededba7ded6b4468', '0x123456', false, 'ironclad_123456');
```
pdtfh, line 5:
```suggestion
INSERT INTO public."user" (id, created_at, updated_at, email, team_id, world_id_nullifier, is_subscribed, ironclad_id) VALUES ('usr_a78f59e547fa5bd3d76bc1a1817c6d92', '2023-02-18 10:44:58.085661+00', '2023-02-18 10:44:58.085661+00', 'test1@team2.example.com', 'team_d7cde14f17eda7e0ededba7ded6b4468', '0x123456', false, 'ironclad_123456');
```
also to make it extra clear it's from a different team

hasura/seeds/default/1678146259008_appSeed.sql
```diff
@@ -207,4 +208,217 @@ SELECT
...
+SELECT
+    id,
+    'Multi-claim App',
+    'This app has a multi-claim custom action',
+    '',
+    'unverified'
+FROM
+    inserted_app7;
+
+WITH
```
pdtfh, line 299: we can probably do with just 2 apps for this additional team?

web/tests/integration/action.test.ts
```diff
@@ -44,7 +44,7 @@ describe("service role", () => {
       }
     }
 
-    expect(signInWithWorldIDCount).toEqual(4); // 4 apps are created (non-archived)
+    expect(signInWithWorldIDCount).toEqual(8); // 4 apps are created (non-archived)
```
pdtfh, line 47:
```suggestion
    expect(signInWithWorldIDCount).toEqual(8); // 8 apps are created (non-archived)
```

web/tests/integration/team.test.ts
```diff
@@ -11,3 +18,128 @@ describe("team model", () => {
...
+      if (teamMember.role === Role_Enum.Owner) {
+        expect(response.data?.update_team?.affected_rows).toEqual(1);
+
+        // NOTE: revert update changes
+        await integrationDBExecuteQuery(
+          // update team hasura postgresql query
+          `UPDATE "public"."team" SET "name" = 'new name' WHERE "id" = '${team_id}' RETURNING "name";`
+        );
+      } else {
+        expect(response.data?.update_team?.affected_rows).toEqual(0);
```
pdtfh, line 73: we could additionally execute a query here to check the name was indeed not changed in the DB

web/tests/integration/team.test.ts
```diff
@@ -11,3 +18,128 @@ describe("team model", () => {
...
+      }
+    }
+  });
+
+  test("cannot select another team", async () => {
```
pdtfh, line 78: **blocking**. one important case we're not covering here because this is calling Hasura GraphQL directly is using the `/v1/graphql` endpoint from Next.js, particularly because anyone can tamper with the team ID header

web/tests/integration/user.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const response = await client.mutate({
+      mutation,
+      variables: {
+        id: userFromTeam2?.user_id,
+        name: "new name",
+      },
+    });
+
+    expect(response.data.update_user_by_pk).toEqual(null);
```
pdtfh, line 64: worth doing an extra SQL query and making sure there were no effected updates?

web/tests/integration/app.test.ts
```diff
@@ -0,0 +1,122 @@
...
+} from "./setup";
+
+import { getAPIUserClient } from "./test-utils";
+
+// TODO: Consider moving this to a generalized jest environment
+beforeEach(integrationDBSetup);
+beforeEach(integrationDBTearDown);
+
+describe("user role", () => {
+  test("cannot select another team apps", async () => {
```
pdtfh, line 16:
```suggestion
  test("cannot select another team's apps", async () => {
```

web/tests/integration/app.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  integrationDBTearDown,
+} from "./setup";
+
+import { getAPIUserClient } from "./test-utils";
+
+// TODO: Consider moving this to a generalized jest environment
+beforeEach(integrationDBSetup);
+beforeEach(integrationDBTearDown);
+
+describe("user role", () => {
```
pdtfh, line 15: same as with team, ideally we would do this through the next.js GraphQL endpoint, to make sure tampering with the team ID header doesn't do anything

pdtfh (changes requested):
comments inline, blocking comments have a specific note

web/tests/integration/team.test.ts
andy-t-wang: Alternatively the test already checks that a user not part of the team cannot update

web/tests/integration/app.test.ts
andy-t-wang: we test an invalid team being passed in now

web/tests/integration/team.test.ts
```diff
@@ -11,3 +21,204 @@ describe("team model", () => {
...
+            id
+          }
+        }
+      `;
+
+      const response = await client.query({ query });
+      expect(response.data.team.length).toEqual(1);
+      expect(response.data.team[0].id).toEqual(team.id);
+    }
+    // Test using an team that the user is not a member of
```
pdtfh, line 113: this should be a different test case

web/tests/integration/team.test.ts
```diff
@@ -11,3 +21,204 @@ describe("team model", () => {
...
+    const { rows: testInvalidTeamMemberships } =
+      (await integrationDBExecuteQuery(
+        `SELECT id, user_id FROM "public"."membership" WHERE "team_id" = '${teams[1].id}' limit 1;`
+      )) as { rows: Array<{ id: string; user_id: string }> };
+
+    const testInvalidClient = await getAPIUserClient({
```
pdtfh, line 119: let's add a note that we want to later add test cases that call the internal /v1/graphql route from Next

web/tests/integration/team.test.ts
```diff
@@ -11,3 +21,204 @@ describe("team model", () => {
...
+      team_id: teams[0].id,
```
pdtfh, line 120: nit: this test case could be simplified by just setting `team_id` to teams[1].id and removing the query above

web/tests/integration/team.test.ts
```diff
@@ -11,3 +21,204 @@ describe("team model", () => {
...
+          id
+        }
+      }
+    `;
+
+    const response = await testInvalidClient.query({ query });
+    expect(response.data.team.length).toEqual(0);
+  });
+
+  test("API Key: cannot select another team", async () => {
```
pdtfh, line 136: to follow the convention above, this should be a `describe("api_key role", () => {`

web/tests/integration/team.test.ts
```diff
@@ -11,3 +21,204 @@ describe("team model", () => {
...
+    const { rows: teams } = (await integrationDBExecuteQuery(
+      `SELECT id FROM "public"."team"`
+    )) as { rows: Array<{ id: string }> };
+
+    for (const team of teams) {
+      const response = await fetch(
+        publicRuntimeConfig.NEXT_PUBLIC_GRAPHQL_API_URL,
```
pdtfh, line 143: `NEXT_PUBLIC_GRAPHQL_API_URL` is the Hasura URL, so there's no benefit to not using the client we use above

web/tests/integration/team.test.ts
```diff
@@ -11,3 +21,204 @@ describe("team model", () => {
...
+
+    expect(response.data.update_team.affected_rows).toEqual(0);
+    // Test using an team that the user is not a member of
+
+    const { rows: testInvalidTeamMemberships } =
+      (await integrationDBExecuteQuery(
+        `SELECT id, user_id FROM "public"."membership" WHERE "team_id" = '${teams[1].id}' limit 1;`
+      )) as { rows: Array<{ id: string; user_id: string }> };
+
+    const testInvalidClient = await getAPIUserClient({
```
pdtfh, line 211: same comments as test case above, simplify test case & split it

web/tests/integration/user.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    const { rows: userFromTeam2AfterUpdate } = (await integrationDBExecuteQuery(
+      `SELECT name FROM "public"."user" WHERE id = '${userFromTeam2?.user_id}'`
+    )) as { rows: Array<{ name: string }> };
+    expect(userFromTeam2AfterUpdate[0].name).not.toBe("new name");
+
+    const response2 = await client.mutate({
```
pdtfh, line 76: here we're still missing the test case that even a team owner in the same team, cannot update another user. only the same user can update themselves

web/tests/integration/app.test.ts
```diff
@@ -0,0 +1,302 @@
...
+      const response = await client.query({ query });
+      expect(response.data.app.length).toEqual(teamApps.length);
+
+      response.data.app.forEach((app: { id: string; team_id: string }) => {
+        expect(app.team_id).toEqual(team.id);
+        expect(teamApps.map((app) => app.id)).toContain(app.id);
+      });
+    }
+    // Use an incorrect team
+    const { rows: teamMemberships } = (await integrationDBExecuteQuery(
```
pdtfh, line 53: let's create a separate test case for this

web/tests/integration/team.test.ts
pdtfh: and actually this will be safer, as technically the user ID from testInvalidTeamMemberships could also be a member of teams[0]

web/tests/integration/app.test.ts
```diff
@@ -0,0 +1,302 @@
...
+      `SELECT id, user_id FROM "public"."membership" WHERE "team_id" = '${teams[1].id}' limit 1;`
+    )) as { rows: Array<{ id: string; user_id: string }> };
+    const client = await getAPIUserClient({
+      team_id: teams[0].id,
```
pdtfh, line 57: worth creating a single test case on `app.test.ts` for what happens if you pass a completely invalid team ID

web/tests/integration/app.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+      variables: {
+        team_id: teams.find((t) => t.id !== tokenTeamId)?.id,
+      },
+    });
+
+    expect(response.data.update_app.affected_rows).toEqual(0);
+
+    // Test invalid team
+    const { rows: testInvalidTeamMemberships } =
```
pdtfh, line 149: separate test case

pdtfh (approved):
comments inline, feel free to address and merge

web/tests/integration/app.test.ts
andy-t-wang: when splitting it out will need to fetch memberships anyways

PR #471, opened 2024-01-23
WID-668: Feat: Button Component

andy-t-wang:
Need advice on how to use GT-America font?
Added the icons in as well.

Screenshot 2024-01-23 at 2 07 48 PM

Completes WID-668


web-app-router/components/Icon/index.tsx
igorosip0v, line 1: I'd say let's add only icons that we are currently using.

And I'd like to use it in another way, but this we can fix in a separate PR, I guess.

web-app-router/components/Button/index.tsx
```diff
@@ -1,7 +1,30 @@
...
+  | (CommonLinkProps & {
+      type?: never;
+      href: string;
+      disabled?: boolean;
+    });
+
+export const Button = memo(function Button(props: CommonButtonProps) {
+  if (props.href) {
+    const { href, ...restProps } = props;
+    const href_parsed = !props.disabled ? href : "";
```
igorosip0v, line 19:
```suggestion
    const hrefParsed = !props.disabled ? href : "";
```

web-app-router/app/layout.tsx
```diff
@@ -1,2 +1,4 @@
 import { RootLayout } from "@/scenes/Root/layout";
+import "./globals.css";
```
igorosip0v, line 2: @andy-t-wang Let's move it inside `RootLayout` component

web-app-router/components/DecoratedButton/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      loading: "border-error-300 text-error-400 pointer-events-none",
+    },
+  };
+  return (
+    <Button
+      disabled={disabled}
+      className={clsx(
+        className,
+        "px-6 py-2.5 rounded-xl border font-medium ",
+        !disabled && !loading && buttonStyles[variant].normal,
```
igorosip0v, line 54: @andy-t-wang you can use something like this with clsx:

```js
{
[buttonStyles.[variant].normal]: !disabled && !loading
}
```

looks better in my opinion

web-app-router/components/DecoratedButton/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const {
+    icon,
+    showArrowRight,
+    loading,
+    disabled,
+    className,
+    variant = "primary",
+    ...restProps
+  } = props;
+  const buttonStyles = {
```
igorosip0v, line 27: @andy-t-wang Lets divide multiline code with empty row.

```
} = props

const buttonStyles = {
```

Same for the row before return

igorosip0v (approved):

web-app-router/components/Icon/index.tsx
andy-t-wang: Updated buttons to use icon directly allowing override

PR #472, opened 2024-01-23
WID-669 Create Input component

andy-t-wang:
Complete WID-669
Reused some colors so I did the change on top of that.
Merge #471 first.

Also worth leaving yup and react-hook form in since we can use that to test later components and will need it in prod version.

---
Screenshot 2024-01-24 at 4 20 19 PM


web-app-router/components/Input/index.tsx
```diff
@@ -1,7 +1,99 @@
-export const Input = () => {
+"use client";
+import clsx from "clsx";
+import { FloatingLabel } from "flowbite-react";
+import { ComponentProps, memo } from "react";
+import { Icon } from "../Icon";
+
+type InputProps = ComponentProps<"input"> & {
```
igorosip0v, line 7: @andy-t-wang We need to add `register: RegisterReturn` (don't remember exact type name)  prop to make compatible with `react-hook-form`. Also it would be great to have `errors` prop to pass field error object from react-hook-form.

web-app-router/components/Input/index.tsx
```diff
@@ -1,7 +1,99 @@
...
+        theme={customTheme}
+        variant="outlined"
+        onChange={onChange}
+        sizing="md"
+        helperText={helperText}
+        placeholder={placeholder}
+        disabled={disabled}
+        required={true}
+      ></FloatingLabel>
+      {canCopy && (
```
igorosip0v, line 81: I'd prefer another approach here if this is possible. To have `addon: ReactNode` and `addonPosition: "left" | "right"` props to make it more flexible, because it's a common component.  For example if we will need to make something like eye button to hide/show password. 

I think we can implement it like this: 

```
<div className="grid grid-cols-[auto_1fr_auto]">
<div>{Boolean(addon) && addonPosition === "left" && addon}<div>
<input/>
<div>{Boolean(addon) && addonPosition === "right" && addon}<div>
</div>
```

If we are passing addon, it will take space of 1st or third grid cell, otherwise it will be 0px and only input will be visible.

And then we can use it like: 

```
<Input addon={<Component onClick={() => {something}}/> addonPosition="left"}
```

web-app-router/components/Input/index.tsx
```diff
@@ -1,7 +1,99 @@
...
+import { Icon } from "../Icon";
+
+type InputProps = ComponentProps<"input"> & {
+  type?: ComponentProps<"input">["type"];
+  disabled?: boolean;
+  helperText?: string;
+  label: string;
+  className?: string;
+  placeholder?: string;
+  required?: boolean;
```
igorosip0v, line 14: @andy-t-wang we don't need this prop in case we will use react-hook-form + yup. This will be defined by yup schema.

web-app-router/tailwind.config.ts
```diff
@@ -6,6 +6,7 @@ const config: Config = {
     "./components/**/*.{js,ts,jsx,tsx,mdx}",
     "./app/**/*.{js,ts,jsx,tsx,mdx}",
     "./scenes/**/*.{js,ts,jsx,tsx,mdx}",
+    "node_modules/flowbite-react/lib/esm/**/*.js",
```
igorosip0v, line 9: @andy-t-wang why do we need this?

igorosip0v (changes requested):

web-app-router/tailwind.config.ts
andy-t-wang: Required to use flowbite

andy-t-wang: https://www.flowbite-react.com/docs/getting-started/quickstart

web-app-router/components/Input/index.tsx
andy-t-wang: I actually think it's more comfortable to use required directly since hook form doesn't pass in the yup required field. You need to pass it in register but even that was finnicky to me. Lmk what you think I made a change in my new update

web-app-router/components/Input/index.tsx
andy-t-wang: I added the component in but requires some absolute style to do so. I think it could be better for the user to specify but lmk

web-app-router/components/Input/index.tsx
igorosip0v: I don't see why we need to duplicate validation rules on the inputs directly while we already have them in the schema

web-app-router/components/Input/index.tsx
igorosip0v: Hm, I can't say I like it. To adjust input side we need to use paddings the theme object as you did. What if addon will be a wide button that have width move than these paddings? Or what if passing an addon with more custom styles. Try to make button on a 3rd input as `flex` it breaks its position. 

My idea with the grid was that if we passing an addon it should take space by its sizes and it will automatically adjust input size. But as I can see there are no too many options to customize `FloatingLabel` in that way.

igorosip0v: Let's update input the way we discussed in Slack and then review again

web-app-router/tailwind.config.ts
```diff
@@ -51,14 +54,19 @@ const config: Config = {
             500: "#00B800",
           },
         },
+
+        gridTemplateColumns: {
+          "1fr/auto": "1fr auto",
+          "auto/1fr": "auto 1fr",
+          "auto/1fr/auto": "auto 1fr auto",
+        },
```
andy-t-wang, line 62: this is not working for me btw @igorosip0v  using grid-cols-auto/1fr/auto

andy-t-wang:
Screenshot 2024-01-25 at 1 40 31 PM

web-app-router/tailwind.config.ts
igorosip0v: it was inside color object, fixed

igorosip0v (approved):
@andy-t-wang Lgtm! One thing that maybe we want also to display errors below the input. But not sure how to handle it when we have helperText. Replace it?

PR #473, opened 2024-01-24
WID-670 Feat: Radio Component

andy-t-wang:
Creates a radio component
Screenshot 2024-01-23 at 4 45 26 PM

### DIFF
--- a/web-app-router/components/Input/index.tsx
+++ b/web-app-router/components/Input/index.tsx
@@ -33,7 +33,7 @@ export const Input = memo(function Input(props: InputInterface) {
     disabled,
     ...restProps
   } = props;
-  const { ...restRegister } = register;
+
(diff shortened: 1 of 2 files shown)


web-app-router/components/Radio/index.tsx
```diff
@@ -1,7 +1,47 @@
...
+"use client";
+
+import clsx from "clsx";
+import React, { memo } from "react";
+
+type RadioProps = {
+  id: string;
+  checked?: boolean;
+  disabled?: boolean;
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
igorosip0v, line 10: @andy-t-wang I think we don't need this prop, it should be controlled by react-hook-form

PR #474, opened 2024-01-24
[App router] configure auth0 login

igorosip0v:
Closes [WID-673](https://linear.app/worldcoin/issue/WID-673/configure-auth0-login)

Introduces login with auth0 with new app router

What is available right now: 
- You can login from the `/login` page it will finish auth0 flow and redirect to empty `/join` page
- Auth required routes protection

Notes: 
- Unfortunately, it's impossible to use `withPageAuthRequired` on `layout.tsx` with `auth0/nextjs-auth0` to protect nested routes :( 
- There is an [issue](https://github.com/auth0/nextjs-auth0/issues/1643) in the repo with the same issue I had.
- Because of this I decided to use `middleware.ts` to protect routes. 
 
Main changes: 
- `/app/api` folder created
    - This is for routing only. It contains only an export from `/api` folder
- `/api` folder created to store API route handlers code. 
- `/api/auth/[auth0]` route handler moved from `/legacy/api` and updated. 
   - To make it work with app router I had to move some code from `/legacy`
   - `/legacy/backend` duplicated and updated to use codegen queries instead of simple Apollo client queries. Current folder: `/backend`
   - Some `/legacy/lib` duplicated in `/lib`
   - `getAPIServiceClient` (was needed to make ApolloClient manual server queries with service role) removed from `backend/graphql`. I left only `getAPIServiceGraphqlClient` to make codegen requests.
- Moved `/join` page under `(portal)` folder, because it also should be protected as we already have auth0 session there.
- Remove `/legacy/pages/api/auth/[auth0]`

---

https://github.com/worldcoin/developer-portal/assets/89008845/be933804-f1b0-4b88-87f0-f31fb09fcb69


web-app-router/tailwind.config.ts
```diff
@@ -1,11 +1,8 @@
 import type { Config } from "tailwindcss";
 
 const config: Config = {
-  content: [
-    "./pages/**/*.{js,ts,jsx,tsx,mdx}",
-    "./components/**/*.{js,ts,jsx,tsx,mdx}",
-    "./app/**/*.{js,ts,jsx,tsx,mdx}",
-  ],
+  content: ["./scenes/**/*.{js,ts,jsx,tsx,mdx}"],
+
```
andy-t-wang, line 5: What's the reason for this?

web-app-router/api/auth/[auth0]/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    return handleLogin(req, ctx, {
+      returnTo: invite_id
+        ? `/api/auth/login-callback?invite_id=${invite_id}`
+        : "/api/auth/login-callback",
+    });
+  },
+
+  callback: handleCallback,
+  "login-callback": loginCallback,
```
Comment (andy-t-wang) [inline web-app-router/api/auth/[auth0]/index.ts:26]: Should leave a TODO here that we need to add delete back

Comment (andy-t-wang) [reply web-app-router/api/auth/[auth0]/index.ts:26]: Or mark it in linear just so we come back

andy-t-wang (approved):
Looks fine to me. Left a couple of comments but if we're mostly migrating old files and it works seems fine to merge

web-app-router/tailwind.config.ts
igorosip0v: You are right, thanks!

web-app-router/api/auth/[auth0]/index.ts
Comment (igorosip0v) [reply web-app-router/api/auth/[auth0]/index.ts:26]: It will be created when I will start onboarding flow pages

Comment (igorosip0v) [reply web-app-router/api/auth/[auth0]/index.ts:26]: I've added a comment

web-app-router/middleware.ts
```diff
@@ -0,0 +1,3 @@
+import { withMiddlewareAuthRequired } from "@auth0/nextjs-auth0/edge";
+export default withMiddlewareAuthRequired();
+export const config = { matcher: ["/teams", "/teams/:path*", "/join"] };
```
pdtfh, line 3: can we instead use a whitelist approach where only certain routes are allowed without authentication?

web-app-router/api/auth/[auth0]/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  // TODO: Add delete account handler
+
+  logout: handleLogout({
+    returnTo: "/login",
```
Comment (pdtfh) [inline web-app-router/api/auth/[auth0]/index.ts:30]: are we sure we don't need login_error handling here now?

web-app-router/api/auth/[auth0]/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  }),
+
+  onError: (_req: NextRequest, error: HandlerError) => {
```
Comment (pdtfh) [inline web-app-router/api/auth/[auth0]/index.ts:33]: does next-auth already log errors? if not, let's log them ourselves

web-app-router/api/login-callback/index.ts
```diff
@@ -0,0 +1,289 @@
+import { getAPIServiceGraphqlClient } from "@/backend/graphql";
```
pdtfh, line 1: assuming no changes here, but let me know if I'm mistaken

web-app-router/api/auth/[auth0]/index.ts
Comment (igorosip0v) [reply web-app-router/api/auth/[auth0]/index.ts:30]: What login error should we have and how it should be displayed?

web-app-router/middleware.ts
igorosip0v: Just tested, it's possible to make it for the `/login` page for example, but we also will need to whitelist `/` and I have no idea how to do this correctly with middleware matcher

PR #476, opened 2024-01-24
feat: dialog component

michaelfurutin:
- Added building blocks for constructing a modal dialog, implemented typing, and provided styling capabilities.
- Added containers for illustrated icons.
- Added some icons.

image

image

Example:
```ts
import { Dialog, DialogProps } from '@/components/Dialog'
import { DialogOverlay } from '@/components/DialogOverlay'
import { DialogPanel } from '@/components/DialogPanel'
import { FeaturedIconContainer } from 'components/FeaturedIconContainer'
import { UserAddIcon } from '@/components/icons/UserAddIcon'

export const AddUserDialog = (props: DialogProps) => {
  return (
    <Dialog {...props}>
      <DialogOverlay />

      <DialogPanel>
        <FeaturedIconContainer>
          <UserAddIcon/>
        </FeaturedIconContainer>
        
        <h1>Add user</h1>
        
        <div>...</div>
      </DialogPanel>
    </Dialog>
  )
}
```


web-app-router/components/Dialog/index.tsx
igorosip0v, line 1: @michaelfurutin Can you give an example of using this Dialog component? 

Is it something like this? 

```jsx
<Dialog>
<DialogOverlay />
<DialogPanel>
<div>...</div>
</DialogPanel>
</Dialog>
```

1. If so what are use cases to use `DialogOverlay` and `DialogPanel` as a separate components?
2. Do we really need it as a separate component?
3. Shouldn't we keep `DialogOverlay` and `DialogPanel` under Dialog dir if we will only use it with `Dialog` component?

web-app-router/components/AlertIconContainer/index.tsx
```diff
@@ -0,0 +1,42 @@
+import { ReactNode } from 'react'
+
+type AlertIconContainerProps = {
+  children: ReactNode
+}
+
+export const AlertIconContainer = (props: AlertIconContainerProps) => {
```
igorosip0v, line 7: @michaelfurutin as I can see from your screenshot it's possible to use this container with different color, maybe it worth to name it something like `DialogStatusIconContainer`

web-app-router/components/Dialog/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { twMerge } from 'tailwind-merge'
+
+export type DialogProps = DialogPropsBase<'div'>
+
+export const Dialog = (props: DialogProps) => {
+  const { className, children, ...otherProps } = props
+
+  return (
+    <DialogBase
+      className={twMerge('fixed inset-0 flex items-center justify-center overflow-y-auto', typeof className === 'string' ? className : undefined)}
```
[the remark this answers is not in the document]
michaelfurutin: `clsx` does not solve the issue that arises when attempting to override styles. For example, if the component has a class `p-4`, and we want to override it with `p-6`, we would have to add !important. `tailwind-merge` will remove `p-4`, and only `p-6` will remain in the end.

It doesn't replace `clsx`. Instead, it complements it. We can use them together:
```
<div className={twMerge(clsx('p-4', { 'mt-2': !isFirst }), props.className)} />
```

web-app-router/package.json
```diff
@@ -37,6 +38,8 @@
     "react": "^18",
     "react-dom": "^18",
     "request-ip": "^3.3.0",
+    "tailwind-merge": "^2.2.1",
+    "tailwind-variants": "^0.1.20",
```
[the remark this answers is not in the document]
michaelfurutin: These are relatively new tools. I suggest giving them a try, and if we find them favorable, let's use them everywhere.

web-app-router/components/AlertIconContainer/index.tsx
michaelfurutin: If we look at [this screen](https://www.figma.com/file/URO9Y5UPOqK3LbLJ9623Gv/%F0%9F%AA%AA-World-ID-2.0?type=design&node-id=1668-34698&mode=dev), we'll see that this component can be used independently of the dialog.

I tried to create a common component and set its color through props, but I couldn't get it to work. There was some red shade left in the gradient that I couldn't overcome. We need to dive in and investigate. I can try again.

web-app-router/components/Dialog/index.tsx
michaelfurutin: Example added.

I believe that separate components are more convenient than a monolithic one.

When we create a full-screen dialog, we won't use DialogOverlay, and we'll only need to override a few classes for Dialog and DialogPanel.

In a monolithic dialog component, we would need some fullScreen prop and several conditional branches. This would make our component less versatile and more challenging to adapt. We've already encountered these difficulties.

web-app-router/package.json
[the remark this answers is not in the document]
igorosip0v: @michaelfurutin Let's discuss and document how to use them

web-app-router/components/AlertIconContainer/index.tsx
igorosip0v: I have an Idea, let me try

igorosip0v (approved):

web-app-router/components/Dialog/index.tsx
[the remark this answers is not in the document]
andy-t-wang: I thought class names over ride based on order? ie 
"p-6 p-4" means p-4 will take precedence so if classname is behind in order it will override

PR #478, opened 2024-01-24
fix: hide secret section for members

igorosip0v:
Closes [this](https://tfh.slack.com/archives/C05NJ4QFCLE/p1706041937278389?thread_ts=1706031364.968379&cid=C05NJ4QFCLE) and [this](https://tfh.slack.com/archives/C05NJ4QFCLE/p1706042415683769?thread_ts=1706031364.968379&cid=C05NJ4QFCLE)

We using Hasura action `_reset-secret-key` to reset secret key and there is only two options for actions: allow or forbid action for the entire user role, so I decided simply not show client secret section for the members

Also hide Redirects and Urls components from MEMBER on /sign-in page

image


pdtfh (approved):
already an improvement, but we should definitely protect the backend. this can be done in the Next.js route handler

PR #479, opened 2024-01-24
fix: hide delete action button for member

igorosip0v:


andy-t-wang:
Looks good. This is also blocked on the by permissions in the back right

andy-t-wang (approved):

PR #492, opened 2024-01-26
Fix: Update next config

andy-t-wang:
Config needs to use virtual style urls to match staging and prod. Locally my bucket wasn't DNS compatible so it was displaying a different behavior. We also update the cloudfront urls to remove verified path since that's added into the origin url already.


igorosip0v (approved):
@andy-t-wang Maybe it's a good idea to create some helpers that will accept app_id and return required URL. Just in case we will need to update it next time, to update it in one place.

igorosip0v (approved):

PR #496, opened 2024-01-27
wid 691 Feat: Create table component

andy-t-wang:
Screenshot 2024-01-27 at 11 44 00 AM
---

A general table component use can specify the values that go in. Note this requires client side since we do pagination see .../actions/page/ListActions for an example of how to use this 

Also included the implementation of the table via actions list


web-app-router/components/Table/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const handlePageChange = (newPage: number) => {
+    onPageChange(newPage);
+  };
+
+  // Calculate the number of pages
+  const pageCount = Math.ceil(totalResults / rowsPerPage);
+
+  return (
+    <div className="w-full h-full">
+      <table className="min-w-full divide-y divide-gray-100">
```
igorosip0v, line 48: @andy-t-wang I think we can keep it as it is for now, but maybe it's a good idea to refactor and use simple `div`s with `grid`. I think it will be easier to adjust table for responsive and mobile layout.

web-app-router/components/Table/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+              ))}
+            </tr>
+          ))}
+        </tbody>
+      </table>
+
+      <div className="sticky bottom-0 bg-white w-full grid grid-cols-3 text-xs items-center justify-between gap-x-4 py-4 border-t-[1px] border-gray-100">
+        <div className="text-grey-400">{totalResults} results</div>
+        <div className="flex items-center justify-center gap-x-4">
+          <button
```
igorosip0v, line 81: We can use common/Button here

web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/page/ListActions.tsx
```diff
@@ -0,0 +1,64 @@
...
+      </button>,
+    ],
+    [
+      <div key={2}>
+        <div className="text-grey-900 text-sm">New Action</div>
+        <div className="text-grey-500 text-xs">new-action</div>
+      </div>,
+      <span key={2}>25</span>,
+      <span key={2}>London</span>,
+      <button key={2} className="rotate-90">
```
igorosip0v, line 33: we can use common/Button everywhere

web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/page/ListActions.tsx
igorosip0v, line 1: @andy-t-wang Let's create components under the folder as default. `<ComponentName>/index.tsx` to keep it consistent. Because we will have components where we need to use gql queries and this component will be a folder with graphql folder and index.tsx inside

web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/page/ListActions.tsx
```diff
@@ -0,0 +1,64 @@
...
+    // Logic to handle page change
+  };
+
+  const handleRowsPerPageChange = (rowsPerPage: number) => {
+    // Logic to handle rows per page change
+  };
+
+  return (
+    <div className="flex items-center justify-center w-[700px]">
+      <TableComponent
```
igorosip0v, line 53: @andy-t-wang Maybe it's a good idea to implement this component in another way, like we did with Dialog and Select, for example. 
Something like: 

```jsx
<Table>
  <Header></Header>
  <Body></Body>
</Table>
```

web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/page/index.tsx
```diff
@@ -11,45 +12,54 @@ type ActionsPageProps = {
 // TODO: Ad TWK Lausanne font
 export const ActionsPage = ({ searchParams }: ActionsPageProps) => {
   const createAction = searchParams?.createAction;
+  const alpha = true;
   if (createAction) {
     return <CreateActionModal />;
```
igorosip0v, line 17: @andy-t-wang Maybe it's better to show/hide this modal with styles (opacity, visibility, pointer-events) instead of changing return of the whole component.

web-app-router/components/Table/index.tsx
igorosip0v, line 48: @andy-t-wang Also, we need to remember to use `grey` instead of `gray`. gray-100 is a different color from the tailwind default config. Please, fix it everywhere.

igorosip0v (changes requested):

web-app-router/components/Table/index.tsx
andy-t-wang: Uh I will try to refactor this in a seperate PR? I think I made changes to it in a later PR which will be a pain to rebase and figure out.

web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/page/index.tsx
andy-t-wang: should we do this with the create action modal as well?

andy-t-wang: just did it

igorosip0v (approved):

PR #508, opened 2024-01-29
fix: Update NEXT_PUBLIC_VERIFIED_IMAGES_CDN_URL

andy-t-wang:
Update NEXT_PUBLIC_VERIFIED_IMAGES_CDN_URL


web/src/hooks/useImage/index.ts
andy-t-wang, line 1: This needed to be updated for lint reasons the hook didn't have the right dependencies.

pdtfh (approved):

PR #514, opened 2024-01-31
Feat: Image verification custom action

andy-t-wang:
- Only reviewer has permission to access
- Updated permissions to allow reviewer to access the app and update images.
- Allow Reviewer to access rows that are also verified in order to delete 

### Functionally we are
- Checking Permissions
- Clearing out the verified images bucket
- Copying the new images over
- Deleting any existing verified rows
- Updating the awaiting_review row to verified with the correct information

Once again I apologize for my linter


igorosip0v (approved):

hasura/metadata/actions.yaml
```diff
@@ -106,6 +106,25 @@ actions:
...
+  - name: verify_app
+    definition:
+      kind: synchronous
+      handler: '{{NEXT_API_URL}}/images/_verify_app'
+      headers:
+        - name: Authorization
+          value_from_env: INTERNAL_ENDPOINTS_SECRET
+      request_transform:
+        method: POST
+        query_params:
```
pdtfh, line 118: will this transform to send the parameters in the body?

web/src/backend/jwts.ts
```diff
@@ -35,6 +35,28 @@ if (!GENERAL_SECRET_KEY) {
 
 // ANCHOR: -----------------HASURA JWTs--------------------------
 
+/**
+ * Generates a 1-min JWT for the `service` role (only for internal use from Next.js API)
+ * @returns
+ */
+export const generateReviewerJWT = async (): Promise<string> => {
```
pdtfh, line 42: given we're still not using this, best to remove

pdtfh: there's actually a misunderstanding here, the reviewer role calls the custom action, the custom action in Hasura calls the backend, the backend always interact with Hasura (for now at least) with the `service` role

web/src/api/images/verify_app.ts
```diff
@@ -0,0 +1,278 @@
...
+    if (
+      !["reviewer", "admin"].includes(
+        req.body.session_variables["x-hasura-role"]
+      )
+    ) {
+      return errorHasuraQuery({
+        res,
+        req,
+        detail: "Unauthorized Role",
+        code: "unauthorized_role",
```
pdtfh, line 54: Let's log the error but we should not send details back for security reason, let's remove the `detail` & `code` to send a generic error.

web/src/api/images/verify_app.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    const {
+      app_id,
+      reviewer_name,
+      is_reviewer_app_store_approved,
+      is_reviewer_world_app_approved,
+    } = req.body.input;
+
+    if (
+      !app_id ||
```
pdtfh, line 75: use yup to validate the input, worth checking existing conventions

hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -77,6 +77,14 @@ select_permissions:
       filter:
         team_id:
           _eq: X-Hasura-Team-Id
+  - role: reviewer
```
pdtfh, line 80: is this needed? considering the backend should connect to Hasura using the service role

web/src/api/images/verify_app.ts
```diff
@@ -0,0 +1,278 @@
...
+/**
+ * This function handles the verification of the app by a reviewer
+ * @param req Expects an app_id and reviewer_name is_reviewer_app_store_approved and is_reviewer_world_app_approved
+ * @param res Returns a success message if the app was verified
+ */
+export const handleVerifyApp = async (
+  req: NextApiRequest,
+  res: NextApiResponse
+) => {
+  try {
```
pdtfh, line 36: let's avoid the global try/catch, either we're returning a handled response or for unhandled errors will be logged and hasura will handle the 500

web/src/api/images/verify_app.ts
```diff
@@ -0,0 +1,278 @@
...
+        Prefix: destinationPrefix,
+      })
+    );
+
+    const verifiedImageKeysToDelete = listObjectsResponse.Contents?.map(
+      (object) => object.Key
+    );
+
+    if (verifiedImageKeysToDelete && verifiedImageKeysToDelete.length > 0) {
+      const expirePromises = verifiedImageKeysToDelete.map((key) =>
```
pdtfh, line 148: Let's add a TODO comment here, this should be done async from a worker

hasura/metadata/databases/default/tables/public_app_metadata.yaml
```diff
@@ -188,6 +188,9 @@ update_permissions:
   - role: reviewer
     permission:
       columns:
+        - logo_img_url
```
pdtfh, line 191: also likely not needed, `service` role performs these updates

web/src/api/images/verify_app.ts
```diff
@@ -0,0 +1,278 @@
...
+          })
+        );
+      }
+    );
+    copyPromises.push(...showcaseCopyPromises);
+    await Promise.all(copyPromises);
+
+    // Delete any existing verified rows
+    if (verifiedAppMetadata) {
+      const deleteVerifiedAppMetadata = await deleteVerifiedAppMetadataSDK(
```
pdtfh, line 219: why not do this in a single transaction with the update below?

web/src/api/images/verify_app.ts
pdtfh: this whole flow should actually be async

hasura/metadata/actions.yaml
andy-t-wang: yes this is generated from the console they just label it as query param but the fields themselves are listed in body.input

web/src/backend/jwts.ts
andy-t-wang: Sent you a message on slack but we are using a reviewer header in the client we use from verify app so we can enforce the reviewer permissions we set previously. I think this is better since we can add more granular permissions based on usecase

hasura/metadata/databases/default/tables/public_app.yaml
andy-t-wang: Moving here but I think it's better if we generate a client with reviewer permissions to perform the verify since we can be more granular with permissions. If we use the service client we will need to add more permissions to service. The backend is currently using the reviewer role since the `const reviewer_client = await getAPIReviewerGraphqlClient();` call is set with a reviewer role header variable

web/src/api/images/verify_app.ts
andy-t-wang: Combined the two but note delete_app_metadata_by_pk requires a String! type so we will pass empty string when there's no verifiedAppMetadata and nothing will happen since there are no apps with ID empty string

web/src/api/images/verify_app.ts
andy-t-wang: Added an issue on linear and todo

hasura/metadata/databases/default/tables/public_app_metadata.yaml
andy-t-wang: aligned

pdtfh (approved):
synced offline

PR #516, opened 2024-01-31
WID-703 Fix: users can reset arbitrary api keys

andy-t-wang:
Moved the _reset_key route to use code gen added a check to ensure the user is an owner on the team that has the api_key.
Created a new hasura relationship between team -> api_key that is an array relationship. This should exist anyways imo and makes the graphql cleaner.


web/src/api/_reset-api-key/graphql/check-user-permission.graphql
```diff
@@ -0,0 +1,11 @@
+query CheckUserPermission($id: String!, $team_id: String!, $user_id: String!) {
+  team(
+    where: {
+      id: { _eq: $team_id }
+      memberships: { user_id: { _eq: $user_id }, role: { _eq: OWNER } }
```
pdtfh, line 5: admin should also be allowed to reset API keys

web/src/api/_reset-api-key/graphql/check-user-permission.graphql
```diff
@@ -0,0 +1,11 @@
+query CheckUserPermission($id: String!, $team_id: String!, $user_id: String!) {
```
pdtfh, line 1: we seem to be recreating this query in multiple places, can we make it reusable? why not look into creating a centralized handler for all custom actions that are user-based that checks permissions and then routes the request to the relevant API function

pdtfh (changes requested):

web/src/api/_reset-api-key/graphql/check-user-permission.graphql
pdtfh: ignore, confusion on my end

pdtfh (approved):

PR #519, opened 2024-02-01
feat: multiple improvements to API permissions & logging

pdtfh:
- Updates `_invite-team-members` route to check ADMIN, OWNER permissions to the team.
- Removes global try/catch block to improve logging.
- Improves route performance to use a single query


web/src/api/_reset-client-secret/index.ts
```diff
@@ -92,16 +86,8 @@ export const handleSecretReset = async (
...
-    return errorHasuraQuery({
-      res,
-      req,
-      detail: "Admin is not allowed to run this query.",
-      code: "admin_not_allowed",
-    });
-  }
-
   // ANCHOR: Make sure the user can perform this client reset
+  // TODO: Merge query into get-membership
```
pdtfh, line 90: @andy-t-wang

andy-t-wang (approved):
Since you're the author I'll approve it and you can leave comments

web/tests/integration/app.test.ts
```diff
@@ -232,6 +233,127 @@ describe("user role", () => {
...
+          "x-hasura-team-id": tokenTeamId,
+        },
+      },
+    });
+
+    await handleSecretReset(req, res);
+
+    const responseData = res._getData();
+    const responseJSON = JSON.parse(responseData);
+    expect(responseJSON.client_secret).toBeDefined();
```
pdtfh, line 355: @andy-t-wang generally a good practice to also check the DB to assert objects are in the state you expect

PR #521, opened 2024-02-02
WID-701 Sign in with World ID

andy-t-wang:
Screenshot 2024-02-01 at 6 56 31 PM

### Changes
- Added sign in with world id
- Made a separate component to render the redirect urls since it's a bit finnicky with forms 
- Adjusted the sizing wrapper to 8vw tbh I think it looks aesthetically better that way the 11 looked way too narrow. If that's off from the designs we can ask liza. 
- TODO will test reset once I have the hasura session variables set. Currently can't call custom actions
- Refactor Input to have props for left and right


web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/SignInWithWorldId/page/ClientInformation/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+  privacy_policy_uri: yup.string().url("Must be a valid URL").optional(),
+  terms_uri: yup.string().url("Must be a valid URL").optional(),
+});
+
+type ClientInformation = yup.InferType<typeof schema>;
+
+export const ClientInformationPage = (props: {
+  appID: string;
+  teamID: string;
+  action: {
```
igorosip0v, line 29:
```suggestion
  action: SignInActionQuery["action"][number]
```

igorosip0v: @andy-t-wang Use type from the generated file, this is one of the advantages of codegen for queries.

igorosip0v: @andy-t-wang Also let's think how we want to handle the case when there is no action. It's actually shouldn't be the case here, because sign in action should be created when you create an app. But I think we should handle this case from the UI point of view.
Two options there, as I can see: 
1. Mark action as optional here and adjust code right on this page to hide some elements to display page properly.
2. Simply catch an error it with error.tsx file. [Docs](https://nextjs.org/docs/app/building-your-application/routing/error-handling)

web-app-router/legacy/api/auth/login-callback.ts
```diff
@@ -35,7 +35,7 @@ import { getAPIServiceGraphqlClient } from "@/legacy/backend/graphql";
 import { urls } from "@/legacy/lib/urls";
 import { Auth0User, LoginErrorCode } from "@/lib/types";
 import { getSdk as updateUserSdk } from "./graphql/update-user.generated";
-import { isEmailUser } from "@/legacy/lib/utils";
+import { isEmailUser } from "@/lib/utils";
```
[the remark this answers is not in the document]
igorosip0v: Related to all imports in /legacy folder

igorosip0v (approved):

web-app-router/legacy/api/auth/login-callback.ts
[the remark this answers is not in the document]
andy-t-wang: Oh I see I though we were using isEmailUser and should move everything from legacy

web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/SignInWithWorldId/page/ClientInformation/index.tsx
andy-t-wang: Yeah I just pulled that server side action we don't use it anyways

PR #523, opened 2024-02-02
Danger Zone [Profile]

igorosip0v:
https://github.com/worldcoin/developer-portal/assets/89008845/0eedffda-d15b-4d3d-88fb-5acbbc08336f


web-app-router/tailwind.config.ts
```diff
@@ -134,7 +135,7 @@ const config: Config = {
         32: "calc(32 * 1rem / 16)",
       },
 
-      width: {
+      spacing: {
```
andy-t-wang, line 138: Why are we changing this?

web-app-router/scenes/Portal/Profile/DangerZone/page/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { DeleteAccountDialog } from "../DeleteAccountDialog";
+
+export const DangerZone = () => {
+  const [open, setOpen] = useState<boolean>(false);
+
+  return (
+    <SizingWrapper className="py-8">
+      <div className="grid gap-y-8">
+        <UserInfo color="pink" name="Lisa" email="lisa@toolsforhumanity.org" />
+
```
andy-t-wang, line 20: Why is this hard coded?

web-app-router/tailwind.config.ts
igorosip0v: This is allows to use these values with other classes, for example, I was needed it with max-w today. It also applies to paddings, margins, etc. Just not to duplicate

web-app-router/scenes/Portal/Profile/DangerZone/page/index.tsx
igorosip0v: If we need to apply logic already, then we need to merge #524 first

andy-t-wang (approved):

PR #527, opened 2024-02-04
WID-711 App Profile [Feature Branch]

andy-t-wang:
### Changes
- Updated endpoints for image upload to GET/POST
- Added next.config
- Use Typography for Action Nav
- Create a general fetch graphql query to keep app profile page up to date. 
- Created a stateless hook to house the image upload logic since we will reuse it multiple places and there's more logic needed beyond the graphql query

Screenshot 2024-02-05 at 5 22 18 PM


web-app-router/next.config.js
andy-t-wang, line 1: Please review this @igorosip0v I added "'unsafe-inline'", since it was breaking things and we still have a refetch-src not used error that shows up in console which isn't impacting anything

web-app-router/api/images/get_unverified_images.ts
```diff
@@ -22,10 +22,7 @@ export type ImageGetAllUnverifiedImagesResponse = {
  * @param req
  * @param res
  */
-export const handleGetAllUnverifiedImages = async (
-  req: NextApiRequest,
-  res: NextApiResponse,
-) => {
+export const GET = async (req: NextApiRequest, res: NextApiResponse) => {
```
igorosip0v, line 25: @andy-t-wang As I can see you import image handlers in legacy pages for now. I think we can keep these image handlers as it is in legacy folder for now. It should work, or am I missing something? 

Because app router's route handlers work a bit different and require a refectoring.

web-app-router/next.config.js
igorosip0v: @andy-t-wang What exactly breaks if you not include unsafe-inline in script-src?

to remove refetch src I think you can use [this](https://trezy.gitbook.io/next-safe/configuration-options/contentsecuritypolicy#disable-the-prefetch-src-directive)

web-app-router/middleware.ts
andy-t-wang, line 1: is dev here is matching the commands from next-safe isDev

PR #531, opened 2024-02-04
Create app form [Create App]

igorosip0v:
Closes [WID-719](https://linear.app/worldcoin/issue/WID-719/create-app-[form])

What is missing: 
- Images: 
  - Format is not ready to upload images
  - Form selector is not support images

---

https://github.com/worldcoin/developer-portal/assets/89008845/668b4846-29f8-482c-bd54-acb011de6ddd


web-app-router/components/Radio/index.tsx
```diff
@@ -45,19 +45,22 @@ export const Radio: React.FC<RadioProps> = memo(function Radio(
...
-            "hover:text-grey-700 hover:peer-checked:text-grey-500":
-              !disabled && !errors,
-          },
-        )}
-      >
-        {label}
-      </span>
-    </label>
+
+      {props.label && (
```
andy-t-wang, line 49: label is extracted as a variable above

andy-t-wang, line 49:
```suggestion
      {label && (
```

web-app-router/scenes/Portal/Teams/TeamId/Apps/page/index.tsx
andy-t-wang, line 1: Because this is server side if the user updates the app name in the app profile section this won't get updated. Worth syncing on how we'll handlet his

igorosip0v: This is an apps page that will only redirect the user to /apps/[id] or will display the client component with a tutorial that does not contain any app name.

Can you clarify what you mean?

igorosip0v: App profile (/apps/[id]/) should be a client component (that can be client component inside server page). To get app name inside that component you will fetch app data with apollo hook, and when you will submit the form to update app name you can make `refetchQueries` to update query that you've used to display app name.

PR #534, opened 2024-02-06
WID-713 App profile danger zone

andy-t-wang:
Add danger zone. Missing actual delete logic waiting on routes

Screenshot 2024-02-06 at 3 15 10 PM


web-app-router/components/Input/index.tsx
```diff
@@ -87,7 +89,8 @@ export const Input = memo(function Input(props: InputInterface) {
         <div className="flex items-center">{addOnRight && addOnRight}</div>
 
         <legend className={labelClassNames}>
-          {label} {required && <span className="text-system-error-500">*</span>}
+          {labelNode ? labelNode : label}{" "}
```
igorosip0v, line 92: @andy-t-wang Maybe it's a good idea to update the label type to `ReactNode`? I think string matches the `ReactNode` type

igorosip0v (approved):

PR #536, opened 2024-02-07
fix https://id.worldcoin.org/v1 claim in /userinfo

pdtfh:


andy-t-wang:
Integration tests need to be updated but otherwise looks good

andy-t-wang (approved):

PR #546, opened 2024-02-09
WID 745: Hasura Changes for Create Action Stats

andy-t-wang:
Added hasura table for action_stats_returning

I need to test the actual table with the chart still, but the migrations run fine with no troubles

We need to rebase still to pull in the verify_app query and other changes that got merged to main

### DIFF
--- a/hasura/migrations/default/1707463125126_create_action_stats_table/up.sql
+++ b/hasura/migrations/default/1707463125126_create_action_stats_table/up.sql
@@ -1,0 +1,33 @@
+CREATE TABLE "public"."action_stats_returning" (
+    "action_id" text NOT NULL, 
+    "date" timestamp with time zone NOT NULL, 
+    "verifications" numeric NOT NULL, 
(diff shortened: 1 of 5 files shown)


hasura/migrations/default/1707463125126_create_action_stats_table/up.sql
```diff
@@ -0,0 +1,33 @@
...
+);
+
+COMMENT ON TABLE "public"."action_stats_returning" IS E'Returning value of action_stats function';
+
+CREATE OR REPLACE FUNCTION public.action_stats(
+    "timespan" character varying,
+    "startsAt" timestamp with time zone, 
+    "actionId" character varying
+)
+-- If you changing returning values of this function, please, change app_stats_returning columns too
```
igorosip0v, line 19:
```suggestion
-- If you changing returning values of this function, please, change public_action_stats_returning columns too
```

igorosip0v (approved):

PR #547, opened 2024-02-09
feat: update user constraint

michaelfurutin:
When we delete a team, the relationship is also removed from the membership table, along with any associated applications and actions. This is achieved through foreign constraints.
However, currently, the user is also deleted along with the team. To prevent this from happening, I've adjusted the team_id field, making it nullable, and also updated the constraint.
We're not using the team_id field in the user table; it's kept for backward compatibility with the old frontend. We'll need to remove it entirely.


andy-t-wang (changes requested):
I believe we do use the teamid as a foreign key through the users table see here. 
https://github.com/worldcoin/developer-portal/blob/main/web/src/api/_invite-team-members/graphql/fetchUser.graphql

andy-t-wang (approved):
Noting as a response to the PR comment above. We should not be deleting team_id from the user table in the future, or if so are ensuring we check that it's been fixed in the queries we use it like in invite-team-members

PR #548, opened 2024-02-09
WID-745 Actions page table frontend

andy-t-wang:
- Update chart component to only show whole numbers (doesn't make sense to have decimals)
- Updated the Chart to show a point if there's no data 
- Added chart with aspect ratio to the graph
- Unfortunately the way we calculate verification is off but it's better imo to show # of verifications but the chart is messed up then not show the correct number of verification. We can revisit in the future how to fix this with better analytics 
- Adds Gradient to chart

Completes WID-731
Screenshot 2024-02-09 at 4 31 08 PM

Screenshot 2024-02-09 at 4 31 45 PM

### DIFF
--- a/hasura/metadata/databases/default/tables/public_nullifier.yaml
+++ b/hasura/metadata/databases/default/tables/public_nullifier.yaml
@@ -39,11 +39,12 @@ select_permissions:
   - role: user
     permission:
       columns:
+        - id
         - action_id
         - created_at
-        - id
-        - uses
+        - updated_at
         - nullifier_hash
+        - uses
       filter:
         action:
           app:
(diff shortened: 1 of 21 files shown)


hasura/metadata/databases/default/tables/public_nullifier.yaml
```diff
@@ -39,11 +39,12 @@ select_permissions:
   - role: user
     permission:
       columns:
+        - id
         - action_id
         - created_at
-        - id
-        - uses
+        - updated_at
```
andy-t-wang, line 45: We sort the nullifiers by updated at so need to add it

igorosip0v (approved):

PR #552, opened 2024-02-09
WID-749 Profile Update

andy-t-wang:
- Add loading skeletons and correct width calculations (was using 2 cols before which would break)
- Added logic so user can submit for review without showcase and hero image if app store listing is un-checked
- Moved image api routes back to legacy
- Removed the image drop if the images are full
- Updated URL error messages to match the labels on the inputs
- 
Screenshot 2024-02-09 at 3 10 21 PM


web-app-router/scenes/Portal/Teams/TeamId/Apps/AppId/Profile/PageComponents/AppTopBar/Environment/index.tsx
```diff
@@ -20,7 +20,7 @@ export const Environment = (props: EnvironmentProps) => {
             <StartUpIcon className="w-4 h-auto text-system-success-300 " />
           )}
           {environment === "staging" && (
-            <SmartPhoneIcon className="w-4 h-auto text-system-warning-300" />
+            <SmartPhoneIcon className="w-4 h-auto text-system-warning-500" />
           )}
```
andy-t-wang, line 24: I am making a personal decision that this looks better since the other yellow is too light

igorosip0v (approved):

PR #553, opened 2024-02-10
Fix: Update layout for large screens

andy-t-wang:
completes WID-729
Created a new navbar wrapper component for navs
I also changed the order for app profile to go last since that's what should make sense flow wise
Create an app -> create an action -> get verified

Screenshot 2024-02-09 at 5 17 14 PM


web-app-router/components/NavWrapper/index.tsx
igorosip0v, line 1: @andy-t-wang I'd prefer not to create additional components. We can leave it in a sizing wrapper with an optional `variant` or `size` prop.

igorosip0v (approved):

web-app-router/components/NavWrapper/index.tsx
andy-t-wang: sounds good

PR #556, opened 2024-02-11
[Fix] Bug fixes and style updates

andy-t-wang:
Added a default color based on the name and a hash function I made up 
I made the create screen steps wider and updated the nav according to what makes the most logical sense as you navigate through

Screenshot 2024-02-10 at 6 16 10 PM


web-app-router/scenes/Portal/layout/AppSelector/index.tsx
andy-t-wang, line 1: I added a color to the app if there's nothing. Use HSL so I have better control of the lightness of the random color generated. The hash function is just one I made up to pick a color and do randomness

web-app-router/scenes/Portal/layout/CreateAppDialog/index.tsx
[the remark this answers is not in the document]
andy-t-wang: Ideally this would be the workflow specifically for the first app but unclear if we can do that without needing to fetch apps

web-app-router/scenes/Portal/layout/AppSelector/index.tsx
igorosip0v: As an alternative, I can suggest getting an array of colors that match (bg + text colors) our design and picking color randomly by array index using char code

igorosip0v (approved):

PR #557, opened 2024-02-11
WID-751 Add Posthog and some fixes

andy-t-wang:


web-app-router/components/LoggedUserNav/HelpNav/index.tsx
```diff
@@ -37,39 +54,48 @@ export const HelpNav = () => {
...
-          <Button
-            href={WORLDCOIN_PRIVACY_URL}
-            className="grid grid-cols-auto/1fr items-center gap-x-2"
-          >
-            <SecurityIcon />
-            <Typography variant={TYPOGRAPHY.R4} className="text-grey-900">
-              Data Privacy & Security
-            </Typography>
-          </Button>
+          <div>
```
andy-t-wang, line 57: had to wrap in div otherwise was throwing a forwardref error

igorosip0v: `<DropdownItem as="div" ...`

web-app-router/pages/api/v1/verify/[app_id].ts
```diff
@@ -23,7 +23,7 @@ import {
 } from "@worldcoin/idkit-core";
 
 import * as yup from "yup";
-import { captureEvent } from "@/legacy/services/posthogClient";
+import { captureEvent } from "@/services/posthogClient";
```
Comment (igorosip0v) [inline web-app-router/pages/api/v1/verify/[app_id].ts:26]: @andy-t-wang As I can see you've moved this helper to actual services folder, but still using only in legacy pages folder. If you need this helper in relevant code just duplicate it  to services and keep `@/legacy/services` for the `/pages`

igorosip0v (approved):

web-app-router/components/LoggedUserNav/HelpNav/index.tsx
andy-t-wang: this then gives me an error since we are wrapping buttons that buttons cannot be divs

web-app-router/pages/api/v1/verify/[app_id].ts
Comment (andy-t-wang) [reply web-app-router/pages/api/v1/verify/[app_id].ts:26]: Ok duplicated it to the legacy folder

PR #584, opened 2024-02-14
[Fix] Make mini kiosk more stable

andy-t-wang:
Before it was changing screen state in idkit as well which is causing weird flashes. We rely on proof verification endpoint to determine if it should be in error or success state.


web/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/ActionId/Settings/TryAction/MiniKiosk/IDKitBridge/index.tsx
```diff
@@ -77,6 +77,7 @@ export const IDKitBridge = memo(function IDKitBridge(props: IDKitBridgeProps) {
 
   // Change the shown screen based on current verificationState and errorCode
   useEffect(() => {
+    console.log(idKitVerificationState);
```
igorosip0v, line 80:
```suggestion
```

igorosip0v (approved):

PR #585, opened 2024-02-14
Fix: Clean up placeholder and fixes on steps

andy-t-wang:
- Removed skip tutorial
- Removed test it step (synced with lisa)
- Added a prefetch, it's slightly faster in some cases from what I've seen
- I ran into the dialog issue again but realize we need to make the base dialog component z-50. Tested it when I ran into the issue seemed to fix it
Screenshot 2024-02-14 at 1 04 25 PM

### DIFF
--- a/web/components/Dialog/index.tsx
+++ b/web/components/Dialog/index.tsx
@@ -12,7 +12,7 @@ export const Dialog = (props: DialogProps) => {
   return (
     <DialogBase
       className={twMerge(
-        "fixed inset-0 flex items-center justify-center overflow-y-auto",
+        "fixed inset-0 z-50 flex items-center justify-center overflow-y-auto",
         typeof className === "string" ? className : undefined,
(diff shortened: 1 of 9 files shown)


web/components/Dialog/index.tsx
```diff
@@ -12,7 +12,7 @@ export const Dialog = (props: DialogProps) => {
   return (
     <DialogBase
       className={twMerge(
-        "fixed inset-0 flex items-center justify-center overflow-y-auto",
+        "fixed inset-0 z-50 flex items-center justify-center overflow-y-auto",
```
igorosip0v, line 15: @andy-t-wang I'd say let's add this via className prop from outside.

igorosip0v (approved):

andy-t-wang:
Noting here: checked the permissions list and change member modals, both look fine

PR #593, opened 2024-02-16
[Fix] More QA formatting and QOL fixes

andy-t-wang:
- Updated the fonts, bit hacky because we need to pull from staging on staging, and prod on prod but we can't write code in css files. I just import both and it takes the one it needs. 
- Updated top links for actions page to be more specific
- Link simulator in debugger and fix width so it doesn't expand on long error message, also updated copy
- Updated copy for on chain apps
- Added padding for mini kiosk so it doesn't look dumb on small screens
- Standardized the copy button across the app
- Lisa and I synced on using small font for desktop since it looks cleaner but using 16 px on mobile to prevent auto zoom, created a custom class R0 to reflect that
- Standardized category select component 
- Generally cleaned up formatting

Screenshot 2024-02-15 at 6 19 39 PM
Screenshot 2024-02-15 at 5 44 56 PM
Screenshot 2024-02-15 at 5 05 28 PM
Screenshot 2024-02-15 at 4 59 48 PM


web/styles/globals.css
```diff
@@ -50,6 +50,15 @@
   font-display: swap;
 }
 
+@font-face {
+  font-family: "TWK Lausanne";
+  src: url("https://staging.world-id-assets.com/fonts/TWKLausanne-550.woff2")
```
igorosip0v, line 55: @andy-t-wang I assume there should be 400 font

igorosip0v: Also, is it possible to make it something like this?

```css
src: url('https://world-id-assets.com/fonts/TWKLausanne-400.woff2') format('woff2'),
url('https://staging.world-id-assets.com/fonts/TWKLausanne-400.woff2') format('woff2');

```

igorosip0v (approved):

web/styles/globals.css
andy-t-wang: hmm this doesn't work but yes added size 400 font

PR #599, opened 2024-02-18
Sync with main3

igorosip0v:


web/legacy/api/images/verify_app.ts
```diff
@@ -126,21 +126,6 @@ export const handleVerifyApp = async (
     (metadata) => metadata.verification_status === "verified",
   );
 
-  // Check if app is allowed to be app store and world app approved
```
igorosip0v, line 129: @andy-t-wang I think we should keep it, or am I missing something?

andy-t-wang: Yeah lets get rid of this of the changes in this file, I made these changes in dev-portal-2 not in main

web/tests/integration/team.test.ts
andy-t-wang, line 1: lets remove this too. Looks like its just undoing the lint

web/tests/integration/app.test.ts
andy-t-wang, line 1: export const IDKitBridge = memo(function IDKitBridge(props: IDKitBridgeProps) {

web/legacy/backend/graphql.ts
andy-t-wang, line 1: export const IDKitBridge = memo(function IDKitBridge(props: IDKitBridgeProps) {

web/legacy/api/images/verify_app.ts
igorosip0v: I've revert this

web/tests/integration/team.test.ts
igorosip0v: No, this is applies current linting correctly.

andy-t-wang (approved):

PR #603, opened 2024-02-18
[Feat] Create an endpoint for reviewer to view images

andy-t-wang:
Let's reviewers see images awaiting review. Separate endpoint so we can have more granular permissions and since conditions are different. This seems cleaner

### DIFF
--- a/web/legacy/api/images/get_app_review_images.ts
+++ b/web/legacy/api/images/get_app_review_images.ts
@@ -1,32 +1,40 @@
-import { getSdk as getUnverifiedImagesSDK } from "@/legacy/api/images/graphql/getUnverifiedImages.generated";
+import { getSdk as getAppReviewImages } from "@/legacy/api/images/graphql/getAppReviewImages.generated";
 import {
   errorHasuraQuery,
   errorNotAllowed,
   errorResponse,
 } from "@/legacy/backend/errors";
-import { getAPIServiceGraphqlClient } from "@/legacy/backend/graphql";
+import { getAPIReviewerGraphqlClient } from "@/legacy/backend/graphql";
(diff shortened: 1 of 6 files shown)


web/legacy/api/images/get_app_review_images.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    if (app.showcase_img_urls && Array.isArray(app.showcase_img_urls)) {
+      const showcaseUrlPromises = app.showcase_img_urls.map((key) =>
+        getSignedUrl(
+          s3Client,
+          new GetObjectCommand({
+            Bucket: bucketName,
+            Key: objectKey + key,
+          }),
+          { expiresIn: 7200 },
```
igorosip0v, line 136: @andy-t-wang May be it's a good idea to put this expires value under single const

web/legacy/api/images/get_app_review_images.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    if (!protectInternalEndpoint(req, res)) {
+      return;
+    }
+
+    if (req.method !== "GET") {
+      return errorNotAllowed(req.method, res, req);
+    }
+
+    const body = JSON.parse(req.body);
```
igorosip0v, line 46: @andy-t-wang May be it's a good idea to use yup schema and `validateRequestSchema` from `'web/legacy/backend/utils'` to validate schema.

igorosip0v: But not sure if we really need it for actions

igorosip0v (approved):

web/legacy/api/images/get_app_review_images.ts
andy-t-wang: Added yup

PR #606, opened 2024-02-19
fix: initial steps header overlap

igorosip0v:
- Initial steps were moved -40px to top and it was blocking navigation on header

---

What was fixed: 

image

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/page/ClientPage/index.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/page/ClientPage/index.tsx
@@ -15,7 +15,7 @@ export const ClientPage = () => {
 
   return (
     <Fragment>
-      <SizingWrapper gridClassName="h-full flex justify-center items-center">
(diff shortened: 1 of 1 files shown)


web/scenes/Portal/Teams/TeamId/Apps/page/ClientPage/index.tsx
```diff
@@ -15,7 +15,7 @@ export const ClientPage = () => {
 
   return (
     <Fragment>
-      <SizingWrapper gridClassName="h-full flex justify-center items-center -mt-10">
+      <SizingWrapper gridClassName="h-full flex justify-center items-center">
```
andy-t-wang, line 18:
```suggestion
      <SizingWrapper gridClassName="h-full flex justify-center items-center pb-10">
```

I added this originally since the content sat a bit low on the page in my opinion. This should fix boundary issues?

andy-t-wang (approved):
Made a small change

PR #607, opened 2024-02-19
fix: add name for the nullifier users

igorosip0v:
- This PR fixes empty name for the user logged in with `sign in with world id`. 
- Also fixes lightOrange color

‼️ I had to add permissions to user role to select world_id_nullifier, so it's a Hasura migration change

### REVIEW:
@andy-t-wang I don't really like current format of nullifier name. I'd prefer if it starts with truncated nullifier value to make it look more unique. Do you have any ideas?  

---

image

### DIFF
--- a/web/lib/utils.ts
+++ b/web/lib/utils.ts
@@ -94,5 +94,5 @@ export const checkUserPermissions = (
 
 export const getNullifierName = (nullifier: string | undefined | null) => {
   if (!nullifier) return null;
-  return `Anonymous User | ...${nullifier.slice(-5)}`;
+  return `${nullifier.slice(0, 6)}...${nullifier.slice(-4)}`;
 };
--- a/web/lib/utils.ts
+++ b/web/lib/utils.ts
@@ -91,3 +91,8 @@ export const checkUserPermissions = (
   }
   return validRoles.includes(membership?.role);
 };
+
+export const getNullifierName = (nullifier: string | undefined | null) => {
+  if (!nullifier) return null;
+  return `Anonymous User | ...${nullifier.slice(-5)}`;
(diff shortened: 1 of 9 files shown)


web/scenes/Portal/Profile/common/UserInfo/index.tsx
```diff
@@ -10,19 +11,18 @@ import { twMerge } from "tailwind-merge";
...
-  const userHasura = user?.hasura as { id: string } | undefined;
+  const { user } = useUser() as Auth0SessionUser;
+  const userHasura = user?.hasura;
   const userId = userHasura?.id;
 
   const userQueryRes = useFetchUserQuery({
     variables: !userId ? undefined : { user_id: userId },
     context: {
-      headers: { team_id: "team_b0b7af3f49ea4b6106332f7b6dd5b708___" },
+      headers: { team_id: "_" },
```
andy-t-wang, line 25: what is this?

web/lib/utils.ts
```diff
@@ -91,3 +91,8 @@ export const checkUserPermissions = (
   }
   return validRoles.includes(membership?.role);
 };
+
+export const getNullifierName = (nullifier: string | undefined | null) => {
+  if (!nullifier) return null;
+  return `Anonymous User | ...${nullifier.slice(-5)}`;
```
andy-t-wang, line 97: why do we need to say anonymous user. Can we just do a 0x{nullifier} or something

andy-t-wang (approved):

web/lib/utils.ts
igorosip0v: Updated

PR #608, opened 2024-02-19
Fixes

igorosip0v:
- Fixes names for nullifier users in other places in the app
- Removes EditTeam modal
  - Makes a link to the /teams/id/settings instead


web/scenes/Portal/Profile/common/UserInfo/index.tsx
```diff
@@ -19,41 +21,48 @@ export const UserInfo = (props: UserInfoProps) => {
...
+          {!data?.user ? <Skeleton width={200} /> : name}
         </Typography>
 
         <Typography variant={TYPOGRAPHY.R4} className="text-grey-500">
-          {!userQueryRes.data?.user ? (
-            <Skeleton width={200} />
-          ) : (
-            userQueryRes.data.user.email || null
-          )}
+          {!data?.user ? <Skeleton width={200} /> : data.user.email || null}
```
andy-t-wang, line 65:
```suggestion
          {!data?.user ? <Skeleton width={200} /> : data.user?.email || null}
```
Will break if email is null?

andy-t-wang (approved):

PR #615, opened 2024-02-20
[Fix] Rever back to config

andy-t-wang:
Try to remove the use server from posthog, we didn't need it before and it was breaking things in main
Add the NEXT_PUBLIC_APP_URL to build time vars

I will merge this to test but you should review anyways @igorosip0v

Note I made one fix from #614 so look at this one


andy-t-wang:
Fix the docker file was wrong env

PR #620, opened 2024-02-20
Me query

igorosip0v:
This PR: 
- Adds Me query to fetch user and user related memberships
- Adds hook to make a me query fetch
- Updates session on user data update with `updateSessionUser` server action
- Replaces client FetchMembersips queries that were related to the user.
- Adds refetch Me query on actions that should update session user
  - Delete team
  - Transfer ownership
  - Update team name
  - Update user name
  

---
Me query in the network tab is a session value from the auth0 

https://github.com/worldcoin/developer-portal/assets/89008845/9491cd68-5ed1-4f73-8f49-38f8c9f3a4ed


andy-t-wang:
Looks fine to me. Though the route guards on middle ware will still be using the session variables

andy-t-wang (approved):

PR #627, opened 2024-02-21
[Fix] Shouldn't be able to access a user from another team

andy-t-wang:
Added a test case
https://linear.app/worldcoin/issue/WID-788/idor-vulnerability-allows-seeing-cross-team-user-information

Problem was that users could spoof team-id as it's just a header so we need to check user_id as well

- Fixed one other permission that was just not in line with the existing permissions

### DIFF
--- a/web/tests/integration/user.test.ts
+++ b/web/tests/integration/user.test.ts
@@ -202,4 +202,63 @@ describe("user role", () => {
 
     expect(response.data.user).toEqual([]);
   });
+
+  test("can't select user from a team you are not a part of, with a spoofed team_id header", async () => {
+    const { rows: teams } = (await integrationDBExecuteQuery(
+      `SELECT id, name FROM "public"."team"`,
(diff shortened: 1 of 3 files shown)


web/tests/integration/user.test.ts
```diff
@@ -85,4 +85,121 @@ describe("user role", () => {
...
+
+    const userFromTeam0 = teamMemberships.find(
+      (membership) => membership.team_id === teams[0].id,
+    );
+
+    const userFromTeam1 = teamMemberships.find(
+      (membership) => membership.team_id === teams[1].id,
+    );
+
+    const client = await getAPIUserClient({
```
pdtfh, line 191: should also try to spoof the `X-Hasura-Team-Id` header directly

pdtfh (approved):

web/tests/integration/user.test.ts
andy-t-wang: added a test with a header

PR #632, opened 2024-02-22
fix: delete team

igorosip0v:
- Removes unused duplicated DeleteTeamDialog
- Replaces update session action with an api endpoint
- Adds more proper loading state for a me query hook. 
- Updates team deelete submit logic to await session update before redirect.


web/scenes/Portal/common/DeleteTeamDialog/index.tsx
```diff
@@ -81,29 +84,36 @@ export const DeleteTeamDialog = (props: DeleteTeamDialogProps) => {
...
-
-      if (path !== urls.profileTeams()) {
-        return router.push(urls.profileTeams());
-      }
-
-      onClose();
+      setDeleteFinished(true);
     } catch (e) {
       console.error(e);
       toast.error("Error team deleting");
```
andy-t-wang, line 90:
```suggestion
      toast.error("Error deleting team");
```

web/scenes/Portal/common/DeleteTeamDialog/index.tsx
```diff
@@ -56,15 +57,17 @@ export const DeleteTeamDialog = (props: DeleteTeamDialogProps) => {
 
   const onClose = useCallback(() => {
     reset();
+    setDeleteFinished(false);
     props.onClose(false);
-  }, [props, reset]);
+  }, [props, reset, setDeleteFinished]);
 
   const [deleteTeam] = useDeleteTeamMutation({
```
andy-t-wang, line 64: can't we just use the loading in here instead of the setDeleteFinished

andy-t-wang:
I dont understand why we make this an endpoint? We are not going to be updating session outside of the app. Server actions seems better then?

web/scenes/Portal/common/DeleteTeamDialog/index.tsx
igorosip0v: @andy-t-wang I handled it like this because we updating the cache in `useEffect` on the `data` object change. I had to do this because apollo's `onCompleted` function executes only when we fetch query first time and it's not triggered when some mutation refetching `meQuery`. 

Also `meQuery` is just a react hook that uses apollo query inside, and it returns a bit modified `user` after the `useEffect`. So we need to wait not only apollo query refetch, but also the modified that is returned from the hook. 
 Otherwise we receiving not updated user in the submit and memberships array is still having already deleted item. 
 
 So, with this refactor we are submitting team delete, waiting while meQuery hook return actual data and then making correct redirect.

igorosip0v:
@andy-t-wang 
> I dont understand why we make this an endpoint? We are not going to be updating session outside of the app. Server actions seems better then?

We can update the session only from server action or from the route handler. I didn't succeed in figuring out why we had `Error: Failed to find Server Action` error. But with the endpoint that does the same we don't have such an error.

andy-t-wang:
Hmm if this isn't related to the server action error then there' no need to make it an endpoint. I feel like it would be better to leave as a server action then so it's not exposed?

andy-t-wang (approved):

andy-t-wang:
Synced on this, server actions are experimental still so we are not using them as they throw weird errors

PR #650, opened 2024-02-27
[Fix] add sharp

andy-t-wang:
https://nextjs.org/docs/messages/sharp-missing-in-production

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -46,6 +46,7 @@
     "react-toastify": "^10.0.4",
     "request-ip": "^3.3.0",
     "server-only": "^0.0.1",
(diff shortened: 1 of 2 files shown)


igorosip0v (approved):

pdtfh (commented):
I believe Next.js requires sharp to be available in production, the standalone version will not include dependencies, this needs to be likely added in the Docker image or adjustments to next.config.js

https://nextjs.org/docs/messages/sharp-missing-in-production

andy-t-wang:
I built the image locally and it was fine. Based on what I'm seeing online you need to add it if you are doing standalone which we aren't currently.

PR #651, opened 2024-02-27
Remove team_id header

igorosip0v:
This PR: 
- Removes checks of `X-Hasura-Team-Id` on Hasura side
- Remove using `team_id` header from web app
- Refactors some queries and client code to get rid of `team_id` header
- Updates schema


hasura/metadata/databases/default/tables/public_user.yaml
```diff
@@ -56,11 +56,10 @@ select_permissions:
           - id:
               _eq: X-Hasura-User-Id
           - memberships:
-              _and:
-                - user_id:
+              team:
```
andy-t-wang, line 59: can we not go straight from team here? It is a foreign key in the table

web/api/helpers/jwts.ts
```diff
@@ -98,10 +97,9 @@ export const getUserJWTPayload = (user_id: string, team_id: string) => ({
  */
 export const generateUserJWT = async (
   user_id: string,
-  team_id: string,
-  expiration: number = dayjs().add(7, "day").unix(),
+  expiration: number = dayjs().add(1, "minute").unix(),
```
andy-t-wang, line 100: why are we expiring this after 1 minute?

web/tests/integration/user.test.ts
```diff
@@ -253,10 +253,10 @@ describe("user role", () => {
 
     const response = await client.query({
       query,
+
       variables: {
         id: userFromTeam1?.user_id,
       },
-      context: { headers: { team_id: teams[1].id } }, // Pass a malicious header
```
andy-t-wang, line 259: this test will pass now because we are not spoofing anything. This test doesn't cover anything anymore

igorosip0v: Removed this one

hasura/metadata/databases/default/tables/public_user.yaml
andy-t-wang: Ah looks like it is null, we should drop this relationship in user then to team

andy-t-wang (approved):

PR #669, opened 2024-03-14
fix: team members can edit tos and privacy policy urls

igorosip0v:
Closes [WID-706](https://linear.app/worldcoin/issue/WID-706/team-members-can-edit-tos-and-privacy-policy-urls)

- Changes update permissions for user role on `action` table
- Adds tests

---

https://github.com/worldcoin/developer-portal/assets/89008845/736283e9-a230-46bc-93b9-85ef984dc3b3

### DIFF
--- a/web/tests/integration/action.test.ts
+++ b/web/tests/integration/action.test.ts
@@ -136,7 +136,7 @@ describe("user role", () => {
     expect(response.data.update_action_by_pk.id).toEqual(actionId);
   });
 
-  test("member can't update sign in with worldcoin action", async () => {
+  test("member can't update sign in with World ID privacy policy", async () => {
     const serviceClient = await getAPIServiceClient();
 
     const query = gql(`query GetUserId {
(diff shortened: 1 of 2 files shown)


web/tests/integration/action.test.ts
```diff
@@ -66,3 +67,139 @@ describe("service role", () => {
...
+    `;
+
+    const response = await client.mutate({
+      mutation,
+      variables: {
+        id: actionId,
+      },
+    });
+
+    expect(response.data.update_action_by_pk.id).toEqual(actionId);
```
andy-t-wang, line 136: to be honest it is better to have the test check the privacy policy and see that it matches "http://example.com"

web/tests/integration/action.test.ts
```diff
@@ -66,3 +67,139 @@ describe("service role", () => {
...
+  });
+
+  test("member can't update sign in with worldcoin action", async () => {
```
andy-t-wang, line 139:
```suggestion
  test("member can't update sign in with World ID", async () => {
```

andy-t-wang (approved):

web/tests/integration/action.test.ts
igorosip0v: Updated

PR #678, opened 2024-03-19
feat: /delete-account test

igorosip0v:


web/tests/integration/auth/delete-account.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  it("Should successfully delete account", async () => {
+    const mockReq = {
+      json: () => Promise.resolve(),
+    } as unknown as NextRequest;
+
+    (getSession as jest.Mock).mockResolvedValue({ user: validSessionUser });
+    const response = await deleteAccount(mockReq, {});
+    expect(getSession).toHaveReturned();
+    expect(response.status).toEqual(307);
+    console.log(response.headers.get("location"));
```
andy-t-wang, line 69:
```suggestion
```

andy-t-wang: We should also check that the data was indeed deleted beyond just the redirect

andy-t-wang (approved):

web/tests/integration/auth/delete-account.test.ts
igorosip0v: Do you mean in the test? Or in the route handler?
This handler is deleting it on the auth0 side.

andy-t-wang: ah I see, yeah that's fine. Though the test doesn't really cover much since we don't check auth0. Doesn't look like you can mock that anyways

PR #685, opened 2024-03-20
[fix] caching public apis

andy-t-wang:
Caching apis and update tests

### DIFF
--- a/web/api/public/app/[app_id]/index.ts
+++ b/web/api/public/app/[app_id]/index.ts
@@ -26,7 +26,7 @@ export async function GET(
   if (!isValidHostName(request)) {
     return NextResponse.json(
       {
-        error: "Invalid Request Origin, please use CDN",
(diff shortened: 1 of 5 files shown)


web/api/public/app/[app_id]/index.ts
```diff
@@ -22,6 +22,16 @@ export async function GET(
     app_id,
   });
 
+  // We only accept requests through the distribution origin
+  if (!isValidHostName(request)) {
+    return NextResponse.json(
+      {
+        error: "Invalid Request Origin, please use CDN",
```
Comment (pdtfh) [inline web/api/public/app/[app_id]/index.ts:29]:
```suggestion
        error: "Please use https://world-id-assets.com for this endpoint",
```
or something along those lines

igorosip0v (approved):

web/api/public/app/[app_id]/index.ts
Comment (andy-t-wang) [reply web/api/public/app/[app_id]/index.ts]: using env value

PR #693, opened 2024-03-22
feat: /api/v2/verify

0xPenryn:
Adds the `/api/v2/verify/[app_id]` endpoint, which deprecates the `credential_type` field and replaces `signal` with `signal_hash`.

`web/lib/verify.ts` is copied from `web/legacy/backend/verify.ts` and modified to support logic changes without modifying behavior of the v1 verify endpoint.


web/app/api/verify/[app_id]/route.ts
Comment (igorosip0v) [inline web/app/api/verify/[app_id]/route.ts:1]: @0xPenryn Don't we want to have `/api/v2/...` path to explicitly show API version? Just clarifying

Comment (0xPenryn) [reply web/app/api/verify/[app_id]/route.ts:1]: fixed, thanks for catching

hasura/migrations/default/1711395233596_update_null_nullifier_uses/down.sql
andy-t-wang, line 1: Let's squash the two migrations into one

andy-t-wang:
Lets write an API test for this as well. Should be some for v1 already we can re-use

igorosip0v (approved):

web/api/helpers/verify.ts
```diff
@@ -0,0 +1,284 @@
...
+    throw new Error("Unexpected response received from sequencer.");
+  }
+
+  return { success: true, status };
+};
+
+/**
+ * Checks whether the person can be verified for a particular action based on the max number of verifications
+ */
+export const canVerifyForAction = (
```
andy-t-wang, line 270: what's the reason for not deleting the other canVerifyForAction inside of api/helpers? Looks like the undefined check can be removed

0xPenryn: removed the undefined check

we could use this definition of canVerifyForAction for the v1/verify endpoint as well, but i think it's best to keep the v1 endpoint code/helpers in the legacy folder to prevent any unintended side effects of making changes for the v2 endpoints

PR #696, opened 2024-03-26
[Fix] HSTS Header

andy-t-wang:
Then we need to submit it here https://hstspreload.org/?domain=developer.worldcoin.org.

Based on what I see online we should set it to over a year. Examples I see suggest 2 years (63072000 seconds)

### DIFF
--- a/web/next.config.mjs
+++ b/web/next.config.mjs
@@ -42,6 +42,15 @@ const nextConfig = {
           },
         ],
       },
+      {
+        source: "/(.*)",
+        headers: [
+          {
+            key: "Strict-Transport-Security",
+            value: "max-age=63072000; includeSubDomains; preload",
+          },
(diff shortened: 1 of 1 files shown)


web/next.config.mjs
```diff
@@ -42,6 +42,15 @@ const nextConfig = {
           },
         ],
       },
+      {
```
pdtfh, line 45: wouldn't this cause issues with local development? perhaps best to set it in the deployment configuration

andy-t-wang: I think this only goes into effect if you add the website to a list above.

andy-t-wang: https://www.haydenbleasel.com/blog/next-hsts-preload

igorosip0v (approved):

PR #698, opened 2024-03-26
feat: oidc 3.1.3.2 redirect_uri validation for token

0xPenryn:
Updates `/api/v1/oidc/token` endpoint logic to comply with [OIDC specification section 3.1.3.2: Token Request Validation](https://openid.net/specs/openid-connect-core-1_0.html#TokenRequestValidation), specifically in regards to handling the `redirect_uri` parameter.

Tests included. Additionally, makes testing DB resets more robust.

Will require associated updates to documentation when deployed.


hasura/migrations/default/1711478492461_alter_table_public_auth_code_add_column_redirect_uri/up.sql
andy-t-wang, line 1: squash migrations

andy-t-wang (approved):

PR #699, opened 2024-03-27
Api v2 verify test

igorosip0v:
This PR: 
- Improves typing of nullifier in the `canVerifyForAction` function
  - This also removes check for the `null` for the uses field as we made this field not nullable
- Allows action to be `""` in `/v2/verify/` body validation schema
  - @0xPenryn @andy-t-wang Please let me know if this is wrong, I'll revert it and remove test case for this.
  - change is `yup.required` -> `yup.defined`
- Moves /verify endpoint code files from `/web/api/verify/...` to `/web/api/v2/verify/...`
  - I assume there will be more /v2/ endpoints in the future, let's keep endpoint code under v2 too.
- Fixes the way we are getting body in API route handler 
  - `const body = req.json()` instead of `req.body`
- Adds tests for `/v2/verify` endpoint


web/tests/api/v2/verify.test.ts
```diff
@@ -0,0 +1,508 @@
...
+// #endregion
+
+// #region Test data
+const getUrl = (app_id: string) =>
+  new URL(`/api/v2/verify/${app_id}`, "http://localhost:3000");
+
+const validBody = {
+  ...semaphoreProofParamsMock,
+  action: "verify",
+};
```
0xPenryn, line 80: unhashed signal shouldn't be included

```suggestion
const validBody = {
  ...semaphoreProofParamsMock,
  signal: undefined,
  action: "verify",
};
```

web/tests/api/v2/verify.test.ts
0xPenryn, line 1: we should add test cases for invalid semaphore proofs, for no signalHash provided (valid when `signal: ""` in IDKit), and for improper signalHash format (should be hex string of 66 characters, inclusive of `0x`)

web/api/v2/verify/index.ts
```diff
@@ -19,7 +19,7 @@ const schema = yup.object({
     .string()
     .strict()
     .nonNullable()
-    .required("This attribute is required."),
+    .defined("This attribute is required."),
```
0xPenryn, line 22: I think I'd prefer to keep this as `.required()` to prevent proof overlap between Incognito Actions and Sign In with World ID, pending approval from @andy-t-wang -- is there any way for us to tell if there are apps using `action: ""` for incognito actions?

If we do move forward with preventing empty actions, we should also update IDKit to make `action` a required parameter.

0xPenryn (changes requested):
looks great for the most part, but there are a few more test cases I think we should have and we need a final decision on how to handle `action: ""`

0xPenryn (changes requested):
looks great for the most part, but there are a few more test cases I think we should have and we need a final decision on how to handle `action: ""`

web/api/v2/verify/index.ts
igorosip0v: Updated.

web/tests/api/v2/verify.test.ts
igorosip0v: Updated

web/tests/api/v2/verify.test.ts
igorosip0v: @0xPenryn I've added test cases, but not quite sure about this one: 
> for no signalHash provided

In yup schema it simply has default value that will be applied if there is no signal_hash.

0xPenryn (approved):

PR #701, opened 2024-03-28
fix: remove current user membership

igorosip0v:
Closes [WID-836](https://linear.app/worldcoin/issue/WID-836/idor-devportal-team-owners-can-delete-themselves-off-of-their-own-team)


andy-t-wang:
This doesn't actually solve the error. Someone could just modify the graphql request and still delete themselves from curl.

andy-t-wang (approved):
We can merge this since this check is good but please don't close the issue. To resolve we should do what I mentioned in slack

PR #708, opened 2024-04-05
[Fix] api endpoints

andy-t-wang:
- Add description
- add team name
- Add return default if country doesn’t exist.


igorosip0v:
Please, fix the api tests

PR #707, opened 2024-04-05
Add consistency to URL validation

ph3t:
Remove redundant `uriHasJS()` in favor of `validateUrl()` to improve consistency and prevent XSS issues

PR #709, opened 2024-04-05
Replace XSS check with broader URL check

RedYetiDev:
> [!NOTE]
> While this may seem like a security patch, backend validation checks in the GraphQL already sanitize redirect URLs.

This PR replaces the case-sensitive `javascript:` check with a broader check which also verifies that the URL provided is a valid URL.

### OLD
`jAvAsCrIpT:alert(/xss/) -> true (valid)`
`javascript:alert(/xss/) -> false (invalid)`
`https://www.example.com/ -> true (valid)`
`htp:example -> true (valid)`

### NEW
`jAvAsCrIpT:alert(/xss/) -> false (invalid)`
`javascript:alert(/xss/) -> false (invalid)`
`https://www.example.com/ -> true (valid)`
`htp:example -> false (invalid)`

### DIFF
--- a/web/lib/utils.ts
+++ b/web/lib/utils.ts
@@ -68,7 +68,7 @@ export const validateEmail = (candidate: string): boolean => {
 
 export const isSSR = () => typeof window === "undefined";
 
-export const uriHasJS = (uri: string) => /javascript:/.test(uri);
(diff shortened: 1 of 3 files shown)


RedYetiDev:
Closing in favor of #707

PR #711, opened 2024-04-12
API endpoint update

andy-t-wang:
### Changes
- Add new field `world_app_button_text` that lets dev specify text they want show in the app default is use integration
- added new permission for reviewer role to update verified apps. This is set so we can remove apps from the app store via setting the `is_reviewer_app_store_approved` flag
- Add localise strings to platform=app
- Updated the return to include featured apps
- Update JSON to now parse the description when returning see spec for new format. 
- App and web return types are different
- Add the new button text field to links section
- Reduced world app description length to 35


igorosip0v (approved):

igorosip0v:
@andy-t-wang Please, fix the integration test

PR #715, opened 2024-04-17
refactor: precheck dynamic action creation

igorosip0v:
Should close [WID-877](https://linear.app/worldcoin/issue/WID-877/disable-on-the-fly-action-generation-[security])

- Moves dynamic action creation to the /verify


web/api/v2/verify/index.ts
```diff
@@ -44,6 +48,7 @@ const schema = yup.object({
     )
     .strict()
     .optional(),
+  api_key: yup.string().strict().optional(),
```
andy-t-wang, line 51: Api keys should be passed in headers `'Authorization': 'ApiKey YOUR_API_KEY'`

andy-t-wang:
What's the difference between v1 and v2 of verify?

I think this breaks the workflow of dynamic actions. Dynamic actions are meant to be on the fly assumedly client side when a user is interacting with an integration. Precheck will always be called since there's no way to verify without a proof and you need the app to do this

andy-t-wang:
Lets talk in standup on solutions. Perhaps we just disable this for now not clear what even the usecases are.

andy-t-wang:
Update from sync today. We are making this a seperate endpoint v2/team/create-action or something like that where we allow users using an api key to create actions for an app id. Moving logic out of verify

web/api/v2/create-action/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  if (!isValid) {
+    return handleError(req);
+  }
+
+  const { app_id } = parsedParams;
+  const base64ApiKey = Buffer.from(api_key, "base64").toString("utf-8");
+  const [id, secret] = base64ApiKey.split(":");
+  const client = await getAPIServiceGraphqlClient();
```
andy-t-wang, line 42: Since we are using api key we need to be generating an API JWT so we use those permissions

andy-t-wang (commented):
Need to make a test for this route

web/api/v2/create-action/index.ts
```diff
@@ -0,0 +1,153 @@
...
+import {
+  CreateDynamicActionMutation,
+  getSdk as createDynamicActionSdk,
+} from "./graphql/create-dynamic-action.generated";
+import { getSdk as fetchApiKeySdk } from "./graphql/fetch-api-key.generated";
+
+const createActionBodySchema = yup.object({
+  app_id: yup.string().strict().required(),
+  action: yup.string().strict().required(),
+  team_id: yup.string().strict().required(),
```
igorosip0v, line 18: @andy-t-wang I think it will be more consistent to get this from the URL because we are doing it like this in other endpoints. But in this case, I think that this endpoint is more related to the app. Thus URL should look like this: `/api/v2/team/[team_id]/app/[app_id]/create-action`. But it will be too long. 

Or we can just leave it as it is and get this data from the body. 

What do you think

andy-t-wang: no lets get it from the body

web/api/v2/create-action/index.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  const { action, app_id } = {
+    ...body,
+    ...params,
+  };
+
+  const keyValue = api_key.replace(/^api_/, "");
+  const base64ApiKey = Buffer.from(keyValue, "base64").toString("utf8");
+  const [id, secret] = base64ApiKey.split(":");
+  const serviceClient = await getAPIServiceGraphqlClient();
```
igorosip0v, line 77: @andy-t-wang Had to use service client here to fetch API key data, as we agreed to get team_id from the key id. 

Without team_id we can't create api_key client because we need team_id in hasura jwt payload.

If we don't want to use service client at all in this endpoint we need to accept team_id via body

web/api/v2/create-action/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  let insertedAction: CreateDynamicActionMutation["insert_action_one"] | null =
+    null;
+
+  try {
+    const { insert_action_one } = await createDynamicActionSdk(
+      apiKeyClient,
+    ).CreateDynamicAction({
+      app_id: app_id,
+      action: body.action,
+      external_nullifier,
```
igorosip0v, line 149: @andy-t-wang Should we also accept some additional params in this endpoint like `max_verification`?

andy-t-wang: ah yeah that's a good idea

hasura/metadata/databases/default/tables/public_api_key.yaml
```diff
@@ -21,6 +21,15 @@ insert_permissions:
         - name
         - team_id
 select_permissions:
+  - role: api_key
+    permission:
```
andy-t-wang, line 25: Do we still need these permissions? We are using service client to check this right?

andy-t-wang (approved):

hasura/metadata/databases/default/tables/public_api_key.yaml
igorosip0v: Removed

web/tests/api/v2/create-action.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    VerifyFetchAPIKey.mockResolvedValue(validApiKeyResponse);
+
+    const res = await POST(mockReq, ctx);
+    expect(res.status).toBe(401);
+  });
+
+  it("returns 404 if api key not exists", async () => {
+    const mockReq = createMockRequest({
+      url: getUrl(validAppId),
+      api_key: "invalid-token",
```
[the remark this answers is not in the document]
igorosip0v: Test value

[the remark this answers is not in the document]
igorosip0v: Fixed btw

web/api/v2/create-action/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+const createActionParamsSchema = yup.object({
+  app_id: yup.string().strict().required(),
+});
+
+export const POST = async (
+  req: NextRequest,
+  { params: rawParams }: { params: { app_id: string } },
+) => {
+  const api_key = req.headers.get("authorization")?.split(" ")[1];
```
andy-t-wang, line 33: One last thing since this will become the base template for our other api endpoints. We should specify that the user does 

Authorization: Bearer `api_token`. I believe here we dont' handle bearer. This seems to be compliant with openapi spec

andy-t-wang: merged

PR #724, opened 2024-05-02
Update Docker, upgrade nextjs, fix a delete bug

andy-t-wang:
Update to the app metadata query broke the delete process. Updating here to fix it
- Updated the docker image to be more efficient as part of dev portal migration
- Updated nextjs version to handle the server action null issue as well as dd logs which will create request issues later
completes WID-881

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -30,7 +30,7 @@
     "jotai": "^2.6.4",
     "jsdom": "^24.0.0",
     "json5": "^2.2.3",
-    "next": "14.1.0",
+    "next": "14.2.3",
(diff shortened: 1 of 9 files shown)


igorosip0v (approved):

web/package.json
```diff
@@ -30,7 +30,7 @@
     "jotai": "^2.6.4",
     "jsdom": "^24.0.0",
     "json5": "^2.2.3",
-    "next": "14.1.0",
```
pdtfh, line 33: let's add an engines statement to make sure it can only be run with `node = 20.x`, and pnpm

PR #737, opened 2024-05-14
Feat: Transaction Endpoint for Mini Apps

andy-t-wang:
This endpoint is used by web2 apps to confirm the transaction was successfully mined on chain. Developers will use their api key in the dev portal in order to call this endpoint. Developers will only be able to see transaction data for apps that they own the API key for. 

Developers will call this via `developer.worldcoin.org/v1/minikit/transactions/transaction_id?app_id=12312`.
Upon successful completion this will return a singular transaction object back to the user in the form of 
[transaction_data]

### DIFF
--- a/web/api/v2/minikit/transaction/[transaction_id]/graphql/fetch-api-key.generated.ts
+++ b/web/api/v2/minikit/transaction/[transaction_id]/graphql/fetch-api-key.generated.ts
@@ -1,0 +1,76 @@
+/* eslint-disable import/no-relative-parent-imports -- auto generated file */
+import * as Types from "@/graphql/graphql";
+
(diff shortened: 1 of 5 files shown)


0xPenryn:
@andy-t-wang what's the reasoning behind gating tx status with an api key when the tx is publicly viewable on-chain?

andy-t-wang:
We only want people using this if they are web2 apps and don't want to set up an RPC all web3 apps should just query on chain. We are emitting an event. 

API Key is meant to enforce that you only query your own transactions. We don't want people using this endpoint outside of that use case since it will add load to our app-backend

0xPenryn (approved):

PR #754, opened 2024-05-18
[temp fix] unblock actions

andy-t-wang:
Will need to add creation_mode = dynamic back


andy-t-wang:
Actually prefer not to fix things this way

PR #763, opened 2024-06-03
feat: static kiosk link

0xPenryn:
Creates a kiosk page that can be shared at a static URL, and is compatible with Apple's "Add to Home Screen" functionality.

Primary purpose is to support kiosks on iPads.


andy-t-wang (approved):
Approving. Though since we aren't doing the public access we could technically just make the existing one static at a URL

0xPenryn:
> Approving. Though since we aren't doing the public access we could technically just make the existing one static at a URL

Eventually i'd like to make kiosks optionally public, since the people using/operating the kiosk are very often not developers, and there's no benefit to me verifying at a kiosk on my own device, since i just prevent myself from verifying at the intended kiosk

so i'll leave this as-is for now and we can make the changes to allow for public kiosks later

PR #771, opened 2024-06-06
Add Create an App automated test

bdoof:
[The scenario is described in Qase.](https://app.qase.io/case/DP-2)
Hasura helpers were added along with the test to prepare and delete test data.

### DIFF
--- a/web/tests/e2e/helpers/hasura/user/create-user.ts
+++ b/web/tests/e2e/helpers/hasura/user/create-user.ts
@@ -1,5 +1,4 @@
 import { adminGraphqlClient } from "@e2e/helpers/hasura";
-import casual from "casual";
 import { InsertUserMutation, insertUserSdk } from "./graphql";
 
 export const createUser = async (params: {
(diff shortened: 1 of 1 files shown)


web/tests/e2e/helpers/hasura/admin-graphql-client.ts
```diff
@@ -0,0 +1,10 @@
+import { GraphQLClient } from "graphql-request";
+
+export const adminGraphqlClient = new GraphQLClient(
+  process.env.NEXT_PUBLIC_GRAPHQL_API_URL!,
+  {
+    headers: {
+      "x-hasura-admin-secret": process.env.HASURA_GRAPHQL_ADMIN_SECRET!,
```
andy-t-wang, line 7: Where will we be getting this admin secret from? Also this function will not be callable? Outside of tests?

andy-t-wang: Can you share how this will be set up? Will you trigger a docker image at start and set the env.test value? otherwise I don't see how admin HASURA_GRAPHQL_ADMIN_SECRET is being set. It's not defined in env.test and probably shouldn't be until runtime if you are creating a new instance and DB on the fly

andy-t-wang (approved):

PR #779, opened 2024-06-12
App Store V2 Endpoints

igorosip0v:
This PR: 
- Adds `/v2/public` endpoints
- Explicitly sets types for typescript gql types autogen that previously was `any`.
  - This should improve types and help avoid type errors in future.

### DIFF
--- a/web/graphql-codegen.types.js
+++ b/web/graphql-codegen.types.js
@@ -18,8 +18,8 @@ module.exports = {
       jsonb: "any",
       json: "any",
       numeric: "number",
-      _text: "any"
-    }
+      _text: "any",
+    },
   },
 
   generates: {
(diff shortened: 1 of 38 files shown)


web/graphql-codegen.types.js
```diff
@@ -10,6 +10,18 @@ const prettier = require("prettier");
...
   ignoreNoDocuments: true,
+  config: {
+    defaultScalarType: "unknown",
+    scalars: {
+      timestamp: "string",
+      timestamptz: "string",
+      jsonb: "any",
+      json: "any",
+      numeric: "number",
+      _text: "any",
```
igorosip0v, line 21: @andy-t-wang I've noticed that for some DB types autogen is making `any` type, which can be improved with this config. For example `supported_countries` field has `text[]` (or `_text`) in the db, which should have `Array<string>` type in the TS, but autogen is making it `any`. 

I've added this config, but some types I still left as `any` because if I'm trying to type it correctly it invokes TS errors in other parts of the application. Had no time to fix them all.

andy-t-wang:
I made some changes. Please look at the commits. 

Overall was formatting, added a slight test

andy-t-wang (approved):

PR #782, opened 2024-06-13
Upgrade Hasura version and fix types

igorosip0v:
- Updates hasura image version
- Updates types

### DIFF
--- a/docker-compose-test.yaml
+++ b/docker-compose-test.yaml
@@ -14,7 +14,7 @@ services:
       retries: 5
 
   hasura_test:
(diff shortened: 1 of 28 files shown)


andy-t-wang:
For your review tom: 

I removed the Array.isArray checks since we allow nullable for all arrays and [] is truthy. Thus if it's null or undefined we will catch it already. If you added these for another reason let me know. 

I also fixed a type for convertArrayToHasuraArray

For the Showcase images I just pass in the array directly now too

andy-t-wang (approved):

PR #792, opened 2024-06-20
feat: switch from RS256 to PS256 for OIDC JWT

0xPenryn:
Updates JWTs used in Sign In with World ID to use PS256 instead of RS256 for signing. No changes made to key generation logic, as both RS256 and PS256 use RSA keys. Must be tested in staging before deploying to production.

Not intended to impact Hasura JWT authentication.


andy-t-wang (approved):
Disabled prod

PR #794, opened 2024-06-21
fix: prevent OIDC /token endpoint race condition

0xPenryn:
Uses one atomic GraphQL mutation to return and delete the `auth_code` when accessing the OIDC `/token` endpoint.
completes sec-697
Not yet tested.

### DIFF
--- a/web/pages/api/v1/oidc/token.ts
+++ b/web/pages/api/v1/oidc/token.ts
@@ -51,7 +51,13 @@ const deleteAuthCodeQuery = gql`
         auth_code: { _eq: $auth_code }
       }
     ) {
-      affected_rows
+      nullifier_hash
+      verification_level
+      scope
+      code_challenge
+      code_challenge_method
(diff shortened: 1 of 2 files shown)


ph3t:
@0xPenryn this change makes sense, but if two requests are sent concurrently I'm not sure this change will suffice to fix the underlying issue. Would it make sense to add a row-level DB lock for more robustness?

andy-t-wang:
cc @m1guelpf might be worth getting your input, IIRC you implemented something similar to prevent verify race conditions

andy-t-wang:
I will merge this change. Atomicity here is preserved by the mutations. Previously the delete even if it deleted no values would continue. Since we are deleting and using the code we deleted this should prevent the race condition.

Hasura treats each mutation requests as atomic https://www.restack.io/docs/hasura-knowledge-hasura-transactions-guide

andy-t-wang (approved):

PR #796, opened 2024-06-24
Legacy Refactor Pt 1: Images endpoints

andy-t-wang:
Start of the migration for legacy left in dev portal. 

- Move images endpoints from legacy into app router
- 1 min default for all JWTs
- Add yup validation for all endpoints
- Delete old routes
- Update Hasura Action
- Version is deprecated so removed it
- Rename images to app-profile


web/api/app-profile/get-app-review-images/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (!protectInternalEndpoint(req)) {
+    return;
+  }
+
+  if (!req.method || req.method !== "POST") {
+    return errorNotAllowed(req.method, req);
+  }
+
+  const body = await req.json();
+  if (body?.action.name !== "get_app_review_images") {
```
igorosip0v, line 39: @andy-t-wang Can be done with yup too, I think

igorosip0v: If you decide to update it, don't forget other actions too

web/api/app-profile/upload-image/index.ts
```diff
@@ -1,41 +1,33 @@
...
+export const POST = async (req: NextRequest) => {
   try {
-    if (!protectInternalEndpoint(req, res)) {
+    if (!protectInternalEndpoint(req)) {
       return;
     }
 
-    if (req.method !== "POST") {
-      return errorNotAllowed(req.method, res, req);
+    if (!req.method || req.method !== "POST") {
```
igorosip0v, line 24: @andy-t-wang I think we can leave it explicitly, but I also think that nextjs route handlers now validate method by function name (`POST` in this case)

igorosip0v: Can you please double-check it?

igorosip0v: And if it is so, then I think Nextjs will throw an error earlier, before even starting to execute endpoint code

igorosip0v (approved):

PR #804, opened 2024-06-26
Cleanup: Removing some endpoints we have no documentation for anymore

andy-t-wang:
Don't think these endpoints are useful anymore. Also we no longer have documentation for them. Potentially worth removing the cache table as well since it looks like it's just used for storing these values?

cc @m1guelpf @0xPenryn


0xPenryn (approved):
approved, only usage i found in our entire github org is from the old and now-archived templates

PR #807, opened 2024-06-27
Remove _gen-jwks

andy-t-wang:
Gen JWKS is no longer used. Based on blame it seems like before we made this to be a cron job to rotate the JWKs for OIDC but since we added that logic to authorize and auto expire the keys.


pdtfh (approved):
right, new keys are generated whenever expiration is close

PR #810, opened 2024-06-28
Disable Delete Sign in Action

andy-t-wang:
We don't have a restriction on deletion for sign in with world id. This fixes that. While this security wise doesn't cause an impact it does create a functionality bug where if you delete the sign in action then the page wont' load

### DIFF
--- a/docker-compose-test.yaml
+++ b/docker-compose-test.yaml
@@ -1,5 +1,3 @@
-version: "3"
-
 services:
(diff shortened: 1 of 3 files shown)


ik-work-gh:
LGTM 
image

ik-work-gh (approved):

igorosip0v (approved):

PR #821, opened 2024-07-02
Add Create an Action test

bdoof:
Added Create an Action automated test as per [the scenario in Qase](https://app.qase.io/case/DP-3).


igorosip0v (approved):
LGTM but let's wait also for Andy's approval.

andy-t-wang:
Why are there so many queries added like create and delete app. is this just missing rebase?

PR #844, opened 2024-07-11
DEV-897 Add Qase reporter for end-to-end tests

bdoof:
Now end-to-end tests send their run results from CI to Qase, so you can get verbose failure instantly - videos, screenshots and scenario descriptions.

image

Test run results link is always posted as a job output:

image


andy-t-wang (approved):

pdtfh:
@bdoof can we get rid of this unmaintained dependency?

PR #850, opened 2024-07-14
Update directory with locales

andy-t-wang:
Add locale pulling and default to english if the locale is not specified

### DIFF
--- a/web/api/v2/minikit/app-metadata/[app_id]/index.ts
+++ b/web/api/v2/minikit/app-metadata/[app_id]/index.ts
@@ -91,9 +91,13 @@ export const GET = async (
 
   const client = await getAPIServiceGraphqlClient();
 
+  const headers = req.headers;
+  const locale = headers.get("x-accept-language") ?? "";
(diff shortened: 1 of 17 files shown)


web/api/v2/minikit/app-metadata/[app_id]/index.ts
```diff
@@ -91,9 +91,13 @@ export const GET = async (
 
   const client = await getAPIServiceGraphqlClient();
 
+  const headers = req.headers;
+  const locale = headers.get("x-accept-language") ?? "";
```
Comment (igorosip0v) [inline web/api/v2/minikit/app-metadata/[app_id]/index.ts:95]: @andy-t-wang Are you sure these header values have the same format we are using for locales? Just clarifying

igorosip0v (approved):

web/api/v2/minikit/app-metadata/[app_id]/index.ts
Comment (andy-t-wang) [reply web/api/v2/minikit/app-metadata/[app_id]/index.ts:95]: Confirming with android team

PR #889, opened 2024-08-08
Allow local host testing

andy-t-wang:
Allow local host testing


andy-t-wang:
Regex was unfortunate but we still have some checks client side

igorosip0v (approved):

PR #892, opened 2024-08-12
test transaction error

andy-t-wang:
Test transaction error


andy-t-wang:
Only in production do we get forbidden errors. I'm just merging this to test if it's firewall issues

PR #902, opened 2024-08-27
Add associated domain support

andy-t-wang:
Add associated domain support to minikit


web/scenes/Portal/Teams/TeamId/Apps/AppId/Profile/Setup/page/SetupForm/index.tsx
```diff
@@ -501,6 +511,23 @@ export const SetupForm = (props: LinksFormProps) => {
             }
             placeholder="0x12312321..., 0x12312312..."
             register={register("whitelisted_addresses")}
+            onChange={(e) => {
```
igorosip0v, line 514: @andy-t-wang Looks like this function is used multiple times, I think we can move it under single function above.

igorosip0v (approved):

web/scenes/Portal/Teams/TeamId/Apps/AppId/Profile/Setup/page/SetupForm/index.tsx
andy-t-wang: ah yeah

PR #932, opened 2024-09-23
Unverified Metadata

andy-t-wang:
We need to always fetch unverified metadata for the test mini app /api


andy-t-wang:
admin merging

igorosip0v (approved):

PR #937, opened 2024-09-24
Fix/menu dropdown styling

michalstruck:
Simple fix, also added a Makefile that makes using the hasura console a lil easier, updated the readme to highlight its usage


andy-t-wang:
Can you include screenshots in front end changes

andy-t-wang (approved):

PR #953, opened 2024-10-04
DEV-1112 app reports & bans

michalstruck:
setup app_report table, add is_banned to app_metadata, ban_app action


hasura/migrations/default/1728039437362_add_is_banned_metadata/up.sql
andy-t-wang, line 1: I think this is better to put at the app table not the metadata since each app can have two metadata rows

web/api/hasura/create-new-draft/graphql/fetch-metadata.graphql
andy-t-wang, line 1: I think we should indicate if this app is banned in the UI. Can you reach out to Kai

andy-t-wang (changes requested):

andy-t-wang (approved):

PR #956, opened 2024-10-07
upgrade ethers

andy-t-wang:
Upgrade ethers


andy-t-wang:
Upgrade ethers to v6 which closes some dependency vulnerabilities

PR #971, opened 2024-10-10
Michalstruck dev 1144 add analytics page

michalstruck:
Reworked the dashboard


web/api/v2/public/apps/index.ts
```diff
@@ -14,6 +14,14 @@ import {
 } from "./graphql/get-app-rankings.generated";
 import { getSdk as getHighlightsSdk } from "./graphql/get-app-web-highlights.generated";
 
+export type GetAppsResponse = {
+  app_rankings: {
```
andy-t-wang, line 18: I think you can import the response type of this directly from the graphql generated file

andy-t-wang (approved):

andy-t-wang:
Also please add screenshots in the future

PR #984, opened 2024-10-15
remove use of status in sequencer response

gracew:
From the protocol team:
> We should also remove the explicit use of status field from the response and rely on HTTPS status codes.


andy-t-wang (approved):

andy-t-wang:
Thanks!

PR #985, opened 2024-10-15
fix token usd amount calculation logic

michalstruck:
improve amount calculation a bit and account for token precision


andy-t-wang (approved):
Could be fine to keep working logs

PR #1000, opened 2024-10-23
Fix new users stat

michalstruck:
This fixes two things, first of all the transaction amount is in USD and had 3 decimals, rounded to 2 now.

Secondly, switched up the default dashboard selector from all time to weekly. All time new users = all time users (we already show that), so maybe let's show meaningful data? The drawback is a smaller number...

Screenshot 2024-10-23 at 17 54 45

Screenshot 2024-10-23 at 17 54 39


michalstruck:
Screenshot 2024-10-23 at 19 04 13
after update we changed the order

andy-t-wang (approved):

PR #1025, opened 2024-10-31
switch camelCase to snake_case in app_metadata

michalstruck:


andy-t-wang (approved):

andy-t-wang:
LGTM. I removed the camelCase ones we should just have everyone change over otherwise it will be confusing

PR #1042, opened 2024-11-13
Copy and error msg fixes

michalstruck:
After submitting on `App Store` tab, we show an error if there isn't a logo image
Screenshot 2024-11-13 at 15 11 57


andy-t-wang:
tbh I think you should make this error look nicer so it's flush and one line. The wrapping here looks weird

andy-t-wang:
separate PR is fine

andy-t-wang (approved):

PR #1043, opened 2024-11-14
Verify Race Condition

andy-t-wang:
Currently verify has a race condition where if you send two or more requests at the same time, the check on canVerifyForAction will pass and if the proof is being issued for the first time, it will cause the second call to return a duplicate nullifier constraint error. 

This is bad behavior imo since technically the only thing blocking users should be the number of uses specified for that action. If you send two verifies with the same proof and you have the uses set to more than 1, then both should succeed. For instance, we have this issue on usernames now where if the user pushes the button twice quickly it will trigger the flow to send the register twice. While devs should protect against this, the verify endpoint shouldn't fail in this situation either. 

Hasura doesn't support mutations for native queries or stored procedures and upsert doesn't allow using the atomic _inc operation. See https://hasura.io/docs/2.0/schema/postgres/native-queries/#mutations.

Thus, to solve this, I add a constraint on uses for the nullifier equal to the action max_verifications param on update. And now the query performs a pseudo upsert that isn't fully atomic but the update itself due to _inc is atomic and the new constraint will make it functionally atomic. 

Thus now the behavior where a user submits a proof for an action twice in a row doesn't result in the second one just failing but now just adds an additional use increment if the action permits


web/api/v2/verify/graphql/atomic-upsert-nullifier.graphql
```diff
@@ -0,0 +1,20 @@
+mutation AtomicUpsertNullifier($action_id: String!, $nullifier_hash: String!) {
```
andy-t-wang, line 1: Here because we cannot do _inc with upsert. We insert an action with an empty value, then increment. This solves the error of duplicate constraints since if it sees a duplicate it just does nothing as update_columns is empty

andy-t-wang: We dont rely on the uses value coming in anymore since that would fail if two proofs are submitted together as the uses comes from a previous graphql query. We should count every valid incoming request

m1guelpf (approved):
very nice

PR #1048, opened 2024-11-15
Cache App Rating Calculation

andy-t-wang:
Currently app_rating is dynamically calculated from the table with each call. This is obviously horribly expensive and so we should only do this once every 24 hours for each app. Pull out app_rating calculation then from the graphql so it doesn't get fetched and called


web/api/v2/minikit/app-metadata/[app_id]/index.ts
```diff
@@ -116,7 +116,7 @@ export const GET = async (
 
   return NextResponse.json(
     {
-      ...formatAppMetadata(appMetadata, [], locale),
+      ...(await formatAppMetadata({ ...appMetadata }, [], locale)),
```
Comment (michalstruck) [inline web/api/v2/minikit/app-metadata/[app_id]/index.ts:119]: unnecessary spread `{ ...appMetadata }`

web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -82,7 +82,11 @@ export async function GET(
     }, null);
   }
 
-  let dataToReturn = formatAppMetadata(parsedAppMetadata, metricsData, locale);
+  let dataToReturn = await formatAppMetadata(
+    { ...parsedAppMetadata },
```
Comment (michalstruck) [inline web/api/v2/public/app/[app_id]/index.ts:86]: unnecessary spread { ...appMetadata }

PR #1049, opened 2024-11-15
fix: add localisation arrows, save on change

michalstruck:
Screenshot 2024-11-15 at 21 51 13

arrows allow to switch left to right & vice versa, wrapping from last to first & vice versa
on top of that they indicate that there's some switching to be done - users will hopefully instinctively know how localisations work

edit: Most importantly, localisations are now saved between switches, so no data loss

### DIFF
--- a/web/components/TextArea/index.tsx
+++ b/web/components/TextArea/index.tsx
@@ -73,7 +73,7 @@ export const TextArea = memo(function TextArea(props: TextAreaInterface) {
   );
 
   return (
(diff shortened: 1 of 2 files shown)


andy-t-wang:
tbh I think it would be good to make some of them red or something if the localisation is not filled out

andy-t-wang:
but we can start with this and see how it goes

andy-t-wang (approved):

PR #1056, opened 2024-11-18
add instrumentation file

igorosip0v:
This PR adds redis client usage with instrumentation file.

Instrumentation file docs: 
https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation

Shortly (but read the doc), we use `register()` inside the instrumentation file to execute Redis client creation only once on the application starting after the build. Then, we store the Redis client instance globally to access it everywhere. 

### TODO 
- PR is not finished. We should apply global client everywhere where we are using redis client.

### DIFF
--- a/web/@types/global.d.ts
+++ b/web/@types/global.d.ts
@@ -1,0 +1,7 @@
+import { Redis, Cluster as RedisCluster } from "ioredis";
+
+type Messages = typeof en;
+
+declare global {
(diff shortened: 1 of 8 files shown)


andy-t-wang (approved):

andy-t-wang:
Will add to ratings soon, after a slight refactor on calculation

andy-t-wang:
Updated to use Redis globally and for tests

PR #1058, opened 2024-11-20
fix: prevent excessive warns

michalstruck:
early return if nil or empty string


andy-t-wang (approved):

andy-t-wang:
I suspect we will still need to investigate why some places return null

andy-t-wang (approved):

PR #1071, opened 2024-11-27
Fix instrumentation file with redis

andy-t-wang:
Should not close redis connection


andy-t-wang:
just for you to look, forgot to remove the quit so I closed my only thread zzz

PR #1075, opened 2024-11-27
500kB images possible

michalstruck:
still it's only client side afaik, will be fixed in the bigger sec update


andy-t-wang:
We have a restriction in api/hasura/upload-image. The signed url we get has a max of 250 please update that too

andy-t-wang:
Also the copy for uploading says 250kb, needs to be update

andy-t-wang (changes requested):
See above

michalstruck:
oops my bad, fixed

PR #1076, opened 2024-11-27
logo error & form logic fix

michalstruck:
logo error:
Screenshot 2024-11-27 at 18 11 57
also fixed some values resetting on save & inability to select a country on a new app

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/AppStoreLocalised/index.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/AppStoreLocalised/index.tsx
@@ -120,6 +120,7 @@ export const AppStoreForm = (props: {
     defaultValues: {
       ...appMetadata,
       ...description,
+      supported_countries: appMetadata?.supported_countries ?? [],
     },
   });
 
@@ -441,9 +442,7 @@ export const AppStoreForm = (props: {
                 <SelectMultiple
                   values={field.value}
                   onRemove={(value) =>
-                    field.onChange(
-                      field.value?.filter((v) => v !== value) ?? [],
-                    )
+                    field.onChange(field.value.filter((v) => v !== value) ?? [])
                   }
                   items={countries}
                   label=""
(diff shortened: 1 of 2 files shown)


web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/AppStoreLocalised/index.tsx
```diff
@@ -140,19 +141,13 @@ export const AppStoreForm = (props: {
       locale !== "en" ? localisedData?.localisations?.[0] : appMetadata;
 
     reset({
+      ...getValues(),
       name: localisedItem?.name ?? "",
       short_name: localisedItem?.short_name ?? "",
       world_app_description: localisedItem?.world_app_description ?? "",
       world_app_button_text: localisedItem?.world_app_button_text ?? "",
       supported_languages: appMetadata?.supported_languages ?? [],
       app_website_url: appMetadata?.app_website_url ?? "",
```
andy-t-wang, line 150: The reason we did this is that support link is the field in DB and it's always a link we just turn an email to a link by adding mailto: I think this will remove that ability.

andy-t-wang (changes requested):

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/AppStoreLocalised/index.tsx
michalstruck: to me this looks like we're turning a mailto link to an email here (for display?)

michalstruck: what you're talking about happens on L259

and also I think that was wrong bc why would support_email value depend on support_link value

nah i'd leave it as is, the form behaves fine I believe

andy-t-wang: I dont think there's a support email field in the db. so if you had an email inputted instead we would parse the suport_link value in the db to show it instead of a link

michalstruck: okay then, i'll parse it when setting defaults for the form (in hook args) and it'll propagate in subsequent resets

PR #1090, opened 2024-12-13
fix: provide compatibility layer for report form

michalstruck:
This is needed so that there aren't any interruptions in the app reporting flow. Other PR introduces breaking changes, which would result in rejected inserts. This ensures that reports always go through in the transition period


web/api/hasura/tmp-insert-app-report/index.ts
andy-t-wang, line 1: Lets add a TODO and a linear task on when this needs to be deprecated. This is gonna be confusing to anyone looking at the code base what's going on. also please add a PR description

andy-t-wang (approved):

web/api/hasura/tmp-insert-app-report/index.ts
michalstruck: this will be removed immediately after the app report form is updated
added in linear for good measure

PR #1101, opened 2024-12-19
fix: add emoji to title regex

michalstruck:
requested by devs to allow emojis in titles and i don't see a reason not to


andy-t-wang (approved):

michalstruck:
After some thought, it may not be a good idea to allow emojis, no other app store allows it, they are inconsistent across platforms, may be ambiguous

PR #1117, opened 2025-01-07
Remove unused params

andy-t-wang:
remove unused params from backend endpoint

### DIFF
--- a/web/api/v2/minikit/send-notification/index.ts
+++ b/web/api/v2/minikit/send-notification/index.ts
@@ -229,6 +229,7 @@ export const POST = async (req: NextRequest) => {
         title,
         message,
         miniAppPath: mini_app_path,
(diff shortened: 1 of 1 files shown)


web/api/v2/minikit/send-notification/index.ts
```diff
@@ -263,5 +264,9 @@ export const POST = async (req: NextRequest) => {
     message,
   );
 
-  return NextResponse.json({ success: true, status: 200, ...data });
+  return NextResponse.json({
+    success: true,
+    status: 200,
+    result: data.results,
```
michalstruck, line 270: would be a good idea to add validation of the expected data shape

andy-t-wang: ah yes good point

PR #1118, opened 2025-01-08
fix: add more special characters to long form text regex

michalstruck:
after testing with some more sample texts in diferent languages, i included more common and harmless characters


michalstruck:
api test broken in different pr

andy-t-wang (approved):

PR #1134, opened 2025-01-15
Regex validation for message

andy-t-wang:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
Adding regex validation for message and title in the `sendNotificationBodySchema` object in the `web/api/v2/minikit/send-notification/index.ts` file.

## Checklist

### DIFF
--- a/web/api/v2/minikit/send-notification/index.ts
+++ b/web/api/v2/minikit/send-notification/index.ts
@@ -22,8 +22,24 @@ const sendNotificationBodySchema = yup.object({
     .min(1)
     .max(1000)
     .required("wallet_addresses is required"),
-  message: yup.string().strict().required().max(200),
-  title: yup.string().strict().optional().max(30),
+  message: yup
+    .string()
+    .strict()
+    .required()
+    .max(200)
+    .matches(
+      /^[\p{L}\p{N}\p{P}\p{Z}]+$/u,
+      "Message can only contain letters, numbers, punctuation, and spaces",
+    ),
+  title: yup
+    .string()
+    .strict()
+    .optional()
+    .max(30)
+    .matches(
+      /^[\p{L}\p{N}\p{P}\p{Z}]+$/u,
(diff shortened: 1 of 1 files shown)


web/api/v2/minikit/send-notification/index.ts
```diff
@@ -22,8 +22,24 @@ const sendNotificationBodySchema = yup.object({
...
     .required("wallet_addresses is required"),
-  message: yup.string().strict().required().max(200),
-  title: yup.string().strict().optional().max(30),
+  message: yup
+    .string()
+    .strict()
+    .required()
+    .max(200)
+    .matches(
+      /^[\p{L}\p{N}\p{P}\p{Z}]+$/u,
```
michalstruck, line 31: do we want to allow all punctuation here? This will also match special chars like ` ~ < > etc
Asking from a security viewpoint

michalstruck (commented):
left comment

web/api/v2/minikit/send-notification/index.ts
andy-t-wang: I dont think ~<> matches no? https://regex101.com/ I tried it here and it blocks

andy-t-wang: OK check now allows currency and .

michalstruck: @andy-t-wang actually i checked it wrong before, sorry
this is cool too but certain punctuations will be definitely requested later (`-.,¡!¿?।:'"()`) so might as well add {P} back

michalstruck (approved):

web/api/v2/minikit/send-notification/index.ts
andy-t-wang: hmm fail

andy-t-wang: lets jsut leave for now and see if we get asks cause I dont want to allow brackets {}

PR #1136, opened 2025-01-15
Fix app rating upsert issue

Takaros999:
## PR Type

- [ ] Regular Task
- [X] Bug Fix
- [ ] QA Tests

## Description

This PR fixes an issue with app ratings and adds a migration script to recalculate app ratings.

### Issue
Currently, we're using a rolling sum to update app ratings. We allow the user to upsert his review and apply the rating difference to the app's rating sum. The **issue** is that we're updating the rating count each time the user updates his review, this means that if multiple user decide to change their app rating they can affect the total app rating score, because the rating count will increase while app reviews stay the same.

### Migration script
Added a migration script to recalculate app ratings.

## Checklist


hasura/migrations/default/1737035548134_recalculate_app_ratings/up.sql
```diff
@@ -0,0 +1,12 @@
+UPDATE app
+SET
+    rating_sum = (
+      SELECT COALESCE(SUM(rating), 0)
+      FROM app_reviews
+      WHERE app_reviews.app_id = app.id
+    ),
+    rating_count = (
```
andy-t-wang, line 8: Good catch we only need to update rating_count though. No need to do sum as well. Will make this operation slightly cheaper

andy-t-wang (approved):
Make that change and we should be good. I would watch the DB though it could need to be scaled up for this

hasura/migrations/default/1737035548134_recalculate_app_ratings/up.sql
andy-t-wang: actually I'll just make this change for you and merge I'm curious if this will fix it

PR #1145, opened 2025-01-23
DEV 1261: Support app metatada id in deeplinking and public app endpoint

Takaros999:
## PR Type

- [X] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

This PR:
- Add support `app_metadata_id` query param in `/public/app/{app_id}` endpoint.
- Update QR code action to include `app_metadata_id`  used for deeplinking
- Add a warning message for draft version qr code/urls

MiniAppQrCode-Approved
MiniAppQrCode-Draft

## Checklist

- [x] I have self-reviewed this PR.
- [ ] I have left comments in the code for clarity.
- [x] I have added necessary unit tests.
- [ ] I have updated the documentation as needed.

## Depedencies
Wait for mobile work to suport these changes, otherwise the warning message above the QR code will confuse developers.

PR #1164, opened 2025-02-03
DEV 1377: Override top apps with draft data

Takaros999:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

This PR adds `override_top_apps_by_draft_ids` query param to `/public/apps` endpoint.

This is part of enabling "staging" mode for mini-apps, if a top app matches one of the given draft ids it will be returned with the draft metadata instead.

**Note:** A `draft_id` represents an `app_metadata_id` for context PTAL: #1145

## Checklist

- [X] I have self-reviewed this PR.
- [X] I have added necessary unit tests.

### DIFF
--- a/web/api/v2/public/apps/graphql/get-draft-metadata.generated.ts
+++ b/web/api/v2/public/apps/graphql/get-draft-metadata.generated.ts
@@ -1,0 +1,154 @@
+/* eslint-disable import/no-relative-parent-imports -- auto generated file */
+import * as Types from "@/graphql/graphql";
+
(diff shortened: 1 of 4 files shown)


andy-t-wang:
We should be able to see both the staging and prod version. It would not be a good developer experience to only see the staging one

andy-t-wang:
The product flow here should be if you have a staging app both staging and prod should show up with staging showing the warning icon

Takaros999:
closing this until we figure out what we want to do

PR #1182, opened 2025-02-12
Add webhook configuration to incognito actions config.

decentralgabe:
## PR Type

- [x] Regular Task

## Description

* Add support for configuring a webhook, associated with an action.
* `webhook_uri` and `webhook_pem` added to the existing action table.
* Updated the preflight gql query to include both fields.
* The webhook must have a corresponding RSA public key PEM.
* The `webhook_uri` and `webhook_pem` are biconditional dependencies, meaning they either both exist or both don't exist. As such, I've added validations to the create and update actions to ensure that this remains true, with appropriate error messages.
* I've manually tested this in a local dev environment (some screenshots attached).

Screenshot 2025-02-13 at 4 15 54 PM
Screenshot 2025-02-13 at 4 44 35 PM
Screenshot 2025-02-13 at 5 03 44 PM
Screenshot 2025-02-13 at 5 06 27 PM
Screenshot 2025-02-13 at 5 06 35 PM

### DIFF
--- a/hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
+++ b/hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
@@ -1,2 +1,6 @@
-alter table "public"."action" drop column "webhook_uri";
-alter table "public"."action" drop column "webhook_pem";
+ALTER TABLE public.action
+DROP COLUMN webhook_uri,
+  DROP COLUMN webhook_pem,
+  DROP COLUMN flow;
+
+DROP TYPE action_flow_enum;
--- a/hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
+++ b/hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
@@ -1,4 +1,2 @@
--- Could not auto-generate a down migration.
--- Please write an appropriate down migration for the SQL below:
--- alter table "public"."action" add column "webhook_uri" text
---  null;
+alter table "public"."action" drop column "webhook_uri";
+alter table "public"."action" drop column "webhook_pem";
--- a/hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
+++ b/hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
@@ -1,6 +1,6 @@
 ALTER TABLE public.action
 DROP COLUMN webhook_uri,
   DROP COLUMN webhook_pem,
-  DROP COLUMN flow;
+  DROP COLUMN app_flow_on_complete;
 
-DROP TYPE action_flow_enum;
+DROP TYPE app_flow_on_complete;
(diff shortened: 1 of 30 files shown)


decentralgabe:
main question - where should we add text to communicate the purpose of the webhook and PEM?
and what should it say? @andy-t-wang

andy-t-wang:
I think we should actually add the flow thing here and only if you specify the flow as partner or "verify" we ask you to put the webhook. Most people won't need the webhook and this will be confusing

hasura/migrations/default/1739394725314_alter_table_public_action_add_column_webhook_uri_pem/down.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."action" drop column "webhook_uri";
+alter table "public"."action" drop column "webhook_pem";
```
andy-t-wang, line 2: I think we should add the flow field here too so it's customizable on the action level

web/pages/api/v1/precheck/[app_id].ts
```diff
@@ -94,6 +94,8 @@ const appPrecheckQuery = gql`
         status
         privacy_policy_uri
         terms_uri
+        webhook_uri
+        webhook_pem
```
Comment (andy-t-wang) [inline web/pages/api/v1/precheck/[app_id].ts:98]: Note for self. Before merging we will test that this doesn't break anything in mobile clients if we add stuff to precheck

andy-t-wang (commented):
lgtm lets just add flow and the conditional logic

decentralgabe:
Updated to include 'flow' selector.

If flow = verify, the webhook settings are hidden.
If flow = partner, the webhook settings are visible.

I made it so that once you create an action the flow setting is fixed.
There is an edge case where you can set flow to partner and then remove the webhook settings.

Happy to adjust this if this doesn't seem right.

Screenshot 2025-02-13 at 9 46 06 PM
Screenshot 2025-02-13 at 9 51 53 PM
Screenshot 2025-02-13 at 9 44 55 PM

sujoyg:
@andy-t-wang @decentralgabe The flow enum is intended to specify which flow the partner wants the app to enter after the idkit flow itself is complete. For example, a partner X may want the world app to launch the app store, another may want us to launch the verification flow (i.e. nfc+orb), someone else may want us to open the wallet contents and so on.

"Partner" is not a valid flow value since there is no such user flow. Besides, all of IDKit is designed for partners, we don't need an enum for that.

andy-t-wang:
Thanks for that clarification

decentralgabe:
Screenshot 2025-02-14 at 4 36 03 PM
Screenshot 2025-02-14 at 4 36 06 PM
Screenshot 2025-02-14 at 4 37 01 PM

renamed to 'app flow on complete' and put it behind a toggle. tested and good to go!

web/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/page/CreateActionModal/index.tsx
```diff
@@ -80,9 +113,12 @@ export const CreateActionModal = (props: CreateActionModalProps) => {
     mode: "onChange",
     defaultValues: {
       max_verifications: 1,
+      app_flow_on_complete: "VERIFY",
```
andy-t-wang, line 116:
```suggestion
      app_flow_on_complete: "NONE",
```

andy-t-wang (approved):

PR #1197, opened 2025-02-19
fix: throw on verified draft

michalstruck:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

throw when draft_id equals the current verified row, to easily filter it out in app backend

## Checklist

### DIFF
--- a/web/api/v2/public/app/[app_id]/index.ts
+++ b/web/api/v2/public/app/[app_id]/index.ts
@@ -84,6 +84,13 @@ export async function GET(
     if (!draft_metadata) {
       return NextResponse.json({ error: "Draft not found" }, { status: 404 });
     }
+
+    if (app_metadata[0].id === draft_id) {
+      return NextResponse.json(
+        { error: "Draft already verified" },
+        { status: 400 },
+      );
+    }
     parsedAppMetadata = draft_metadata;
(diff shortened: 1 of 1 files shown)


andy-t-wang (approved):

web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -84,6 +84,13 @@ export async function GET(
     if (!draft_metadata) {
       return NextResponse.json({ error: "Draft not found" }, { status: 404 });
     }
+
```
Comment (Takaros999) [inline web/api/v2/public/app/[app_id]/index.ts:87]: I think we're missing the following scenario here:
- App metadata array has 2 entries but the second element is the verified one, I'm not sure the condition checks what you want

I would suggest modifying the filtering 3 lines of code above like this:
```
 const draft_metadata = app_metadata.find((meta) => meta.id === draft_id 
    !meta.is_reviewer_world_app_approved &&
        meta.verification_status !== "verified",
 );
    if (!draft_metadata) {
      return NextResponse.json({ error: "Draft not found" }, { status: 404 });
    }
 ```

Comment (michalstruck) [reply web/api/v2/public/app/[app_id]/index.ts:92]: yeah good catch, shouldn't rely on order lol

PR #1206, opened 2025-02-24
Server side RSA PEM Validation

decentralgabe:
## PR Type

- [x] Regular Task

## Description

Allows you to paste multi-line PEMs and have it still work, since we account for this (normalization) and validate the PEM is valid before submission on the server side.

In the future we can add a larger text box, but this works simply.

## Checklist

### DIFF
--- a/web/graphql/graphql.schema.json
+++ b/web/graphql/graphql.schema.json
@@ -4419,12 +4419,6 @@
             "deprecationReason": null
           },
           {
-            "name": "action_external_nullifier_key",
-            "description": "unique or primary key constraint on columns \"external_nullifier\"",
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
-          {
             "name": "action_pkey",
             "description": "unique or primary key constraint on columns \"id\"",
             "isDeprecated": false,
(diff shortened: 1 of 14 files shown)


web/graphql/graphql.schema.json
```diff
@@ -4418,12 +4418,6 @@
             "isDeprecated": false,
             "deprecationReason": null
           },
-          {
-            "name": "action_external_nullifier_key",
-            "description": "unique or primary key constraint on columns \"external_nullifier\"",
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
```
decentralgabe, line 4426: re-ran generate schemas/types and this disappeared

web/pages/api/v1/precheck/[app_id].ts
```diff
@@ -36,6 +36,8 @@ interface _Action
     | "external_nullifier"
     | "status"
     | "__typename"
+    | "webhook_uri"
```
Comment (decentralgabe) [inline web/pages/api/v1/precheck/[app_id].ts:39]: @andy-t-wang had to expose this

andy-t-wang (approved):

PR #1253, opened 2025-03-31
feat: Create transaction debug URL proxy

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Introduces a proxy to the transaction debug URL endpoint in [app-backend-main](https://github.com/worldcoin/app-backend-main/blob/ce06ea47f0ceda5c84bf7cd0fbd6a51963d1f5fc/src/transactions/transactions.internal.controller.ts#L76).

## Checklist

### DIFF
--- a/web/api/v2/minikit/transaction/debug/index.ts
+++ b/web/api/v2/minikit/transaction/debug/index.ts
@@ -1,0 +1,96 @@
+import { errorResponse } from "@/api/helpers/errors";
+import { validateRequestSchema } from "@/api/helpers/validate-request-schema";
+import { logger } from "@/lib/logger";
+import { createSignedFetcher } from "aws-sigv4-fetch";
+import { NextRequest, NextResponse } from "next/server";
+import * as yup from "yup";
+
+function corsHandler(response: NextResponse) {
+  response.headers.set("Access-Control-Allow-Origin", "*");
+  response.headers.set("Access-Control-Allow-Methods", "GET, OPTIONS");
+  return response;
+}
+
+const appIdRegex = /^app_[a-f0-9]{32}$/;
+
+const schema = yup.object({
+  app_id: yup
+    .string()
+    .matches(appIdRegex, "app_id must be in format app_{32 hex chars}")
+    .required(),
+});
+
+export const GET = async (req: NextRequest) => {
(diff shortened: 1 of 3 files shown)


web/api/v2/minikit/transaction/debug/index.ts
```diff
@@ -0,0 +1,96 @@
...
+import { NextRequest, NextResponse } from "next/server";
+import * as yup from "yup";
+
+function corsHandler(response: NextResponse) {
+  response.headers.set("Access-Control-Allow-Origin", "*");
+  response.headers.set("Access-Control-Allow-Methods", "GET, OPTIONS");
+  return response;
+}
+
+const appIdRegex = /^app_[a-f0-9]{32}$/;
```
michalstruck, line 14: we already have it in transaction/id, can we put it in one place and share
maybe in `web/lib/schema/index.ts`

michalstruck (approved):

web/api/v2/minikit/transaction/debug/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  response.headers.set("Access-Control-Allow-Methods", "GET, OPTIONS");
+  return response;
+}
+
+const schema = yup.object({
+  app_id: appIdSchema,
+});
+
+export const GET = async (req: NextRequest) => {
+  const { searchParams } = new URL(req.url);
```
andy-t-wang, line 20: We need to enforce API key here otherwise anyone can fetch your tenderly urls

andy-t-wang (changes requested):
See some other endpoints in the repo, we request api key for send_notification for example you can reuse the logic there

web/api/v2/minikit/transaction/debug/index.ts
Gr1dlock: Sure, makes sense, will add it
I used transaction/[transaction_id] as a boilerplate, meaning it's unprotected as well, shall we cover it too?

andy-t-wang (approved):

andy-t-wang:
added a follow on task for you to do docs and add tests

PR #1255, opened 2025-03-31
feat: [DEV-1375] App Store assets localization

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

* Moves the image form to the localized form section
* Adds the image URL fields to the localisation structure
* Allows the user to add and delete images for each localisation
* Partitions images by locale
* Adds the locale argument to the hasura actions
* Allows the localisation fields to be empty
* Creates an empty localisation when the user selects a new language
* Makes the English language non-removable
* Modifies the endpoints to use the localised images

## Checklist


Gr1dlock:
@andy-t-wang cos of bugs we discovered when testing the transition from the unverified state to the verified state

michalstruck (approved):

PR #1387, opened 2025-05-13
change deeplinking and delisting logic

michalstruck:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
as per legal requirements, apps will be deeplinkable despite country restrictions

this also causes apps to be hidden from home screen when the new flag is set

ref DEV-1729

## Checklist

### DIFF
--- a/web/api/v2/public/app/[app_id]/index.ts
+++ b/web/api/v2/public/app/[app_id]/index.ts
@@ -128,9 +128,16 @@ export async function GET(
     ? null
     : headers.get("CloudFront-Viewer-Country");
   const override_country = searchParams.get("override_country") || country;
+  const shouldUninstallOnDelist = parsedAppMetadata.should_uninstall_on_delist;
+  const isDelisted = !parsedAppMetadata.is_reviewer_app_store_approved;
 
-  // do not restrict for drafts, so developers can work on the app
+  // only restrict based on country if app is delisted and should be uninstalled on delist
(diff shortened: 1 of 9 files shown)


web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -128,9 +128,16 @@ export async function GET(
     ? null
     : headers.get("CloudFront-Viewer-Country");
   const override_country = searchParams.get("override_country") || country;
```
Comment (michalstruck) [inline web/api/v2/public/app/[app_id]/index.ts:130]: this, migration and tests are the only interesting files here, rest is autogenerated

web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -128,9 +128,16 @@ export async function GET(
...
+  const shouldUninstallOnDelist = parsedAppMetadata.should_uninstall_on_delist;
+  const isDelisted = !parsedAppMetadata.is_reviewer_app_store_approved;
 
-  // do not restrict for drafts, so developers can work on the app
+  // only restrict based on country if app is delisted and should be uninstalled on delist
+  // this means that the only two cases where an app is not deeplinkable are:
+  // - if it's banned
+  // - if the uninstall flag is set and app is delisted
   if (
+    isDelisted &&
```
Comment (andy-t-wang) [inline web/api/v2/public/app/[app_id]/index.ts:139]: There are two ways to delist right now
If it's not available in your country it shouldn't be delisted either

PR #1413, opened 2025-05-22
stricter validation of notifications

michalstruck:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
add stricter rules, requires notification from app to deeplink into the app, enforces wallet addresses to be 42 chars long

## Checklist

### DIFF
--- a/web/api/v2/minikit/send-notification/index.ts
+++ b/web/api/v2/minikit/send-notification/index.ts
@@ -18,11 +18,11 @@ import {
 } from "./graphql/create-notification-log.generated";
 import { getSdk as fetchMetadataSdk } from "./graphql/fetch-metadata.generated";
 
-const sendNotificationBodySchema = yup.object({
+export const sendNotificationBodySchema = yup.object({
   app_id: yup.string().strict().required(),
   wallet_addresses: yup
     .array()
-    .of(yup.string())
+    .of(yup.string().length(42))
     .min(1)
     .max(1000)
     .required("wallet_addresses is required"),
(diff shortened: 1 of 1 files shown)


andy-t-wang (approved):
I think we can be stricter here. 

Technically I could link to another app here and just add a garbage path query param with my app ID

ie my app is app_1


worldapp://mini-app?app_id=app_malicious&path=?app_1

andy-t-wang:
Also test failing

web/api/v2/minikit/send-notification/index.ts
```diff
@@ -18,11 +18,11 @@ import {
 } from "./graphql/create-notification-log.generated";
 import { getSdk as fetchMetadataSdk } from "./graphql/fetch-metadata.generated";
 
-const sendNotificationBodySchema = yup.object({
```
Gonzamany, line 21: db24a22

Gonzamany (approved):

web/api/v2/minikit/send-notification/index.ts
Gonzamany: db24a22

web/api/v2/minikit/send-notification/index.ts
```diff
@@ -18,11 +18,11 @@ import {
...
+export const sendNotificationBodySchema = yup.object({
```
Gonzamany, line 21: db24q22

Gonzamany (changes requested):

michalstruck:
> I think we can be stricter here. Check starts with mini `worldapp://mini-app?app_id={this.parent}`
> 
> Otherwise I could link to another app here and just add a garbage path query param with my app ID ie my app is app_1
> 
> worldapp://mini-app?app_id=app_malicious&path=?app_1

true good point

PR #1417, opened 2025-05-23
add new flag & logic

michalstruck:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
add flag so that we don't need more deps in our other services

DEV-1732 related

## Checklist

### DIFF
--- a/web/api/helpers/app-store.ts
+++ b/web/api/helpers/app-store.ts
@@ -164,7 +164,25 @@ export const formatAppMetadata = async (
     permit2_tokens: permit2Tokens,
     contracts: contracts,
     supported_countries: supportedCountries,
+    should_show_notification_badge: getShouldShowNotificationBadge(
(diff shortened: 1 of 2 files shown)


web/api/helpers/app-store.ts
```diff
@@ -164,9 +164,27 @@ export const formatAppMetadata = async (
     permit2_tokens: permit2Tokens,
     contracts: contracts,
     supported_countries: supportedCountries,
+    should_show_notification_badge: getShouldShowNotificationBadge(
+      singleAppStats?.open_rate_last_14_days,
+    ),
   };
 };
 
+const getShouldShowNotificationBadge = (
```
Gr1dlock, line 173: For bool functions you can keep it at `shouldBlaBlaBla`

PR #1421, opened 2025-05-23
feat: Replace app_stats function with table and incremental trigger

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Ticket - [DEV-1258](https://linear.app/worldcoin/issue/DEV-1258/dev-portal-optimize-app-stats-query)
Removes the `app_stats` function due to performance issues.
Introduces a new table `app_stats`, which is updated cumulatively using a Hasura event trigger on the `nullifier` table.
Every time the `uses` filed gets updated, the trigger updates the `app_stats` table with the new data:
- If a row with the same `app_id` and `date` already exists, the `verifications` and `unique_users` columns are updated
  - the `verifications` column is incremented by 1
  - the `unique_users` column is incremented by 1 if the `nullifier_hash` is unique
- If no row with the same `app_id` and `date` exists, a new row is inserted with the `verifications` and `unique_users` columns set to 1
- The `nullifier_hash` column is used to keep track of unique users

## Checklist


hasura/metadata/databases/default/tables/public_app_stats.yaml
Takaros999, line 1: possible issue: I think we're missing permissions here

I think we need `user` and maybe `service` as well, please verify and test locally to make sure we don't miss here

andy-t-wang (changes requested):

hasura/migrations/default/1748005396949_create_and_fill_table_app_stats/up.sql
```diff
@@ -0,0 +1,86 @@
...
+BEGIN
+  SELECT app_id INTO _app_id
+  FROM action
+  WHERE id = NEW.action_id;
+
+  IF _app_id IS NULL THEN
+    RETURN NEW;
+  END IF;
+
+  SELECT NEW.nullifier_hash = ANY(nullifier_hashes)
```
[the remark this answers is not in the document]
andy-t-wang: If we fail to update the stats table, it's ok not a huge deal. It's not business logic critical

hasura/migrations/default/1748005396949_create_and_fill_table_app_stats/up.sql
```diff
@@ -0,0 +1,86 @@
...
+    ON DELETE CASCADE
+    ON UPDATE CASCADE
+);
+
+-- Index for app_id lookups
+CREATE INDEX IF NOT EXISTS idx_app_stats_app_id ON app_stats (app_id);
+
+-- Backfill data
+INSERT INTO app_stats (app_id, date, verifications, nullifier_hashes, unique_users)
+SELECT
```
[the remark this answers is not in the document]
Gr1dlock: I will do it in a separate script in batches then

hasura/migrations/default/1748005396949_create_and_fill_table_app_stats/up.sql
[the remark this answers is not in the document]
Gr1dlock: Agreed, moved it to the Hasura event trigger

hasura/migrations/default/1748005396949_create_and_fill_table_app_stats/up.sql
```diff
@@ -0,0 +1,72 @@
+-- Drop the old app_stats function and return type table if they exist
+DROP FUNCTION IF EXISTS public.app_stats(varchar, timestamptz, varchar);
```
andy-t-wang, line 2: To be honest this could have been done in parts and made review a lot easier.
1. We create new logic
2. We drop old logic

web/graphql-codegen.types.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     scalars: {
       timestamp: "string",
       timestamptz: "string",
+      date: "string",
```
andy-t-wang, line 18: ?

web/app/api/%5Fincrement-app-stats/route.ts
andy-t-wang, line 1: This is named this way because of next js naming conventions?

andy-t-wang (approved):

web/app/api/%5Fincrement-app-stats/route.ts
Gr1dlock: Yep, App Router would consider it a private folder otherwise

web/graphql-codegen.types.js
Gr1dlock: Needed for GraphQL type generation

hasura/migrations/default/1748005396949_create_and_fill_table_app_stats/up.sql
Gr1dlock: Fair point, but then I wouldn't have been able to use the same name and would have had to create 3 PRs (create, reset, rename) instead of 1

PR #1423, opened 2025-05-26
feat: Add character constraint on team name

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Ticket - [DEV-1737](https://linear.app/worldcoin/issue/DEV-1737/h1-devportal-team-name-field-needs-a-maximum)

Introduces a character constraint on the team name field in the onboarding form and in the database.
Max length is limited to 128 characters and the team name cannot contain `<` or `>` characters.

## Checklist


web/scenes/Onboarding/CreateTeam/page/Form/index.tsx
```diff
@@ -23,7 +23,11 @@ export const Form = (props: { hasUser: boolean }) => {
     () =>
       yup.object({
         hasUser: yup.boolean().default(props.hasUser),
-        teamName: yup.string().required("Please enter a team name"),
+        teamName: yup
+          .string()
+          .required("Please enter a team name")
+          .max(128, "Team name must be 128 characters or less")
+          .matches(/^[^<>]*$/, "Team name cannot contain < or > characters"),
```
michalstruck, line 30: i think it would be best to use something similar to `appNameSchema` in web/li/schema/index.ts

hasura/migrations/default/1748272793146_limit_team_name_characters/up.sql
```diff
@@ -0,0 +1,8 @@
+-- 1. Limit length to 128 characters
+ALTER TABLE public.team
+ALTER COLUMN name TYPE VARCHAR(128);
+
+-- 2. Disallow '<' and '>' characters using a CHECK constraint
+ALTER TABLE public.team
+ADD CONSTRAINT team_name_no_angle_brackets
+CHECK (name !~ '[<>]');
```
michalstruck, line 8: first of all this will most likely fail since there are known team names with this
i think a better approach would be to set these team names to some string like 'change this name' or sth
and if there's server side validation for the form, there's no need to adjust the db schema

michalstruck (approved):
as in comments

michalstruck (changes requested):
haha approved and i meant to request changes

web/lib/permissions/index.ts
```diff
@@ -153,3 +153,24 @@ export const getIsUserAllowedToDeleteLocalisation = async (
...
+  if (!session) {
+    return false;
+  }
+
+  const userId = session.user.hasura.id;
+  const response = await getTeamUpdatePermissionsSdk(
+    await getAPIServiceGraphqlClient(),
+  ).GetIsUserPermittedToModifyTeam({ teamId, userId });
+
+  if (response.team.length) {
```
andy-t-wang, line 172: Nit: Good to be explicit and check the id actually matches

andy-t-wang (approved):

web/lib/schema/index.ts
```diff
@@ -218,3 +218,12 @@ export const notificationMessageSchema = yup
       );
     },
   );
+
+export const teamNameSchema = yup
+  .string()
+  .required("Please enter a team name")
+  .max(128, "Team name must be 128 characters or less")
+  .matches(/^[^<>]*$/, {
```
michalstruck, line 226: i suggest to use `allowedTitleCharactersRegex` or `allowedCommonCharactersRegex` but up to you

Gr1dlock: We restrict only those because they break the frontend (check the ticket)

michalstruck: regexes i recommended were created after previous security reports
they also restrict these characters afaik

Gr1dlock: Alright, used `allowedTitleCharactersRegex`

web/lib/permissions/index.ts
Gr1dlock: Sure, added an extra check for that

michalstruck (approved):

PR #1443, opened 2025-06-03
remove versioning for contacts miniapp

michalstruck:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
remove version filtering

improve typings and add tests which would've prevented yesterday's outage

## Checklist


web/lib/constants.ts
```diff
@@ -56,9 +56,12 @@ export const NATIVE_MAPPED_APP_ID = {
   invites: "invites",
   network: "network",
   contacts: "contacts",
-};
+} as const;
 
-export const NativeAppToAppIdMapping: Record<string, Record<string, string>> = {
+export const NativeAppToAppIdMapping: Record<
+  "dev" | "staging" | "production",
```
Gr1dlock, line 62: How about creating an enum for that instead of listing all the possible values every time?

Gr1dlock (approved):
Godspeed (please block the prod transition 😅 )

PR #1476, opened 2025-06-23
feat: Add user grant cycle endpoint

Takaros999:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

This PR adds a new minikit endpoint to fetch information about the user grant cycle. 

This PR also adds a util to verify an API key. I will follow up with a PR that uses this util in older endpoints too.

Completes DEV-1920

## Checklist


web/api/v2/minikit/user-grant-cycle/index.ts
```diff
@@ -0,0 +1,107 @@
...
+import { createSignedFetcher } from "aws-sigv4-fetch";
+import { NextRequest, NextResponse } from "next/server";
+import * as yup from "yup";
+
+const userGrantCycleQuerySchema = yup.object({
+  wallet_address: yup
+    .string()
+    .length(42)
+    .required("wallet_address is required"),
+  required_app_id: yup
```
andy-t-wang: for consistency I believe other endpoints just use. The endpoint will error if it's required
```suggestion
  app_id: yup
```

web/tests/api/v2/minikit-user-grant-cycle.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    },
+  },
+};
+
+const validAppId = validApiKeyResponse.api_key_by_pk.team.apps[0].id;
+const validWalletAddress = "0x1234567890123456789012345678901234567890";
+const validApiKey = `api_${apiKeyValue}`;
+
+const mockSuccessResponse = {
+  nextGrantClaimDate: "2024-12-31T23:59:59.999Z",
```
andy-t-wang: This is in UTC?

web/tests/api/v2/minikit-user-grant-cycle.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    const res = await GET(mockReq);
+    expect(res.status).toBe(200);
+    const body = await res.json();
+
+    expect(body).toEqual({
+      success: true,
+      status: 200,
+      result: {
+        nextGrantClaimDate: mockSuccessResponse.nextGrantClaimDate,
```
andy-t-wang:
```suggestion
        nextGrantClaimUTCDate: mockSuccessResponse.nextGrantClaimDate,
```

if this is in UTC I think it would be good to specify. Meant to put this in the endpoint just noticed in tests

web/tests/api/v2/minikit-user-grant-cycle.test.ts
andy-t-wang, line 1: Unfortunately tests for this endpoint aren't very useful with dima we should think about better e2e testing as we are simply mocking most of the logic

PR #1483, opened 2025-06-27
Set default root expiration to 1 week

Takaros999:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Part of DEV-1952

## Checklist

### DIFF
--- a/web/tests/api/v2/verify.test.ts
+++ b/web/tests/api/v2/verify.test.ts
@@ -715,18 +715,12 @@ describe("/api/v2/verify [error cases]", () => {
     const stagingAppId = "app_staging_558238f8f6380338449a552aeffccf29";
     const ctx = { params: { app_id: stagingAppId } };
 
-    const mockReq = new NextRequest(
-      new URL(`https://foo.bar/api/v2/verify/${stagingAppId}`),
-      {
-        method: "POST",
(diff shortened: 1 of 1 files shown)


web/api/helpers/verify.ts
```diff
@@ -6,6 +6,9 @@ import { VerificationLevel } from "@worldcoin/idkit-core";
 import { AbiCoder, toBeHex } from "ethers";
 import * as yup from "yup";
 
+// Default maximum root age in seconds (7 days)
+const DEFAULT_MAX_ROOT_AGE = 60 * 60 * 24 * 7;
```
thereisnogabe, line 10: can we make this shorter, perhaps 1 day?

what are the reasons to keep it at a week?

Takaros999: in the `/verify` endpoint, we had this max age param as optional, a user could choose from 1 hour up to 1 week.

Made sense to me to default to the maximum we allow.

andy-t-wang: This is temporary no just for testing to make sure no one's app gets broken? We will make it 2 hours

andy-t-wang (approved):

PR #1484, opened 2025-06-28
Migrate partially to v2 verify sequencer endpoint

Takaros999:
# Summary
This PR migrates to V2 of the sequencer verify endpoint -find the new spec [here](https://github.com/worldcoin/signup-sequencer/blob/main/schemas/openapi-v2.yaml)- feature flagged for specific apps as this is a high risk change.

### DIFF
--- a/web/api/helpers/verify.v2.ts
+++ b/web/api/helpers/verify.v2.ts
@@ -1,0 +1,138 @@
+import { IInternalError } from "@/lib/types";
+import { sequencerMapping } from "@/lib/utils";
+import {
+  IInputParams,
+  IVerifyParams,
+  parseProofInputs,
+} from "./verify";
+
+const KNOWN_ERROR_CODES = [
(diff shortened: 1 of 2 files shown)


web/api/helpers/verify.v2.ts
andy-t-wang, line 1: I dont think this file naming is consistent with what we currently have

web/api/helpers/verify.v2.ts
```diff
@@ -0,0 +1,135 @@
...
+          "There was an internal issue processing the verification response.",
+        code: "internal_error",
+        statusCode: 500,
+      },
+    };
+  }
+};
+
+// Re-export utilities from verify.ts for backward compatibility
+export { canVerifyForAction, nullifierHashToBigIntStr } from "./verify";
```
andy-t-wang: Nit: I prefer this at the top

PR #1487, opened 2025-07-01
app store refactor - showcase and meta tag images upload

michalstruck:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
- fix image uploads in localisations
- allow concurrent uploads - previously you could only upload one image in the entire form, and during that all fields were disabled
- create an uploader component that handles loading states and dropzone
- make sure localisations have no duplicates

## Checklist


web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/MetaTagImageField.tsx
```diff
@@ -31,22 +37,79 @@ export const MetaTagImageField = (props: MetaTagImageFieldProps) => {
...
+    },
+    onError: (error) => {
+      console.error("autosave failed:", error);
+      toast.error("Failed to auto-save meta tag image");
+      onAutosaveError?.(error);
+    },
+  });
+
+  const performAutosave = useCallback(
+    async (possiblyActualUrl: string | null) => {
```
andy-t-wang: what is this variable name

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/MetaTagImageField.tsx
```diff
@@ -139,20 +209,19 @@ export const MetaTagImageField = (props: MetaTagImageFieldProps) => {
...
-          : imagesData?.unverified_images?.meta_tag_image_url;
-      if (!imageUrl) return null;
-      return getCDNImageUrl(appId, imageUrl, true, locale);
+      if (!value) return null;
+
+      return getCDNImageUrl(appId, value, true, locale);
     } else {
-      return imagesData?.unverified_images?.meta_tag_image_url;
+      return (
+        unverifiedImagesData?.unverified_images?.meta_tag_image_url || null
```
andy-t-wang: Is there a way we can reduce the number of ? nullish chaining. We often run into issues because we miss some. Wonder if we can type more strongly

michalstruck: agreed on stronger typing (but generally it's quite strong i'd say), don't see a point in reducing the number of these since it's a valid pattern/language feature

we should have as strict tsconfig as possible and maybe some more lints (forcing explicit function return types) - i'll put it in the backlog

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+      const fileTypeEnding = file.type.split("/")[1];
+
+      try {
+        // validate first, before showing any progress
+        await validateImageAspectRatio(file, width, height);
+
+        // only start showing progress after validation passes
+        setIsUploading(true);
+        onUploadStart?.();
```
andy-t-wang, line 102: Why does this function need a nullish chain? Why would it not be defined when being used

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
```diff
@@ -0,0 +1,315 @@
...
+        const extractedPath =
+          extractImagePathWithExtensionFromActualUrl(imageUrl);
+        const newUrls =
+          maxImages === 1 ? [extractedPath] : [...value, extractedPath];
+
+        await onAutosave(newUrls);
+        await onRefetchImages();
+        onChange(newUrls);
+
+        onUploadSuccess?.();
```
andy-t-wang, line 138: same

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
```diff
@@ -0,0 +1,315 @@
...
+      } catch (error) {
+        console.error("error uploading image:", error);
+
+        if (error instanceof ImageValidationError) {
```
andy-t-wang, line 142: Can just invert then? 
```suggestion
        if (!(error instanceof ImageValidationError)) {
```

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/hook/use-image.tsx
```diff
@@ -4,9 +4,11 @@ import { useGetUploadedImageLazyQuery } from "./graphql/client/get-uploaded-imag
 import { useUploadImageLazyQuery } from "./graphql/client/upload-image.generated";
 
 export class ImageValidationError extends Error {
+  public readonly toastId: string;
```
andy-t-wang, line 7: If only one modal is meant to be shown for all we should just make it static so there's no need to do prototype. Prototype is less explicit

```suggestion
  public static readonly toastId: string;
```

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/hooks/useFormData.ts
```diff
@@ -16,9 +16,6 @@ export const useFormData = (
 
     return {
       ...appMetadata,
-      // don't set these boolean fields from appMetadata so they start unchecked
-      is_android_only: undefined,
```
andy-t-wang, line 20: why delete? We just use  default?

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/LogoImageUpload/index.tsx
```diff
@@ -68,7 +68,7 @@ export const LogoImageUpload = (props: LogoImageUploadProps) => {
           autoClose: false,
         });
 
-        toast.dismiss("ImageValidationError");
+        toast.dismiss(ImageValidationError.prototype.toastId);
```
andy-t-wang, line 71: See comment below on making this static

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/hook/use-image.tsx
michalstruck: yep good catch this i much better

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/hooks/useFormData.ts
michalstruck: yes its using the default with the spread one line up.defaulting to undefined was an unintentional mistake

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
michalstruck: not needed, removed

PR #1488, opened 2025-07-01
feat: Create QA role with full rights

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-1985](https://linear.app/worldcoin/issue/DEV-1985/set-up-permissions-for-qa-team)
Creates a Hasura role for the QA team to allow CRUD operations on all tables.

## Checklist

### DIFF
--- a/hasura/metadata/databases/default/tables/public_action.yaml
+++ b/hasura/metadata/databases/default/tables/public_action.yaml
@@ -47,6 +47,30 @@ insert_permissions:
         - app_flow_on_complete
         - webhook_pem
         - webhook_uri
+  - role: qa
+    permission:
+      check: {}
+      columns:
+        - kiosk_enabled
+        - app_id
+        - client_secret
+        - creation_mode
+        - id
+        - status
+        - description
+        - privacy_policy_uri
+        - terms_uri
+        - created_at
+        - updated_at
+        - app_flow_on_complete
+        - external_nullifier
+        - name
+        - max_accounts_per_user
+        - max_verifications
+        - webhook_pem
+        - action
+        - webhook_uri
+    comment: ""
   - role: service
     permission:
       check: {}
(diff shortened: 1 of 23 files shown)


hasura/metadata/databases/default/tables/public_action.yaml
```diff
@@ -47,6 +47,30 @@ insert_permissions:
         - app_flow_on_complete
         - webhook_pem
         - webhook_uri
+  - role: qa
+    permission:
+      check: {}
+      columns:
+        - kiosk_enabled
+        - app_id
```
andy-t-wang, line 55: This looks like way too many permissions. There would be no test where the QA role needs more permissions than what a user or the service role even gets.

andy-t-wang: I prefer we start with what tests we want and iteratively add permissions. We had security incidents with loose permissioning

andy-t-wang (changes requested):
Please see my above comment you are just giving admin access to a key which is not good

hasura/metadata/databases/default/tables/public_action.yaml
Gr1dlock: I would argue that the QA team needs more permissions than typical users, cause they need to be able to set up and tear down data structures without going through the backend calls.
I agree though, that we need to list the test scenarios and based on them we'll be able to figure out the required tables/columns

Gr1dlock: @bdoof what's your take?

bdoof: Yes, we need quite broader permissions for the QA role as it's going to be used as a technical role for the setup and the teardown.

Are users permitted to add/delete data to all the common tables?

Gr1dlock: No, we restrict a lot of columns to be set only by backend

andy-t-wang: We should just add permissions as we need

Gr1dlock:
As aligned with @andy-t-wang and @bdoof, we will add permissions on the go to follow the principle of least privilege.
Closing for now.

PR #1491, opened 2025-07-03
feat: Add endpoint to query verifications by action

Takaros999:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

This PR adds a GET endpoint to query verifications by action id, it also includes optional filters like pagination and date ranges.

## Checklist

### DIFF
--- a/web/api/v2/verifications/index.ts
+++ b/web/api/v2/verifications/index.ts
@@ -1,0 +1,173 @@
+import { verifyApiKey } from "@/api/helpers/auth/verify-api-key";
+import { errorResponse } from "@/api/helpers/errors";
+import { getAPIServiceGraphqlClient } from "@/api/helpers/graphql";
+import { validateRequestSchema } from "@/api/helpers/validate-request-schema";
+import dayjs from "dayjs";
+import { NextRequest, NextResponse } from "next/server";
+import * as yup from "yup";
+import { getSdk as fetchVerificationsSdk } from "./graphql/fetch-verifications.generated";
+
+// Default and max pagination limits
+const DEFAULT_LIMIT = 100;
+const MAX_LIMIT = 1000;
+
+const querySchema = yup.object({
+  app_id: yup.string().strict().required("app_id is required"),
+  action_id: yup.string().strict().required("action_id is required"),
+  start_date: yup.date().optional(),
+  end_date: yup.date().optional(),
+  offset: yup
+    .number()
+    .integer()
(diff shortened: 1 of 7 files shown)


web/api/v2/verifications/index.ts
```diff
@@ -0,0 +1,186 @@
...
+import { validateRequestSchema } from "@/api/helpers/validate-request-schema";
+import dayjs from "dayjs";
+import { NextRequest, NextResponse } from "next/server";
+import * as yup from "yup";
+import { getSdk as fetchActionSdk } from "./graphql/fetch-action.generated";
+import { getSdk as fetchVerificationsSdk } from "./graphql/fetch-verifications.generated";
+
+// Default and max pagination limits
+const DEFAULT_LIMIT = 100;
+const MAX_LIMIT = 1000;
```
andy-t-wang, line 13: This is like 1000 nullifiers max. Think this doesn't solve the studios issue of needing to know how many people used an action.

web/api/v2/verifications/index.ts
```diff
@@ -0,0 +1,186 @@
...
+    }));
+
+    // For test compatibility, we'll approximate the total count
+    // In production, this should use nullifier_aggregate
+    const returnedCount = nullifier.length;
+    const totalCount =
+      returnedCount === limit ? offset + limit + 1 : offset + returnedCount;
+    const hasNextPage = returnedCount === limit;
+    const currentPage = Math.floor(offset / limit) + 1;
+    const totalPages = Math.ceil(totalCount / limit);
```
andy-t-wang, line 159: What is the point of this? This is only useful once you fetched all but each time you fetch it's gonna change since total count depends on the offset

andy-t-wang: The other question is if they need all the nullifier metadata or simply a count in a time period is enough. We can return nullifier_hash and uses but nothing else imo, we should minimize data

web/api/v2/verifications/index.ts
andy-t-wang: I think the shape should be more like 
Action_id, app_id, start_date, end_date. -> returns {n nullifiers}

web/tests/api/v2/minikit-user-grant-cycle.test.ts
michalstruck, line 1: accidentally committed?

PR #1501, opened 2025-07-09
soft delete apps

michalstruck:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Introduces a `deleted_at` property to the `app` table. 
- `user` can mark app as deleted once, after that the app doesn't show up in their team, and cannot be modified,
- "no modifications" applies to `app`, `app_metadata`, `localisations` associated with the deleted app
- `reviewer` and `user` roles are not permitted to select or update app data once it's marked as deleted
- deleted apps are unavailable in /apps, but are returned from /app
- metadata in endpoints now has a nullable `deleted_at` property

## Checklist


andy-t-wang:
First question from a product standpoint how would we expect delete to work? A user can still access the miniapp? If no then we shouldn't return it from /app.

hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -123,7 +128,9 @@ update_permissions:
       columns:
         - is_banned
       filter: {}
-      check: null
+      check:
```
andy-t-wang, line 131: We might need an index on this. The way hasura works is that it only grabs authorized rows then runs the query. So this is adding a where operation to every single SQL operation for app and app_metadata table.

andy-t-wang: Worth doing some resarch

andy-t-wang (approved):

hasura/metadata/databases/default/tables/public_app.yaml
michalstruck: no because all the tables affected are very small so an index wouldnt be used anyway

PR #1503, opened 2025-07-10
flag to return a deleted app instead of 404

michalstruck:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
Deleted apps should not be returned to home-screen

- introduces new `show_deleted` flag
- changes endpoint behavior to return a 404 when metadata is deleted, unless this new flag is present and true
- adds tests for this behavior

## Checklist

### DIFF
--- a/web/api/v2/public/app/[app_id]/index.ts
+++ b/web/api/v2/public/app/[app_id]/index.ts
@@ -117,8 +117,12 @@ export async function GET(
     }
   }
 
-  const isMetadataVerified =
-    parsedAppMetadata.verification_status === "verified";
+  const shouldShowDeleted = searchParams.get("show_deleted") === "true";
+  const isAppDeleted = parsedAppMetadata.app.deleted_at != null;
+
+  if (isAppDeleted && !shouldShowDeleted) {
+    return NextResponse.json({ error: "App not found" }, { status: 404 });
+  }
 
   // skip checking cf country, when coming from app-backend
   const skipCloudfrontCheck = Boolean(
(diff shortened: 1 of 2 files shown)


web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -117,8 +117,12 @@ export async function GET(
     }
   }
 
-  const isMetadataVerified =
-    parsedAppMetadata.verification_status === "verified";
+  const shouldShowDeleted = searchParams.get("show_deleted") === "true";
+  const isAppDeleted = parsedAppMetadata.app.deleted_at != null;
```
Comment (michalstruck) [inline web/api/v2/public/app/[app_id]/index.ts:121]: `!=` because it's typed as `null | undefined` and this will match both

Comment (aurel-fr) [reply web/api/v2/public/app/[app_id]/index.ts:121]: I love those loose equality comparisons. Eslint won't allow it in app-backend but they're very convenient.

web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -130,6 +134,8 @@ export async function GET(
   const override_country = searchParams.get("override_country") || country;
   const shouldUninstallOnDelist = parsedAppMetadata.should_uninstall_on_delist;
   const isDelisted = !parsedAppMetadata.is_reviewer_app_store_approved;
+  const isMetadataVerified =
+    parsedAppMetadata.verification_status === "verified";
```
Comment (aurel-fr) [inline web/api/v2/public/app/[app_id]/index.ts:138]: This looks unrelated to this PR?

Comment (michalstruck) [reply web/api/v2/public/app/[app_id]/index.ts:138]: yes i just moved it down

andy-t-wang (approved):

PR #1517, opened 2025-07-17
fix app store rankings

andy-t-wang:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Checklist

### DIFF
--- a/web/api/helpers/app-store.ts
+++ b/web/api/helpers/app-store.ts
@@ -5,14 +5,21 @@ import {
   whitelistedAppsPermit2,
 } from "@/lib/constants";
 import { generateExternalNullifier } from "@/lib/hashing";
+import { logger } from "@/lib/logger";
 import {
   AppStatsItem,
   AppStatsReturnType,
   AppStoreFormattedFields,
   AppStoreMetadataDescription,
   AppStoreMetadataFields,
+  MetricsServiceAppData,
 } from "@/lib/types";
-import { getCDNImageUrl, getLogoImgCDNUrl, tryParseJSON } from "@/lib/utils";
+import {
+  fetchWithRetry,
+  getCDNImageUrl,
+  getLogoImgCDNUrl,
+  tryParseJSON,
+} from "@/lib/utils";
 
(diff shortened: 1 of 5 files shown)


aurel-fr (approved):

web/api/helpers/app-store.ts
```diff
@@ -278,3 +288,92 @@ export const rankApps = (
...
+  let metricsData: AppStatsReturnType = [];
+
+  if (country) {
+    // If country is specified return the value for that Country code
+    // for unique_users_last_7_days, new_users_last_7_days, and total_users_last_7_days
+    metricsData = responseData.map((app) => {
+      return {
+        ...app,
+        unique_users_last_7_days: app.unique_users_last_7_days?.find(
+          (user) => user.country === country,
```
aurel-fr, line 346: nit: I'd make that comparison case insensitive

web/api/helpers/app-store.ts
```diff
@@ -278,3 +288,92 @@ export const rankApps = (
...
+): Promise<AppStatsReturnType> => {
+  // Configure caching based on expirationTime
+  const fetchOptions: RequestInit = {};
+
+  if (expirationTime <= 0) {
+    // No cache - force fresh data
+    fetchOptions.cache = "no-store";
+    fetchOptions.headers = {
+      "Cache-Control": "no-cache, no-store, must-revalidate",
+      Pragma: "no-cache",
```
aurel-fr, line 310: this header is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma)

web/api/helpers/app-store.ts
```diff
@@ -278,3 +288,92 @@ export const rankApps = (
...
+      Expires: "0",
+    };
+  } else {
+    // Cache for specified time
+    fetchOptions.headers = {
+      "Cache-Control": `max-age=${expirationTime}`,
```
aurel-fr, line 316: nit: consider Math.floor() to enforce the value will be an int

PR #1518, opened 2025-07-17
fix comments

andy-t-wang:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
Missed comments from #1517 

## Checklist

### DIFF
--- a/web/api/helpers/app-store.ts
+++ b/web/api/helpers/app-store.ts
@@ -342,13 +342,13 @@ export const fetchMetrics = async (
       return {
         ...app,
         unique_users_last_7_days: app.unique_users_last_7_days?.find(
-          (user) => user.country.toUpperCase() === country.toUpperCase(),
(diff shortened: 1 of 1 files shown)


web/api/helpers/app-store.ts
```diff
@@ -343,13 +342,13 @@ export const fetchMetrics = async (
       return {
         ...app,
         unique_users_last_7_days: app.unique_users_last_7_days?.find(
-          (user) => user.country === country,
+          (user) => user.country.toUpperCase() === country.toUpperCase(),
```
aurel-fr: Lacking context, if there is any instance where `user.country` is undefined or null this might be needed.
```suggestion
          (user) => user.country?.toUpperCase() === country.toUpperCase(),
```

andy-t-wang: hmm I guess yeah can do that. Comes from metrics service so better to handle on our side

PR #1521, opened 2025-07-18
Pending memory cache for fetching metrics

andy-t-wang:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
In memory cache to prevent duplicate concurrent requests for a given country from the metrics service.

Also given we care about performance here I opted to use explicit for loops loops directly as given this is a simple sum over large data the function calling with reduce actually takes more overhead than the operation (javascript!). ([src](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/))

## Checklist


web/api/helpers/fetch-metrics.ts
```diff
@@ -0,0 +1,141 @@
...
+
+          if (app.new_users_last_7_days) {
+            for (const user of app.new_users_last_7_days) {
+              newUsersSum += Number(user.value) || 0;
+            }
+          }
+
+          if (app.total_users_last_7_days) {
+            for (const user of app.total_users_last_7_days) {
+              totalUsersSum += Number(user) || 0;
```
aurel-fr, line 119: Is it not `user.value` here?

aurel-fr (approved):
Nice, I wasn't aware of the performance impact of `reduce`. 

If we reuse this request coalescing pattern elsewhere, we can consider abstracting it into a `withRequestCoalescing` util in a future PR.

michalstruck (approved):

PR #1527, opened 2025-07-18
Refactor v2 of Fetch Metrics

andy-t-wang:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Fetch metrics currently returns a huge list of data. 

This implementation changes the logic to such
1. We do a single call to the metrics service that calculates the data for each country and global. The data is then cached in memory for 10 minutes.
2. Any requests that come in while this call is in progress will await the call. 
3. If there is already data in the memory cache we will return that immediately. 

Upon invalidating cache we also clear the memory cache as this is what we'd expect when invalidating on the retool side to get fresh rankings

## Checklist

### DIFF
--- a/web/api/hasura/invalidate-cache/index.ts
+++ b/web/api/hasura/invalidate-cache/index.ts
@@ -1,4 +1,5 @@
 import { errorHasuraQuery } from "@/api/helpers/errors";
+import { clearMetricsCache } from "@/api/helpers/fetch-metrics";
 import { protectInternalEndpoint } from "@/api/helpers/utils";
 import { logger } from "@/lib/logger";
 import {
@@ -81,6 +82,8 @@ export const POST = async (req: NextRequest) => {
 
     const command = new CreateInvalidationCommand(input);
     await client.send(command);
+
+    await clearMetricsCache();
 
     return NextResponse.json({ success: true });
   } catch (error) {
--- a/web/api/hasura/invalidate-cache/index.ts
+++ b/web/api/hasura/invalidate-cache/index.ts
@@ -80,10 +80,10 @@ export const POST = async (req: NextRequest) => {
       },
     };
 
+    await clearMetricsCache();
+
     const command = new CreateInvalidationCommand(input);
     await client.send(command);
-
-    await clearMetricsCache();
 
     return NextResponse.json({ success: true });
   } catch (error) {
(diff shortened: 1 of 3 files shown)


web/api/hasura/invalidate-cache/index.ts
```diff
@@ -82,6 +83,8 @@ export const POST = async (req: NextRequest) => {
     const command = new CreateInvalidationCommand(input);
     await client.send(command);
 
+    await clearMetricsCache();
+
```
aurel-fr: This should come before the Cloudfront invalidation to ensure when CF queries for fresh data it won't ingest the old cache entries again.

web/api/helpers/fetch-metrics.ts
```diff
@@ -1,141 +1,193 @@
...
+interface ProcessedMetricsCache {
+  byCountry: Map<string, AppStatsReturnType>;
+  global: AppStatsReturnType;
+  timestamp: number;
+}
+
+let memoryCache: ProcessedMetricsCache | null = null;
+let pendingFetch: Promise<ProcessedMetricsCache | null> | null = null;
+
+const MEMORY_CACHE_TTL = 2 * 60 * 60 * 1000; // 2 hours
```
aurel-fr: This feels like a lot, since the dev portal is the source of truth for every other services. Even with a robust cache invalidation strategy, I'd set this ttl to a dozen minutes, considering our query is now highly efficient.

web/api/helpers/fetch-metrics.ts
```diff
@@ -1,141 +1,193 @@
...
+
+/**
+ * Fetch and process raw metrics data, caching results for all countries
+ */
+const fetchAndProcessMetrics =
+  async (): Promise<ProcessedMetricsCache | null> => {
```
aurel-fr: this doesn't look like it ever returns `null`?

web/api/helpers/fetch-metrics.ts
```diff
@@ -1,141 +1,193 @@
...
-  if (pendingRequest) {
-    return pendingRequest;
+  // Cache is stale or doesn't exist, need to fetch fresh data
+  // If there's already a pending fetch, await it (request deduplication)
+  if (pendingFetch) {
+    const result = await pendingFetch;
+    if (!result) {
+      console.error("Failed to fetch metrics data");
+      return [];
+    }
```
aurel-fr: I could be wrong but this check will not be needed if the outcome is either a result or an exception

aurel-fr (approved):

PR #1534, opened 2025-07-25
feat: Create a helper client for AWS SSM ParameterStore

Gr1dlock:
## PR Type

-  [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-1916](https://linear.app/worldcoin/issue/DEV-1916/setup-ssm-for-dev-portal-so-we-can-do-feature-flags)

This PR adds a TypeScript utility class ParameterStore that simplifies working with AWS SSM Parameter Store. It allows:
	•	Getting a parameter value, with support for a default fallback if the parameter does not exist
	•	Listing all parameters registered for the given service prefix, returning their normalized names

### Parameter Store Overview

AWS Systems Manager Parameter Store is a centralized storage service for configuration data and secrets management. It’s commonly used for managing environment variables, feature flags, secrets, and other service-level configuration.

We follow a naming convention where parameters are stored as hierarchical paths using slashes, e.g.: 
```
/developer-portal/feature-flags/enable-new-auth
```
Within each path segment, names are normalized to lowercase kebab-case (e.g. enable-new-auth).

On initialization, the helper class requires a serviceName, which will be prepended to all parameter names as a prefix (e.g. `/my-service/...`). This allows grouping and scoping of parameters by service.

There are three supported types: String (plaintext), StringList (comma-separated values), and SecureString (encrypted).

### Example Usage
```ts
const store = new ParameterStore("developer-portal");

const showRefactor = await store.getParameter("feature-flags/show-refactor", "false");
console.log("Show refactored version:", showRefactor === "true"); // Show refactored version: true

const allParams = await store.listParameters();
console.log(allParams); // [ 'feature-flags/show-refactor', ... ]
```

## Checklist


web/package.json
```diff
@@ -6,6 +6,7 @@
     "@aws-sdk/client-cloudfront": "^3.664.0",
     "@aws-sdk/client-kms": "^3.664.0",
     "@aws-sdk/client-s3": "^3.664.0",
+    "@aws-sdk/client-ssm": "^3.849.0",
```
andy-t-wang: One small note is it bad to use different aws package versions sometimes they might have errors

web/lib/parameter-store.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Create or update a parameter.
+   *
+   * @param name - The name of the parameter to create or update.
+   * @param value - The value of the parameter to create or update.
+   * @param type - The type of the parameter to create or update.
+   * @param overwrite - Whether to overwrite if already exists.
+   */
+  async putParameter(
```
andy-t-wang: Why would we allow PUT?

andy-t-wang (approved):

web/package.json
Gr1dlock: Good point, pinned it to `3.664.0`

web/lib/parameter-store.ts
Gr1dlock: It depends on the use case - if we have some parameters, that should be set dynamically, then we could use this function.
If you ask why it is `put` and not smth like `update`, then the answer is - there's no update function in the AWS SSM API, it's always `put` (you just specify the `overwrite` flag if you want to update the existing parameter)

andy-t-wang: I think we should just remove PUT for now then add when we need.

Gr1dlock: Alright, YAGNI it is

PR #1539, opened 2025-07-29
feat: Create a helper client for AWS SSM ParameterStore

Gr1dlock:
## PR Type

-  [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-1916](https://linear.app/worldcoin/issue/DEV-1916/setup-ssm-for-dev-portal-so-we-can-do-feature-flags)

This PR adds a TypeScript utility class ParameterStore that simplifies working with AWS SSM Parameter Store. It allows:
	•	Getting a parameter value, with support for a default fallback if the parameter does not exist
	•	Listing all parameters registered for the given service prefix, returning their normalized names

All the parameters are cached in memory for 30 minutes to avoid rate limiting.

### Parameter Store Overview

AWS Systems Manager Parameter Store is a centralized storage service for configuration data and secrets management. It’s commonly used for managing environment variables, feature flags, secrets, and other service-level configuration.

We follow a naming convention where parameters are stored as hierarchical paths using slashes, e.g.: 
```
/developer-portal/feature-flags/enable-new-auth
```
Within each path segment, names are normalized to lowercase kebab-case (e.g. enable-new-auth).

On initialization, the helper class requires a serviceName, which will be prepended to all parameter names as a prefix (e.g. `/my-service/...`). This allows grouping and scoping of parameters by service.

There are three supported types: String (plaintext), StringList (comma-separated values), and SecureString (encrypted).

### Example Usage
```ts
const store = new ParameterStore("developer-portal");

const showRefactor = await store.getParameter("feature-flags/show-refactor", "false");
console.log("Show refactored version:", showRefactor === "true"); // Show refactored version: true

const allParams = await store.listParameters();
console.log(allParams); // [ 'feature-flags/show-refactor', ... ]
```

## Checklist


web/api/helpers/app-store.ts
```diff
@@ -55,13 +51,23 @@ export const formatAppMetadata = async (
...
+  );
+
+  const whitelistedAppsContracts = await global.ParameterStore?.getParameter(
+    "whitelisted-apps/contracts",
+    [] as string[],
+  );
+
   // Check if the app is whitelisted for permit2
-  const permit2Tokens = whitelistedAppsPermit2.includes(appMetadata.app_id)
+  const permit2Tokens = whitelistedAppsPermit2?.includes(appMetadata.app_id)
```
andy-t-wang, line 65: We should log if it's not found so we can be explicit the item exists. Since the only reason they wouldn't exist is if someone accidentally deleted it

web/lib/constants.ts
```diff
@@ -136,25 +136,6 @@ export const NativeApps: Record<string, NativeAppsMap> = {
   },
 };
 
-// These are addresses that have been whitelisted from the permit2 restriction
```
andy-t-wang, line 139: You probably did but just in case, make sure these are already migrated into parameter store

andy-t-wang (approved):

web/api/helpers/app-store.ts
Gr1dlock: We already log it in the ParameterStore class

web/lib/constants.ts
Gr1dlock: Yep, they already exist there

PR #1553, opened 2025-08-04
add individual meta queries

andy-t-wang:
## PR Type

- [ ] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Add individual meta queries 

## Checklist

### DIFF
--- a/web/api/v2/public/app/[app_id]/index.ts
+++ b/web/api/v2/public/app/[app_id]/index.ts
@@ -17,6 +17,16 @@ import { getSdk as getAppMetadataSdk } from "./graphql/get-app-metadata.generate
  * @param res
  */
 
+const independentFieldsToFetch = [
+  "name",
+  "short_name",
+  "supported_countries",
+  "world_app_description",
+  "unique_users",
+  "impressions",
+  "logo_img_url",
+];
+
 export async function GET(
   request: Request,
   { params }: { params: { app_id: string } },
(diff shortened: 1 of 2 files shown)


web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -214,8 +225,49 @@ export async function GET(
...
+    // Return raw value as text/plain for primitive types
+    if (
+      typeof fieldValue === "string" ||
+      typeof fieldValue === "number" ||
+      typeof fieldValue === "boolean"
+    ) {
+      const responseText = String(fieldValue);
+      const contentLength = Buffer.byteLength(responseText, "utf-8").toString();
+
+      return new NextResponse(responseText, {
```
Comment (Gr1dlock) [inline web/api/v2/public/app/[app_id]/index.ts:243]: Is it a requirement to return it plain?
Otherwise I'd unify it to return a JSON in the format
`{"metadata_field": "value"}`

web/api/v2/public/app/[app_id]/index.ts
```diff
@@ -133,6 +143,7 @@ export async function GET(
     ? null
     : headers.get("CloudFront-Viewer-Country");
   const override_country = searchParams.get("override_country") || country;
+  const metadata_field = searchParams.get("metadata_field");
```
Comment (Gr1dlock) [inline web/api/v2/public/app/[app_id]/index.ts:146]: Should we allow fetching only a single field or could it be an array?

web/tests/api/v2/public/apps/app.test.ts
Gr1dlock, line 1: Could you also add these tests to `tests/api/specs/public/public-apps.spec.ts` please?

web/api/v2/public/app/[app_id]/index.ts
Comment (andy-t-wang) [reply web/api/v2/public/app/[app_id]/index.ts:243]: Yes for their braze stuff they need plain text

web/api/v2/public/app/[app_id]/index.ts
Comment (andy-t-wang) [reply web/api/v2/public/app/[app_id]/index.ts:146]: Unfortunately braze only lets you fetch one field at a time. It's very rudimentary

PR #1566, opened 2025-08-06
feat: Add dd-trace setup

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-2097](https://linear.app/worldcoin/issue/DEV-2097/dev-portal-rework-dd-trace-initialization)

Introduces custom `dd-trace` initialization, augmenting HTTP requests trace processing and removing Next.js traces.

## Checklist

### DIFF
--- a/web/instrumentation.ts
+++ b/web/instrumentation.ts
@@ -6,6 +6,41 @@ export async function register() {
       typeof window === "undefined" &&
       process.env.NEXT_RUNTIME === "nodejs"
     ) {
+      console.log("Registering dd-trace");
+
+      const ddTrace = await import("dd-trace");
+
+      const tracer = ddTrace.default.init();
+
+      // Monitor HTTP requests
+      tracer.use("http", {
+        hooks: {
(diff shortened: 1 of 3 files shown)


web/instrumentation.ts
```diff
@@ -6,6 +6,41 @@ export async function register() {
...
+
+      // Monitor HTTP requests
+      tracer.use("http", {
+        hooks: {
+          request(span, req, res) {
+            if (span && req) {
+              const urlString = "path" in req ? req.path : req.url;
+
+              if (urlString) {
+                const url = new URL(urlString, "http://localhost");
```
andy-t-wang, line 25: Should fallback be local host? Not just the base domain we have. I think it's in an env var for staging and prod

Gr1dlock: It doesn't really matter, we use URL class to retrieve the path group, so the host doesn't play any role

PR #1579, opened 2025-08-11
nested scroll -> tabs

michalstruck:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
Change UI from nested scroll back into tabs.
- adds error highlights on the tab (red border), as well as on individual fields when erroring tab is selected
- all form changes are persisted, without save queries between tab switches - single "Save" to update all localisations
- if user starts uploading an image, then switches tabs without waiting for it to finish -> the upload is lost. This can be addressed separately, extensive changes are required since the current set up is not designed to handle this. In the previous version the solution was to block any form interactions before upload finished.
- no errors when adding a new tab
- errors clear both from tab and fields dynamically, when addressed
- images load lazily, when switching tabs. after first load they're in memory and responsive fast
- components split up to separate concerns

Informative toasts when fields are missing in a separate PR

Screenshot 2025-08-11 at 14 19 39

Screenshot 2025-08-11 at 14 20 30

Screenshot 2025-08-11 at 14 20 41

## Checklist


michalstruck:
https://github.com/worldcoin/developer-portal/pull/1580
followup with better handling for aborted uploads

PR #1580, opened 2025-08-11
abort image upload with user feedback

michalstruck:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
There isn't a mechanism to continue uploads in the background right now. When user starts uploading an image and switches localization tabs, the image is lost without any feedback.

An interim solution is to abort the upload, and let the user know. Additional benefit is no orphaned files on s3

## Checklist

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
@@ -149,8 +149,7 @@ export const ImageUploadField = (props: ImageUploadFieldProps) => {
           return;
         }
 
-        if (error instanceof ImageValidationError) {
-        } else {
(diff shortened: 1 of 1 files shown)


web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/ImageForm/ImageUploadField.tsx
```diff
@@ -139,12 +144,17 @@ export const ImageUploadField = (props: ImageUploadFieldProps) => {
       } catch (error) {
         console.error("error uploading image:", error);
 
+        if (error instanceof Error && error.name === "AbortError") {
+          toast.error("Upload was cancelled", { autoClose: 5000 });
+          return;
+        }
+
         if (error instanceof ImageValidationError) {
```
andy-t-wang: I removing the comment can just make this inverted so it doesn't look weird
```suggestion
        if (!(error instanceof ImageValidationError)) {
```

andy-t-wang (approved):

PR #1582, opened 2025-08-12
fix: Fix transaction error handling

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
Ticket - [DEV-2192](https://linear.app/worldcoin/issue/DEV-2192/dev-portal-transactions-tab-throws-an-error-because-of-the-server-side)

Currently, if the transaction request fails, frontend crushes with an error.
This happens because:
* Next.js fails to serialize errors, that are passed from the server-side actions
* We use toast in the server-rendered component

This PR:
* Adds a functionality to properly serialize server-side errors
* Adds an error screen to handle the case, when the transactions request fails. It should be handled separately from the case, when the request actually goes through, but the data is empty.
image

## Checklist

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Transactions/page/index.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Transactions/page/index.tsx
@@ -11,55 +11,89 @@ type TransactionsPageProps = {
   params: Record<string, string> | null | undefined;
 };
 
-export const TransactionsPage = async (props: TransactionsPageProps) => {
-  const { params } = props;
-  const appId = params?.appId as `app_${string}`;
-
-  const result = await getTransactionData(appId);
-  let transactionData: PaymentMetadata[] = [];
-  if (result.success) {
-    transactionData = result.data as PaymentMetadata[];
-  }
-
+const TransactionsPageLayout = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
   return (
     <div className="my-6 min-h-[100dvh]">
       <div className="flex items-center justify-start gap-x-2 text-gray-900">
         <Typography variant={TYPOGRAPHY.H6}>Payments</Typography>
       </div>
       <hr className="mt-5 w-full border-dashed text-grey-200" />
-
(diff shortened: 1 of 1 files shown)


web/api/helpers/errors.ts
```diff
@@ -207,5 +207,9 @@ export function errorFormAction({
     team_id,
   });
 
-  return { success: false, message, error };
+  const serializedError = error
```
michalstruck, line 210: so we're not adding the package after all

web/scenes/Portal/Teams/TeamId/Apps/AppId/Transactions/page/index.tsx
```diff
@@ -17,9 +17,7 @@ export const TransactionsPage = async (props: TransactionsPageProps) => {
 
   const result = await getTransactionData(appId);
   let transactionData: PaymentMetadata[] = [];
-  if (!result.success) {
-    toast.error(result.message);
-  } else {
+  if (result.success) {
     transactionData = result.data as PaymentMetadata[];
```
michalstruck: instead of assigning a var and using it later on, we can just return the JSX and directly use the result - this would use a more established react pattern. Same goes for `<ErrorState />` etc. So basically early returns with JSX

web/api/helpers/errors.ts
Gr1dlock: Yeah, the most used package for that has conflicts with jest, so had to implement its simplified version

web/scenes/Portal/Teams/TeamId/Apps/AppId/Transactions/page/index.tsx
Gr1dlock: Refactored it, is that what you meant?

michalstruck: yep exactly

michalstruck (approved):
please check if it looks good on smaller screens/mobile before merging

PR #1583, opened 2025-08-12
add informative toasts to the app store form

michalstruck:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Adds toasts when there are errors in form and user tries to submit. Right now we're only highlighting all errors, now we also let them know via a toast.
- if there's a single error, the toast lets the user know which one and where. For errors in localizations, specifies the localization name + field error. Example: `English: Short name is required`
- if there are more than 1 error, toast says `There are multiple errors in the form` - this is easily changeable to show the first found error. Issue here is that the "first" error is different when submitting the form using `Save` and `Submit for Review`. This is due to how yup and react-hook-form group errors. If we're fine with these two errors being possibly different (but still valid) then I will remove the `There are multiple errors in the form` string.
- improved various error strings in validators (from `This field is required` to a field-specific error)
- improved logo image error display margins

## Checklist

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/FormSchema/form-schema.ts
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/FormSchema/form-schema.ts
@@ -31,25 +31,18 @@ export const mainAppStoreFormSchema = yup
   .object({
     category: categorySchema,
     app_website_url: appWebsiteUrlSchema,
-    support_link: supportLinkSchema.test({
-      message: "Either support link or support email must be provided",
-      test: function (this, _, ctx) {
-        const { support_link, support_email } = ctx.parent || {};
-
-        // exactly one must be provided
-        return !!(support_link || support_email);
-      },
-    }),
-    support_email: supportEmailSchema.test({
-      message: "Either support link or support email must be provided",
-      test: function (this, _, ctx) {
-        const { support_link, support_email } = ctx.parent || {};
-
-        // exactly one must be provided
-        return !!(support_link || support_email);
-      },
-    }),
     support_type: yup.string().oneOf(["email", "link"]),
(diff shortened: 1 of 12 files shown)


web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStoreRefactored/FormSchema/form-schema.ts
```diff
@@ -31,25 +31,18 @@ export const mainAppStoreFormSchema = yup
...
     app_website_url: appWebsiteUrlSchema,
-    support_link: supportLinkSchema.test({
-      message: "Either support link or support email must be provided",
-      test: function (this, _, ctx) {
-        const { support_link, support_email } = ctx.parent || {};
-
-        // exactly one must be provided
-        return !!(support_link || support_email);
-      },
+    support_type: yup.string().oneOf(["email", "link"]),
```
michalstruck, line 34: these are copied from the submit for review validator, behavior stayed the same, imo cleaner and better errors

web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppTopBarRefactored/index.tsx
```diff
@@ -142,21 +154,22 @@ export const AppTopBarRefactored = (props: AppTopBarProps) => {
...
             },
             (errors) => {
-              console.error("Form validation errors:", errors);
-              toast.error("Please fix form errors before submitting");
-              reject(new Error("Form validation failed"));
+              const errorMessage = getFirstFormError(
+                errors,
+                form.getValues("localisations"),
+              );
+              reject(new Error(errorMessage ?? "Form validation failed"));
```
michalstruck, line 168: these rejections are handled in the `catch` block, to prevent duplicate toasts and aggregate errors

PR #1584, opened 2025-08-12
improve open rate chart displays

michalstruck:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description
- currently, the label at the top of the chart shows the last entry in metrics list
  - this could result in it being 0, even though users may expect an average instead

- show average notification open rate always, instead of last value
- if there's less than 7 days of data, don't show the badge eligibility icon
- if there's more than 7 days of data, show the icon

The above is in line with app-backend logic - it checks `avg_notification_open_rate` on metadata, which is only set if more than 7 days of data is present

## Checklist

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/page/AppStatsGraph/GraphsSection/index.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/page/AppStatsGraph/GraphsSection/index.tsx
@@ -224,10 +224,12 @@ const GraphCard: React.FC<GraphCardProps> = ({
 
 const isEligibleForNotificationBadge = (
   avg_notification_open_rate: number | null,
-  is7DaysOfNotificationOpenRateData: boolean,
+  has7DaysOfNotificationOpenRateData: boolean,
 ) => {
   if (!avg_notification_open_rate) return false;
-  return avg_notification_open_rate > 0.1 && !is7DaysOfNotificationOpenRateData;
+  if (!has7DaysOfNotificationOpenRateData) return false;
+
+  return avg_notification_open_rate > 0.1;
 };
 
 // ==================================================================================================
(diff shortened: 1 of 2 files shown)


web/scenes/Portal/Teams/TeamId/Apps/AppId/page/AppStatsGraph/GraphsSection/index.tsx
```diff
@@ -225,9 +224,10 @@ const GraphCard: React.FC<GraphCardProps> = ({
 
 const isEligibleForNotificationBadge = (
   avg_notification_open_rate: number | null,
+  is7DaysOfNotificationOpenRateData: boolean,
 ) => {
   if (!avg_notification_open_rate) return false;
-  return avg_notification_open_rate > 0.1;
+  return avg_notification_open_rate > 0.1 && !is7DaysOfNotificationOpenRateData;
```
andy-t-wang: I'm confused if they have 7 days of data and the open rate is above .1 we want to show the badge right? This will just always be false if they do have 7 days of data

web/scenes/Portal/Teams/TeamId/Apps/AppId/page/AppStatsGraph/GraphsSection/index.tsx
```diff
@@ -225,9 +224,10 @@ const GraphCard: React.FC<GraphCardProps> = ({
 
 const isEligibleForNotificationBadge = (
   avg_notification_open_rate: number | null,
+  is7DaysOfNotificationOpenRateData: boolean,
```
andy-t-wang:
```suggestion
  has7DaysOfNotificationOpenRateData: boolean,
```

web/scenes/Portal/Teams/TeamId/Apps/AppId/page/AppStatsGraph/GraphsSection/index.tsx
michalstruck: you're right i didn't invert the condition after tests

michalstruck: that's correct, i forgot to invert the condition after tests

PR #1589, opened 2025-08-14
feat: Add field to indicate first verification timestamp

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Ticket - [DEV-2201](https://linear.app/worldcoin/issue/DEV-2201/dev-portal-create-a-field-for-first-verification-timestamp)

Introduces `first_verified_at` field to indicate when the app was verified for the first time.

## Checklist

### DIFF
--- a/hasura/metadata/databases/default/tables/public_app.yaml
+++ b/hasura/metadata/databases/default/tables/public_app.yaml
@@ -40,7 +40,6 @@ insert_permissions:
       check: {}
       columns:
         - engine
-        - first_verified_at
         - is_staging
         - name
         - team_id


hasura/metadata/databases/default/tables/public_app.yaml
```diff
@@ -40,6 +40,7 @@ insert_permissions:
       check: {}
       columns:
         - engine
+        - first_verified_at
```
andy-t-wang: Are we backfilling? Why does service need the ability to insert this?

Gr1dlock: Hmm, true, checked and even when the field is null it's considered an update

PR #1600, opened 2025-08-22
feat: Allow multiple domains

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-2173](https://linear.app/worldcoin/issue/DEV-2173/dev-portal-migrate-to-new-domain)

Allows running Developer Portal simultaneously on multiple domains.

## Checklist

### DIFF
--- a/web/api/auth/[auth0]/index.ts
+++ b/web/api/auth/[auth0]/index.ts
@@ -1,4 +1,5 @@
 import { deleteAccount } from "@/api/delete-account";
+import { getAppUrlFromRequest } from "@/api/helpers/utils";
 import { loginCallback } from "@/api/login-callback";
 import { urls } from "@/lib/urls";
 
@@ -17,19 +18,43 @@ export const GET = handleAuth({
   login: (req: NextRequest, ctx: AppRouteHandlerFnContext) => {
     const invite_id = req.nextUrl.searchParams.get("invite_id") ?? undefined;
     const returnTo = req.nextUrl.searchParams.get("returnTo") ?? undefined;
+    const appUrl = getAppUrlFromRequest(req);
+    const redirect_uri = `${appUrl}/api/auth/callback`;
+    const authReturnTo = new URL(
+      urls.api.loginCallback({ invite_id, returnTo }),
+      appUrl,
+    ).toString();
 
     return handleLogin(req, ctx, {
-      returnTo: urls.api.loginCallback({ invite_id, returnTo }),
+      returnTo: authReturnTo,
+      authorizationParams: {
+        redirect_uri: redirect_uri,
+      },
     });
   },
 
-  callback: handleCallback,
+  callback: (req: NextRequest, ctx: AppRouteHandlerFnContext) => {
+    const appUrl = getAppUrlFromRequest(req);
+    const redirect_uri = `${appUrl}/api/auth/callback`;
+
+    return handleCallback(req, ctx, {
+      authorizationParams: {
+        redirect_uri: redirect_uri,
+      },
+      redirectUri: redirect_uri,
+    });
+  },
+
   "login-callback": loginCallback,
   "delete-account": deleteAccount,
 
-  logout: handleLogout({
-    returnTo: urls.login(),
-  }),
+  logout: (req: NextRequest, ctx: AppRouteHandlerFnContext) => {
+    const appUrl = getAppUrlFromRequest(req);
+    const logoutReturnTo = new URL(urls.login(), appUrl).toString();
+    return handleLogout(req, ctx, {
+      returnTo: logoutReturnTo,
+    });
+  },
 
   onError: (_req: NextRequest, error: HandlerError) => {
     NextResponse.json({
--- a/web/api/create-team/index.ts
+++ b/web/api/create-team/index.ts
@@ -12,6 +12,7 @@ import { NextRequest, NextResponse } from "next/server";
 import * as yup from "yup";
 import { getAPIServiceGraphqlClient } from "../helpers/graphql";
 import { isEmailUser } from "../helpers/is-email-user";
+import { getAppUrlFromRequest } from "../helpers/utils";
 
 import {
   InsertTeamMutation,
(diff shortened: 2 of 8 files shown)


web/api/helpers/utils.ts
```diff
@@ -71,3 +71,34 @@ export function corsHandler(response: NextResponse, methods: string[]) {
...
+    if (
+      (protocol === "https" && host.endsWith(":443")) ||
+      (protocol === "http" && host.endsWith(":80"))
+    ) {
+      host = host.split(":")[0];
+    }
+    return `${protocol}://${host}`;
+  }
+
+  return process.env.NEXT_PUBLIC_APP_URL || "http://localhost:3000";
```
andy-t-wang: is local host a good default here? Perhaps we default it to the prod domain

web/api/helpers/utils.ts
```diff
@@ -71,3 +71,34 @@ export function corsHandler(response: NextResponse, methods: string[]) {
...
   return response;
 }
+
+/**
+ * Extracts the app URL from the request headers, supporting multiple domains
+ * Falls back to NEXT_PUBLIC_APP_URL if unable to determine from request
+ * @param req - The NextRequest object
+ * @returns The app URL (protocol + host)
+ */
+export function getAppUrlFromRequest(req: NextRequest): string {
```
andy-t-wang: I wonder if we should code in two defaults we expect since it's not like there are really N different domains. If we for some reason see this getting the wrong domain it will reject. 

Mb there's some security risk with a non hard coded constant domain url. Additionally if we get some weird broken logic can throw an error we can catch in DD

andy-t-wang (approved):

web/api/hasura/invite-team-members/index.ts
```diff
@@ -222,7 +225,8 @@ export const POST = async (req: NextRequest) => {
   const promises = [];
 
   for (const invite of invites) {
-    const link = `${process.env.NEXT_PUBLIC_APP_URL}/join?invite_id=${invite.id}`;
+    const appUrl = getAppUrlFromRequest(req);
+    const link = `${appUrl}/join?invite_id=${invite.id}`;
```
igorosip0v: @Gr1dlock I'd prefer crafting links with URL constructor

```
const link = new URL('/join', appUrl)
link.searchParams.append('invite', invite.id)
```

web/api/create-team/index.ts
```diff
@@ -95,7 +96,8 @@ export const POST = withApiAuthRequired(async (req: NextRequest) => {
     ironCladUserId = crypto.randomUUID();
 
     try {
-      const url = new URL(`${process.env.NEXT_PUBLIC_APP_URL}/signup`);
+      const appUrl = getAppUrlFromRequest(req);
+      const url = new URL(`${appUrl}/signup`);
```
igorosip0v:
```suggestion
      const url = new URL("/signup", appUrl);
```

web/api/auth/[auth0]/index.ts
```diff
@@ -17,19 +18,43 @@ export const GET = handleAuth({
...
+
+  callback: (req: NextRequest, ctx: AppRouteHandlerFnContext) => {
+    const appUrl = getAppUrlFromRequest(req);
+    const redirect_uri = `${appUrl}/api/auth/callback`;
+
+    return handleCallback(req, ctx, {
+      authorizationParams: {
+        redirect_uri: redirect_uri,
+      },
+      redirectUri: redirect_uri,
```
Comment (igorosip0v) [inline web/api/auth/[auth0]/index.ts]: @Gr1dlock are you sure this change is needed? I think handleCallback already takes the base URL from the host; otherwise, how was it working before without explicit URL setting?

Comment (igorosip0v) [reply web/api/auth/[auth0]/index.ts]: But if you want prio other headers, then it makes sense

web/api/helpers/utils.ts
igorosip0v: Maybe it's worth to add some checks for expected domains. To make it less hardcoded, you can put a stringified array of domains in some ENV and then JSON.parse it.

web/api/auth/[auth0]/index.ts
```diff
@@ -17,19 +18,43 @@ export const GET = handleAuth({
...
+      returnTo: authReturnTo,
+      authorizationParams: {
+        redirect_uri: redirect_uri,
+      },
+    });
+  },
+
+  callback: (req: NextRequest, ctx: AppRouteHandlerFnContext) => {
+    const appUrl = getAppUrlFromRequest(req);
+    const redirect_uri = `${appUrl}/api/auth/callback`;
```
Comment (igorosip0v) [inline web/api/auth/[auth0]/index.ts]:
```suggestion
    const redirect_uri = new URL("/api/auth/callback", appUrl).toString();
```

web/api/join-callback/index.ts
```diff
@@ -75,7 +77,7 @@ export const POST = withApiAuthRequired(async (req: NextRequest) => {
   ironCladUserId = crypto.randomUUID();
 
   try {
-    const url = new URL(`${process.env.NEXT_PUBLIC_APP_URL}/signup`);
+    const url = new URL(`${appUrl}/signup`);
```
igorosip0v:
```suggestion
    const url = new URL("/signup", appUrl);
```

igorosip0v: also maybe worth use it with `urls.signUp()` as it is in other places instead of "/signup"

web/api/auth/[auth0]/index.ts
Comment (Gr1dlock) [reply web/api/auth/[auth0]/index.ts]: Hmm, I wanted to override the default `AUTH0_BASE_URL`

web/api/helpers/utils.ts
Gr1dlock: Well, `NEXT_PUBLIC_APP_URL` should always be set, and localhost is a second default. But sure, can change it to the prod one

web/api/helpers/utils.ts
Gr1dlock: Good idea, let's move it to Parameter Store, so that it's gonna be env-specific

PR #1607, opened 2025-08-28
feat: Introduce team soft-deletion

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-2054](https://linear.app/worldcoin/issue/DEV-2054/dev-portal-prevent-app-deletions-when-team-is-deleted)

Stop deleting teams and introduce a soft-deletion mechanism instead (like we have for apps).
When a team is deleted, the row remains in the DB, but it gets the `deleted_at` column value set, all its memberships, invites  and api keys get deleted and all its apps get soft-deleted.

## Checklist

### DIFF
--- a/hasura/metadata/databases/default/tables/public_action.yaml
+++ b/hasura/metadata/databases/default/tables/public_action.yaml
@@ -228,7 +228,7 @@ delete_permissions:
       filter:
         _and:
           - action:
-              _neq: '""'
+              _neq: ""
           - app:
               _and:
                 - deleted_at:


hasura/metadata/databases/default/tables/public_team.yaml
```diff
@@ -70,13 +72,17 @@ select_permissions:
         - name
         - updated_at
       filter:
-        memberships:
-          user_id:
-            _eq: X-Hasura-User-Id
+        _and:
+          - memberships:
```
andy-t-wang, line 76: I wonder if once we delete a team we can rename the rows to something like name="deleted_team" so we still preserve data deletion. I dont like how information is persisted here

hasura/metadata/databases/default/tables/public_app.yaml
andy-t-wang, line 1: Don't we need this in more places. Like could you still verify a proof for an app from a deleted team? That shouldn't be possible. I think you can still write to the nullifier table for isntance

hasura/metadata/databases/default/tables/public_team.yaml
Gr1dlock: Wasn't it the intent of this task, to preserve the team data?

hasura/metadata/databases/default/tables/public_app.yaml
Gr1dlock: True, not for the team though, but for the app. Seems like Michał missed checks for actions and nullifiers

andy-t-wang (approved):

andy-t-wang:
Integration test failing

PR #1629, opened 2025-09-09
feat: Add apps stats rollup cron job

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-2242](https://linear.app/worldcoin/issue/DEV-2242/dev-portal-rework-apps-stats-query)

Switches app stats from per-event updates to an hourly, set-based rollup invoked by a cron endpoint.
Removes the hot-path mutation that ran on every nullifier change.

## Checklist


hasura/migrations/default/1757453727113_rework_app_stats_functionality/up.sql
```diff
@@ -0,0 +1,142 @@
...
+
+-- 3) Last-seen snapshot per nullifier (so we apply only deltas)
+CREATE TABLE IF NOT EXISTS public.nullifier_uses_seen (
+  nullifier_hash  TEXT PRIMARY KEY,
+  last_seen_uses  INTEGER NOT NULL,
+  last_seen_at    TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+
+-- 4) Helpful indexes for the rollup
+CREATE INDEX IF NOT EXISTS idx_nullifier_action_id  ON public.nullifier (action_id);
```
andy-t-wang: Hmm this might be expensive the prod nullifier table is huge.

hasura/migrations/default/1757453727113_rework_app_stats_functionality/up.sql
```diff
@@ -0,0 +1,142 @@
...
+      (_since IS NULL OR n.updated_at >= _since)
+      AND n.updated_at < _until
+  ),
+  changed AS (
+    SELECT
+      c.nullifier_hash,
+      c.action_id,
+      c.updated_at,
+      c.uses,
+      GREATEST(c.uses - COALESCE(c.last_seen_uses, 0), 0)::BIGINT AS delta_uses
```
andy-t-wang: Why would this ever be negative?

andy-t-wang (commented):
This cron job might be really expensive on CPU. At a certain point we might just want to move to metrics service. I don't think this will scale well

hasura/migrations/default/1757453727113_rework_app_stats_functionality/up.sql
Gr1dlock: Ah, true, c.uses > COALESCE(c.last_seen_uses, 0) makes it impossible

hasura/migrations/default/1757453727113_rework_app_stats_functionality/up.sql
Gr1dlock: Changed it to concurrent execution and will run manually from pgadmin

PR #1630, opened 2025-09-10
feat: Move dev_rewards_approved and has_airdrop_component to app table

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Ticket - [DEV-2237](https://linear.app/worldcoin/issue/DEV-2237/dev-portal-make-sure-apps-keep-developer-awards-status)

Moves the `dev_rewards_approved` and `has_airdrop_component` columns to the `app` table since they are set only in the reviewer to the app table (to avoid them being duplicated every time a new metadata row is created)

## Checklist

### DIFF
--- a/hasura/migrations/default/1757506407060_move_columns_dev_rewards_approved_and_has_airdrop_component/up.sql
+++ b/hasura/migrations/default/1757506407060_move_columns_dev_rewards_approved_and_has_airdrop_component/up.sql
@@ -1,1 +1,6 @@
 alter table "public"."app_metadata" drop column "has_airdrop_component" cascade;
+alter table "public"."app_metadata" drop column "dev_rewards_approved" cascade;
+alter table "public"."app" add column "dev_rewards_approved" bool
+ not null default 'false';
+alter table "public"."app" add column "has_airdrop_component" bool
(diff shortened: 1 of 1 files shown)


hasura/migrations/default/1757506407060_move_columns_dev_rewards_approved_and_has_airdrop_component/up.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."app_metadata" drop column "has_airdrop_component" cascade;
```
Takaros999, line 1: this down migration seems incomplete

Takaros999 (commented):
where are these fields set? don't we need to change app store form or something?

hasura/migrations/default/1757506407060_move_columns_dev_rewards_approved_and_has_airdrop_component/up.sql
Gr1dlock: well, have you read the file name? 😅

Takaros999 (approved):

PR #1646, opened 2025-10-01
feat: Increase request timeout

Gr1dlock:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Ticket - [DEV-2322](https://linear.app/worldcoin/issue/DEV-2322/dev-portal-investigate-get-transaction-endpoint-failures) & [DEV-2309](https://linear.app/worldcoin/issue/DEV-2309/fix-transaction-endpoint-error-propagation)

* Increases the request timeout to 5 seconds (we have some requests that take more than 3 seconds on the app backend side)
* Fixes the error propagation from app backend

## Checklist

### DIFF
--- a/web/api/v2/minikit/transaction/[transaction_id]/index.ts
+++ b/web/api/v2/minikit/transaction/[transaction_id]/index.ts
@@ -4,7 +4,7 @@ import { validateRequestSchema } from "@/api/helpers/validate-request-schema";
 import { logger } from "@/lib/logger";
 import { appIdSchema } from "@/lib/schema";
 import { TransactionTypes } from "@/lib/types";
-import { fetchWithRetry } from "@/lib/utils";
+import { fetchWithTimeout } from "@/lib/utils";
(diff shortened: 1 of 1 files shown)


web/api/v2/minikit/transaction/[transaction_id]/index.ts
```diff
@@ -95,22 +107,13 @@ export const GET = async (
...
-    });
-
     return corsHandler(
       errorResponse({
         statusCode: res.status,
         code: errorBody?.error?.code ?? "internal_api_error",
         detail:
-          errorBody?.error?.details ?? "Transaction fetch to backend failed",
-        attribute: errorBody?.error?.message ?? "transaction",
+          errorBody?.error?.message ?? "Transaction fetch to backend failed",
```
Comment (andy-t-wang) [inline web/api/v2/minikit/transaction/[transaction_id]/index.ts:115]: can you double check this once deployed? I had some issues with parsing the JSON before where it was being returned as a string and indexing gave undefined

PR #1857, opened 2026-04-07
fix: optimize /api/v2/public/apps performance and increase app limit

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

The `/api/v2/public/apps` endpoint was hitting performance limits, causing the dev portal to go down and mini apps to become unavailable. This PR fixes several algorithmic bottlenecks and raises the default app limit from 750 to 1000 (max from 1000 to 1500):

- **Hoist ParameterStore calls**: `formatAppMetadata` was calling `ParameterStore.getParameter` 3 times per app (~2250 async calls for 750 apps). Now fetched once via `fetchParameterStoreValues()` and passed in, fetched in parallel with metrics.
- **Use Maps for O(1) lookups**: Replaced `Array.find()` with `Map.get()` for metrics data and app stats throughout the endpoint and ranking logic.
- **Pre-compute ranking scores**: `rankApps` sort comparator was doing `.find()` on every comparison (O(n² log n)). Now pre-computes scores into a Map before sorting (O(n log n)).
- **Simplify category ranking**: Replaced O(n²) `.filter()` + `.indexOf()` pattern with a single O(n) counter pass.

## Checklist

### DIFF
--- a/web/api/helpers/app-store.ts
+++ b/web/api/helpers/app-store.ts
@@ -3,23 +3,57 @@ import { NATIVE_MAPPED_APP_ID, NativeAppToAppIdMapping } from "@/lib/constants";
 import { generateExternalNullifier } from "@/lib/hashing";
 import {
   AppStatsItem,
-  AppStatsReturnType,
   AppStoreFormattedFields,
   AppStoreMetadataDescription,
   AppStoreMetadataFields,
 } from "@/lib/types";
 import { getCDNImageUrl, getLogoImgCDNUrl, tryParseJSON } from "@/lib/utils";
 
+export interface ParameterStoreValues {
+  whitelistedAppsPermit2: string[];
+  whitelistedAppsContracts: string[];
+  implicitCredentialsApps: string[];
+}
+
+export const fetchParameterStoreValues =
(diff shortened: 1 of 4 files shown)


web/api/helpers/app-store.ts
```diff
@@ -3,23 +3,57 @@ import { NATIVE_MAPPED_APP_ID, NativeAppToAppIdMapping } from "@/lib/constants";
...
+      whitelistedAppsPermit2: whitelistedAppsPermit2 ?? [],
+      whitelistedAppsContracts: whitelistedAppsContracts ?? [],
+      implicitCredentialsApps: implicitCredentialsApps ?? [],
+    };
+  };
+
 export const formatAppMetadata = async (
   appData: AppStoreMetadataFields,
-  appStats: AppStatsReturnType,
+  appStatsMap: Map<string, AppStatsItem>,
```
andy-t-wang, line 48: Does it make sense for this endpoint to still be async? I don't think we await anything anymore

andy-t-wang (approved):

PR #1861, opened 2026-04-09
fix: optimize notification_log_wallet_address insert performance

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Optimizes the `notification_log_wallet_address` bulk insert query that was consuming a large percentage of DB query runtime. Two changes:

1. **Batch inserts into groups of 100** — Hasura wraps bulk inserts in a CTE with `RETURNING *` and constraint checks. Large batches (hundreds of wallet addresses) produce oversized queries with significant planning overhead. Batching keeps each query small and efficient.

2. **Fix v2 schema duplicating wallet address logs per localisation** — Previously, for v2 schema notifications, `logNotification` was called once per localisation entry, each time inserting the *same* full set of wallet addresses. This multiplied insert volume by the number of localisations. Now wallet addresses are logged once regardless of schema version.

Also adds error handling for insert failures that were previously silently swallowed (fire-and-forget without try/catch).

## Checklist

### DIFF
--- a/web/api/v2/minikit/send-notification/index.ts
+++ b/web/api/v2/minikit/send-notification/index.ts
@@ -93,10 +93,12 @@ export const logNotification = async (
         })),
       });
     } catch (error) {
-      logger.error(
-        "NotificationLog - failed to insert wallet address batch",
-        { app_id, notificationLogId, batchIndex: i, error },
-      );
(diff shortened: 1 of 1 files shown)


Gr1dlock:
@codex review

andy-t-wang (approved):

PR #1862, opened 2026-04-14
fix: remove one-time nullifier block in v4 uniqueness verify

Takaros999:
## Summary
- Removes the one-time verification restriction for v4 uniqueness proofs in production
- Nullifier reuse is now always allowed — proofs can be verified multiple times
- Supports auditable log workflows and HITL use cases that need to re-verify the same proof

## What changed
- Removed the `allowNullifierReuse` conditional in `web/api/v4/verify/uniqueness-proof/handler.ts`
- Existing nullifier now always returns 200 success instead of 400 `max_verifications_reached`
- Race condition on insert also returns 200 success unconditionally

## Test plan
- [ ] Verify a v4 uniqueness proof in production environment
- [ ] Re-verify the same proof — should return 200 with `nullifier_reused: true`
- [ ] Verify staging and v3.0 paths still work as before

### DIFF
--- a/web/api/v4/verify/uniqueness-proof/handler.ts
+++ b/web/api/v4/verify/uniqueness-proof/handler.ts
@@ -211,58 +211,41 @@ export async function handleUniquenessProofVerification(
   });
 
   const existingNullifier = checkNullifierResult.nullifier_v4[0];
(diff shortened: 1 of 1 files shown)


Gr1dlock:
@codex review

Gr1dlock (approved):

PR #1863, opened 2026-04-20
fix: invalidate rp_status cache after signer key rotation

Takaros999:
## Summary

- After a signer key rotation, the `ClaimRotationSlot` mutation sets DB status to `pending`, but the `rp_status` Redis cache still holds a stale `"registered"` response (1h TTL). The rp-status polling endpoint serves from cache and never queries on-chain, so the DB is never synced back to `registered`. This blocks all subsequent rotation attempts with `rotation_in_progress`.
- Adds the same Redis cache invalidation that `switch-to-self-managed` already performs, so the next rp-status poll does a real on-chain check and syncs the DB.
- Existing stuck customers will self-heal within ~1h when their cache expires naturally.


Takaros999:
@codex review

andy-t-wang (approved):

PR #1869, opened 2026-04-28
fix: persist staging rotation outcome so retries surface

Gr1dlock:
## PR Type

- [x] Bug Fix

## Description

Signer key rotation duplicates to the staging contract best-effort, but the handler never wrote anything to `staging_status` / `staging_operation_hash`. Combined with the fact that signer rotation does not flip `initialized`/`active` on-chain, `rp-status` always returns `staging_status: "registered"` regardless of the staging tx result — so a failed staging rotation went silent (production gets the new signer, staging keeps the old one, no UI retry button surfaces).

This change captures the staging tx outcome and persists it via a new `UpdateStagingRotationResult` mutation: `pending` + new hash on success, `failed` (hash null) on submission error. The DB write is wrapped in best-effort try/catch so a Hasura blip can't fail an otherwise-successful production rotation. New unit tests cover the four meaningful branches (success, failure, no staging config, DB write throws) plus the non-production env skip.

## Checklist


Gr1dlock:
@codex review

Gr1dlock:
@cursor review

bin-umar (approved):

PR #1871, opened 2026-04-28
fix: refresh service JWT before notification audit log writes

Gr1dlock:
## PR Type

- [x] Bug Fix

## Description

`logNotification` in `/api/v2/minikit/send-notification` was reusing the `serviceClient` minted at the start of the request handler, but the Hasura service JWT only lives 1 minute and the upstream `fetchWithRetry` can run longer with backoff. By the time the fire-and-forget `CreateNotificationLog` mutation ran, Hasura rejected it with `JWTExpired`, the audit row plus per-wallet rows were dropped, and the missing try/catch around the create call turned the failure into an unhandled rejection (the noisy raw `graphql-request` log we were seeing). `logNotification` now mints its own fresh service client, the create call is wrapped in try/catch with a structured `logger.error`, and the call site has a `.catch` backstop.

## Checklist

### DIFF
--- a/web/api/v2/minikit/send-notification/index.ts
+++ b/web/api/v2/minikit/send-notification/index.ts
@@ -6,7 +6,6 @@ import { logger } from "@/lib/logger";
 import { fetchWithRetry } from "@/lib/utils";
 import { createSignedFetcher } from "aws-sigv4-fetch";
 import { createHash } from "crypto";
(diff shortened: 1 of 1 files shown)


Gr1dlock:
@codex review

bin-umar (approved):

PR #1873, opened 2026-04-29
fix: temporary endpoint to backfill drifted staging signers

Takaros999:
Temporary API-key-gated endpoint to backfill 191 RPs with drifted
  staging signers after the RpRegistry impl upgrade (worldcoin/world-id-protocol#701). Queries DB for all managed RPs, compares on-chain staging signer vs DB signer, submits updateRp for any that differ. Restricted
  to two team IDs. Depends on worldcoin/world-id-deploy#515. DELETE after backfill.


Takaros999:
@codex review

Takaros999:
@codex review

andy-t-wang (approved):

PR #1881, opened 2026-04-29
feat: verify integrity bundle

Takaros999:
This PR ...
- adds optional integrity_bundle verification to the v4 verify endpoint before proof verification
- verifies AG JWTs, Redis-caches JWKS by kid, and checks platform-specific device signatures
- adds request schema support, env defaults, and focused tests for Android, iOS, cache reuse, and failure cases

Validation:
- pnpm jest tests/api/v4/integrity-bundle.test.ts tests/api/v4/request-schema.test.ts --runInBand
- pnpm typecheck
- pnpm lint
- pnpm install --frozen-lockfile --ignore-scripts

**Note**: This is an experimental feature.


Takaros999:
@codex review

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,630 @@
...
+  );
+  const authenticatorData = toBytes(
+    (assertion as { authenticatorData?: unknown }).authenticatorData,
+    "authenticator_data",
+  );
+
+  const nonce = sha256(Buffer.concat([authenticatorData, params.digest]));
+  const signedDigest = sha256(nonce);
+
+  const isValid = p256.verify(signature, signedDigest, params.devicePublicKey, {
```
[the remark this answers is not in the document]
Takaros999: This is correct as written. Apple App Attest signs the assertion with ECDSA-P256 over `nonce = SHA256(authenticatorData || clientDataHash)`. ECDSA's primitive always hashes its input internally — the value the curve operates on is `SHA256(nonce)`, not `nonce`.

In `@noble/curves` `p256.verify`, `prehash: true` hashes the message internally; `prehash: false` treats the input as the already-computed digest. These two forms are equivalent:

```ts
p256.verify(sig, nonce, pubKey, { prehash: true })
p256.verify(sig, sha256(nonce), pubKey, { prehash: false })
```

The code uses form (B): compute `nonce`, then `signedDigest = sha256(nonce)`, then verify with `prehash: false`. The second SHA256 is the standard ECDSA pre-hash, not an extra application-level hash. Standards-compliant `generateAssertion` outputs verify correctly — no client-side change required.

web/api/v4/verify/index.ts
```diff
@@ -137,6 +141,30 @@ export async function POST(
     const rpId = rpRegistration.rp_id;
     const appId = rpRegistration.app_id;
 
+    if (parsedParams.integrity_bundle) {
```
NnnOooPppEee, line 144: Are we fine to allow the request to go through if the RP does not send the bundle or should we fail if bundle is not present?

web/api/v4/verify/request-schema.ts
```diff
@@ -131,6 +131,10 @@ export const schema = yup
     // Optional environment to specify which verifier contract to use
     environment: yup.string().oneOf(["production", "staging"]).optional(),
 
+    // Optional World App integrity attestation bundle. When present, it is
+    // verified before proof verification.
+    integrity_bundle: yup.string().strict().optional(),
```
NnnOooPppEee: NIT, I would add a .max(8192) or something close to the double of the expected size just to have a bit more security in depth.

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,630 @@
...
+  return {
+    version,
+    signatureFormat,
+    timestamp,
+    signatureHex,
+    jwt,
+  };
+}
+
+function validateTimestamp(timestamp: number) {
```
NnnOooPppEee, line 172: The signup service https://github.com/worldcoin/signup-service/blob/main/shared/authentication/device_signature_middleware.go#L287-L292 does  SHA256(timestamp_be || payload) so that the freshness check is bound to the signature. But here at line L215-L241 we drops the timestamp from the digest on both v3 and v4 paths, so validateTimestamp can be modified by a malicious actor.

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,630 @@
...
+        });
+      }
+    }),
+  );
+}
+
+async function fetchJwks(jwksUrl: string) {
+  let response: Response;
+  try {
+    response = await fetch(jwksUrl, { cache: "no-store" });
```
NnnOooPppEee: nit, I would add a small AbortSignal of like 4 secs, otherwise default timeout can be 5min, in reality I think one of the alb would timeout after a minute but still worth timing out before that.

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,630 @@
...
+  digest: Uint8Array;
+  signatureHex: string;
+}) {
+  const signature = Buffer.from(params.signatureHex, "hex");
+  const isValid = p256.verify(
+    signature,
+    params.digest,
+    params.devicePublicKey,
+    {
+      lowS: false,
```
NnnOooPppEee, line 532: NIT, do we think we have case where we can't guarantee lowS to be true?

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,630 @@
...
+
+async function verifyJwtWithJwk(params: {
+  integrityJwt: string;
+  jwk: JWK;
+  rpId: string;
+}) {
+  const publicKey = await importJWK(params.jwk, "ES256");
+  const { payload } = await jwtVerify(params.integrityJwt, publicKey, {
+    algorithms: ["ES256"],
+    audience: params.rpId,
```
NnnOooPppEee: when `aud !== rpId`, jose's mismatch error gets coalesced into `invalid_integrity_token` at L464 will end up as a warn at L621. Could we emit a dedicated `logger.warn("integrity_aud_mismatch", { rp_id, jwt_aud, kid })` with `reason: "audience_mismatch"`?
We should have then an alert on this in datadog, so that we have a way to catch any wrong implementation in the mobile where we would use an integrity which does not belong to the correct rp which would be an important privacy issue. 
This should obviously not be the main way to detect this, but it's an additional mechanism.

web/api/v4/verify/index.ts
Takaros999: Confirmed this is intentionally optional for this PR. I left the existing behavior unchanged so requests without `integrity_bundle` still go through; making it mandatory would be a separate API/rollout decision.

web/api/v4/verify/request-schema.ts
Takaros999: Done. Added `.max(8192)` on `integrity_bundle` and a schema test covering oversized values.

web/api/v4/verify/integrity-bundle.ts
Takaros999: Done. The device signature now verifies `SHA256(timestamp_be || payload_digest)`, matching the signup-service pattern, and there is a regression test that mutates `t` after signing and expects `invalid_device_signature`.

web/api/v4/verify/integrity-bundle.ts
Takaros999: Done. JWKS fetches now use an `AbortController` with a 4s timeout and still normalize timeout/network failures into `jwks_fetch_failed`.

web/api/v4/verify/integrity-bundle.ts
Takaros999: Checked the local mobile implementations before changing this. iOS delegates to `DCAppAttestService.generateAssertion`, and Android signs with `NONEwithECDSA` via Android Keystore; neither path appears to canonicalize or guarantee low-S in app code. I kept `lowS: false` for compatibility.

web/api/v4/verify/integrity-bundle.ts
Takaros999: Done. JWT verification now validates signature/issuer first, then checks audience explicitly so mismatches return `audience_mismatch` and emit `logger.warn("integrity_aud_mismatch", { rp_id, jwt_aud, kid, reason })`. Added a test for `aud !== rpId` as well.

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,671 @@
...
+
+function validateAudience(params: {
+  kid: string;
+  payload: IntegrityTokenClaims;
+  rpId: string;
+}) {
+  const aud = params.payload.aud;
+  const matches =
+    typeof aud === "string"
+      ? aud === params.rpId
```
chipp: [P1] at the moment we should support `developer.worldcoin.org` and `staging-developer.worldcoin.org`

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -0,0 +1,671 @@
...
+
+  return publicKey;
+}
+
+async function verifyJwtWithJwk(params: { integrityJwt: string; jwk: JWK }) {
+  const publicKey = await importJWK(params.jwk, "ES256");
+  const { payload } = await jwtVerify(params.integrityJwt, publicKey, {
+    algorithms: ["ES256"],
+    issuer: getIntegrityExpectedIssuer(),
+  });
```
chipp, line 439: nit: we can add `audience` verification here
and I would suggest adding `requiredClaims` here too
`requiredClaims: ['cnf', 'exp', 'platform', 'pass']`

chipp (approved):

chipp (changes requested):
please check audience comment

web/api/v4/verify/integrity-bundle.ts
chipp: cc @NnnOooPppEee

chipp (approved):

web/api/v4/verify/integrity-bundle.ts
```diff
@@ -28,6 +28,7 @@ const JWKS_FETCH_TIMEOUT_MS = 4_000;
 const DEFAULT_INTEGRITY_JWKS_URL =
   "https://attestation.worldcoin.org/.well-known/jwks.json";
 const DEFAULT_INTEGRITY_ISSUER = "attestation.worldcoin.org";
+const DEFAULT_INTEGRITY_AUDIENCE = "developer.worldcoin.org";
```
NnnOooPppEee, line 32: Please add a comment explaining that this should never be used for live implementation and this has to be removed before enabling any RP implementation.

NnnOooPppEee: @chipp, I want to make clear that this cannot end up being used by any RP, so please put in place the necessary safeguard so that no one does uses this by mistake or forget about it. That would be very bad for privacy.

PR #1897, opened 2026-05-05
fix: avoid uniqueness violation in /api/create-team for existing users

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Investigated a recurring 500 on `POST /api/create-team` ([DD trace](https://app.datadoghq.com/apm/trace/69f9ba470000000041bc855291516d82)) — Hasura returns a uniqueness violation on `InsertUser` because the handler trusts a body-supplied `hasUser` flag derived from `session.user.hasura`, which can be stale and cause the insert to collide with an existing record's `auth0Id`/`email` (24 occurrences in the last 30 days).

The handler now performs a defensive `GetUserByAuth0Id` lookup at the top: if Hasura already has the user, `effectiveHasUser` becomes true and we skip Ironclad acceptance, the `InsertUser` mutation, and the `signup_success` PostHog event, reusing the existing user's id for the membership. All GraphQL `logger.error` calls in `create-team` and `login-callback` now also pass `graphqlResponse: error?.response` (only the fingerprint was previously persisted, so the constraint name was lost), and `login-callback` warns when no Hasura user is found so we can correlate future occurrences.

## Checklist

### DIFF
--- a/web/api/create-team/graphql/get-user-by-auth0id.generated.ts
+++ b/web/api/create-team/graphql/get-user-by-auth0id.generated.ts
@@ -1,0 +1,62 @@
+/* eslint-disable import/no-relative-parent-imports -- auto generated file */
+import * as Types from "@/graphql/graphql";
+
+import { GraphQLClient, RequestOptions } from "graphql-request";
+import gql from "graphql-tag";
+type GraphQLClientRequestHeaders = RequestOptions["requestHeaders"];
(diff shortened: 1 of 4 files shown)


Gr1dlock:
@codex review

Takaros999 (approved):

PR #1898, opened 2026-05-06
fix: prevent timeout in delete-expired-notification-logs cron

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

The `/api/delete-expired-notification-logs` cron handler has been failing recurrently (8 times in the past 24h, ongoing since Aug 2025 — Datadog issue `ccb17576-81e0-11f0-b499-da7ad0900002`). It tries to drain the entire backlog in one request with 10s sleeps between batches, but Hasura's cron trigger has a 60s `timeout_seconds`, so the connection gets aborted after ~6 batches and the next GraphQL call throws — caught and re-thrown as a generic `"Failed to delete expired notification logs"` 500 with no underlying error logged. This PR adds a `TIME_BUDGET_MS=45000` deadline check that exits the loop cleanly before Hasura times out (the next 15-minute cron tick picks up the remainder), and changes the catch block to log the underlying `err` so future failures are diagnosable instead of opaque.

Reference trace: https://app.datadoghq.com/apm/trace/69fb3b9b000000001ff2bf88ab3adcf3

## Checklist

### DIFF
--- a/web/api/delete-expired-notification-logs/index.ts
+++ b/web/api/delete-expired-notification-logs/index.ts
@@ -1,12 +1,16 @@
 import { errorNotAllowed, errorResponse } from "@/api/helpers/errors";
 import { getAPIServiceGraphqlClient } from "@/api/helpers/graphql";
 import { protectInternalEndpoint } from "@/api/helpers/utils";
+import { logger } from "@/lib/logger";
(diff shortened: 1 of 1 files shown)


Takaros999 (approved):
we could also bump the timeout

PR #1901, opened 2026-05-08
fix: disable notifications for external apps

Gr1dlock:
## PR Type

- [x] Bug Fix

## Description

External apps were producing repeated "App Directory Metadata call failed" / "Invalid category" 404s because app-backend's send-notification flow fetches app metadata from `/api/v2/public/app/[app_id]`, which 404s when the app's category is "external" and `show_external` isn't passed. Rather than punching a hole through that endpoint, this PR blocks the feature for external apps at the developer portal: `/api/v2/minikit/send-notification` now returns a clear \`external_app_not_allowed\` 400 (checking both \`app_mode === "external"\` and \`category === "External"\`, matching the existing filter pattern in the apps list endpoint), and the Notifications UI page renders an "unavailable" notice instead of the form. New unit tests cover both rejection paths.

## Checklist

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -22,8 +22,11 @@ WORKDIR /app
 # Install working version of corepack
 RUN npm install -g corepack@0.31.0 && corepack --version
 
+# This stage has no project package.json, so corepack would otherwise fall back
+# to the latest pnpm (incompatible with the Node 20 base). Pin to the same
+# version declared in the project's packageManager field.
 # TODO: Can be further optimized to remove next peer dependency
-RUN corepack enable pnpm && pnpm i next-logger@5.0.0 pino@9.2.0 dd-trace@5.12.0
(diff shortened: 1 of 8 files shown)


web/Dockerfile
```diff
@@ -22,8 +22,11 @@ WORKDIR /app
 # Install working version of corepack
 RUN npm install -g corepack@0.31.0 && corepack --version
 
+# This stage has no project package.json, so corepack would otherwise fall back
+# to the latest pnpm (incompatible with the Node 20 base). Pin to the same
+# version declared in the project's packageManager field.
 # TODO: Can be further optimized to remove next peer dependency
-RUN corepack enable pnpm && pnpm i next-logger@5.0.0 pino@9.2.0 dd-trace@5.12.0
+RUN corepack enable pnpm && corepack prepare pnpm@9.15.4 --activate && pnpm i next-logger@5.0.0 pino@9.2.0 dd-trace@5.12.0
```
Takaros999, line 29: hmm did this break just now?

Takaros999 (approved):

web/Dockerfile
Gr1dlock: yup

PR #1905, opened 2026-05-22
fix: display configuration images at their natural aspect ratio

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Showcase, meta-tag, content-card, and uploaded-image previews on the app configuration page were rendered into a fixed `h-[200px] w-full` box with `object-cover`, which cropped every image whose source aspect ratio didn't match the container's runtime aspect. Containers now derive `aspectRatio` from each field's `imageConstraints` (1080×1080 / 1200×600 / 345×240), and the images use `object-contain` so they're never clipped. Also fixed the Next `Image` width/height on the content-card preview which were incorrectly set to 1200×600 (meta-tag's dimensions) instead of 345×240.

image
image
image
image

## Checklist


andy-t-wang (approved):

andy-t-wang:
Add screenshot too pls

PR #1913, opened 2026-05-26
fix: reload on stale Server Action errors via global-error boundary

Gr1dlock:
## Summary

- Adds `web/app/global-error.tsx`. The app previously had no client-side error boundary at any level.
- Detects the three known stale-bundle error variants and triggers a one-time `window.location.reload()` so clients holding an older bundle recover transparently after a deploy.
- Guards against reload loops with a 30s `sessionStorage` cooldown.

## Why

After a deploy, clients with an older JS bundle call Server Action IDs the new server no longer recognises and Next.js throws:

- `Failed to find Server Action ...` — ~3,009/wk ([`07359352`](https://app.datadoghq.com/error-tracking/issue/07359352-1acb-11f1-b6aa-da7ad0900002))
- `Failed to find Server Action ... Cannot read properties of undefined (reading 'workers')` — ~951/wk ([`ee11556a`](https://app.datadoghq.com/error-tracking/issue/ee11556a-2bb8-11f1-bf12-da7ad0900002))
- `TypeError: Failed to parse body as FormData.` — ~455/wk ([`355d359a`](https://app.datadoghq.com/error-tracking/issue/355d359a-d8ab-11ef-910b-da7ad0900002))

All three flow through the same Next.js Server Action body-parsing path. Documented behavior: https://nextjs.org/docs/messages/failed-to-find-server-action.

## Behavior

- Matches on `error.message?.includes(...)` for all three variants.
- On match: shows minimal "Reloading…" UI and calls `window.location.reload()` inside `useEffect`.
- On non-match: neutral fallback with a "Try again" button wired to Next's `reset()` — strict improvement over no boundary.
- `sessionStorage` key `sa_reload_at` stores last-reload timestamp; if a stale-action error fires again within 30s, we skip the auto-reload and show the manual fallback. Private-mode safe via `try/catch`.

## Test plan

- [ ] Locally trigger a Server Action ID change between builds and verify a stale client reloads exactly once
- [ ] Force the cooldown: throw the matching error twice within 30s, confirm the second occurrence shows the manual fallback (no infinite reload)
- [ ] Verify unrelated runtime errors still render the neutral fallback (and that "Try again" calls `reset()`)
- [ ] After merge, monitor the three Datadog issues and expect a drop in combined firing rate

### DIFF
--- a/web/app/global-error.tsx
+++ b/web/app/global-error.tsx
@@ -47,20 +47,28 @@ const shouldAutoReload = () => {
     }
     return Date.now() - lastAt > RELOAD_COOLDOWN_MS;
   } catch {
-    // sessionStorage can throw in private modes / disabled storage. If we
-    // can't read it we err on the side of reloading once.
-    return true;
+    // sessionStorage is unavailable (private mode / disabled storage).
+    // Without a persisted marker we can't enforce the cooldown across
+    // the reload, so a recurring stale-action error would trigger a
+    // reload on every render. Refuse to auto-reload — the manual
+    // fallback UI handles it.
+    return false;
   }
 };
 
-const markReload = () => {
+// Persist the reload marker. Returns true only if the write succeeded;
+// callers MUST NOT trigger window.location.reload() unless this returns
+// true, otherwise a recurring error after the reload becomes a tight
+// loop with no enforceable cooldown.
+const markReload = (): boolean => {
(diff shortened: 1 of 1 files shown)


Takaros999:
@codex review

Takaros999 (approved):

PR #1918, opened 2026-05-27
fix: retry shared GraphQLClient on transport errors (query-only)

Gr1dlock:
## Summary

- Adds a retry-on-transport-error wrapper to the three shared `GraphQLClient` factories in `web/api/helpers/graphql.ts` (used at ~370 call sites).
- Each attempt also gets an explicit 15s timeout via `AbortSignal.timeout`.
- Closes ~35/wk Datadog noise from [`71f93256`](https://app.datadoghq.com/error-tracking/issue/71f93256-d8d3-11ef-a029-da7ad0900002) (`TypeError: fetch failed`).

## Why

The shared client previously had **zero retry and no explicit timeout**. Any transient transport failure to Hasura (`ECONNRESET`, socket hang up, undici `UND_ERR_*`) hit users immediately. Triage write-up: `.context/outbound-http-triage.md` in the #6 worktree.

## Load-bearing safety: retry never fires for mutations

We parse the GraphQL document with the `graphql` package's `parse()` AST — the same call `graphql-request` itself uses in `resolveRequestDocument`. The helper `isRetryableOperation(body)` returns `true` only if:

- the body is a non-empty `string`
- it parses as a JSON object (rejecting `Array` bodies, i.e. batch requests)
- the resulting `.query` is a string that parses as a valid GraphQL document
- **every** `OperationDefinition` in the doc has `def.operation === "query"` or `"subscription"`

If we can't positively identify a doc as query-only, we fall through to a single attempt. No string heuristics (`/^\s*mutation\b/`) — comments and lexer subtleties can fool those.

## Other safety properties

- **Transport-error detection** (`isTransportError`) checks `name`, `code`, `cause.code`, `cause.name` against a whitelist (`ECONNRESET`, `ECONNREFUSED`, `ETIMEDOUT`, `EAI_AGAIN`, `ENOTFOUND`, `EPIPE`, `UND_ERR_SOCKET`, `UND_ERR_CONNECT_TIMEOUT`, `UND_ERR_HEADERS_TIMEOUT`, `UND_ERR_BODY_TIMEOUT`) plus message substrings as a last-resort match. A 4xx/5xx response is not an error — `fetch` resolves with the `Response` and the retry loop returns it immediately.
- **Bounded retry**: `[200ms, 500ms]` backoff → max 3 attempts total for queries; 1 attempt for everything else.
- **Per-attempt fresh timeout** so a slow first attempt doesn't shorten the retry budget. The caller-supplied `signal` is composed via `AbortSignal.any` (with a manual fallback for older Node).

## Tests

`web/tests/api/helpers/graphql.test.ts` — 13 tests covering the retry helper directly without mocking the SDK:
- query success on first attempt
- query retry-and-succeed
- query gives up after 3 attempts
- mutation never retries (single op, multi-op, anonymous mutation)
- batch request never retries
- 5xx response never retries (no transport error)
- non-transport error never retries
- caller-supplied signal pass-through
- all `isRetryableOperation` branches (non-string body, unparseable JSON, missing `.query`, fragment-only doc, etc.)

## Test plan

- [ ] `cd web && npx jest tests/api/helpers/graphql.test.ts` — all green (already verified locally)
- [ ] Locally point the app at a flaky/dropped Hasura endpoint and confirm a query retries while a mutation does not
- [ ] After merge, watch Datadog issue [`71f93256`](https://app.datadoghq.com/error-tracking/issue/71f93256-d8d3-11ef-a029-da7ad0900002) — expect firing rate to fall toward zero (the Hasura component will collapse; other `fetch failed` callers stay until their own fixes)
- [ ] Confirm overall p95 latency on `/api/hasura/*` doesn't regress materially (added retries add tail latency on transient failures)

### DIFF
--- a/web/api/helpers/graphql.ts
+++ b/web/api/helpers/graphql.ts
@@ -203,8 +203,12 @@ export const graphqlFetchWithRetry: typeof fetch = async (input, init) => {
       lastError = err;
       const isLast = attempt === maxAttempts;
       const transport = isTransportError(err);
+      // If the caller asked to cancel (route timeout, client disconnect,
+      // upstream short-circuit), honour it immediately rather than treating
+      // the resulting AbortError as a retryable transport failure.
+      const callerAborted = callerSignal?.aborted ?? false;
 
-      if (isLast || !transport) {
+      if (isLast || !transport || callerAborted) {
         throw err;
       }
 
--- a/web/api/helpers/graphql.ts
+++ b/web/api/helpers/graphql.ts
@@ -5,7 +5,228 @@ import {
   generateReviewerJWT,
   generateServiceJWT,
 } from "@/api/helpers/jwts";
+import { logger } from "@/lib/logger";
+import { parse } from "graphql";
 import { GraphQLClient } from "graphql-request";
+
+/** Total per-request timeout, including any retry attempts. */
+const REQUEST_TIMEOUT_MS = 15_000;
+
+/** Retry plan for idempotent (query) operations only. Backoffs in ms, applied
+ *  between attempts. Length = max retries beyond the first attempt. */
+const QUERY_RETRY_BACKOFFS_MS = [200, 500];
+
+/**
+ * Inspect a serialized GraphQL request body (the JSON `graphql-request` puts on
+ * the wire) and decide whether the operation is safe to retry.
+ *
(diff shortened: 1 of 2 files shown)


Takaros999:
@codex review

Takaros999 (approved):

PR #1914, opened 2026-05-26
chore: log unauthorized app access at warn level

Gr1dlock:
## Summary

- Downgrades `logLevel: \"error\"` to `\"warn\"` for the two `errorFormAction` calls that fire when a user navigates to a transaction page for an app they don't own.
- Closes ~66/wk Datadog noise from issue [`be8c106c`](https://app.datadoghq.com/error-tracking/issue/be8c106c-49c4-11f1-966c-da7ad0900002).

## Why

The team-membership checks in `getAccumulativePaymentsData` and `getAccumulativeTransactionsData` (file `web/scenes/Portal/Teams/TeamId/Apps/AppId/MiniApp/Transactions/page/server/getAccumulativeTransactionData.ts`) were added in #1860 as a CVE-grade fix — before that, any authenticated user could query payment data for any `appId`.

The rejections themselves are legitimate (users following stale links, support staff without team membership, crawlers hitting authenticated routes). They are not errors that warrant pages or investigation — they are the authz check doing its job. Logging them at `warn` preserves the audit trail without polluting error tracking.

The other `logLevel: \"error\"` calls in the same file (missing `NEXT_SERVER_INTERNAL_PAYMENTS_ENDPOINT` env var, generic fetch failure) are unchanged — those are real errors.

## Test plan

- [ ] After merge, watch Datadog issue [`be8c106c`](https://app.datadoghq.com/error-tracking/issue/be8c106c-49c4-11f1-966c-da7ad0900002) — expect firing rate to fall to zero (the entries will reappear as `warn`-level logs)
- [ ] Confirm `warn`-level logs are still emitted by checking Datadog Logs for the message "User is not authorized to view this app's data"
- [ ] Confirm the user-facing behavior is unchanged (still returns a `FormActionResult` rejection)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1926, opened 2026-06-02
chore: log "RP registration not found" at warn across RP action handlers

Gr1dlock:
## Summary
- Adds `logLevel: "warn"` to the `errorHasuraQuery` call at three RP-action handler guards that throw "RP registration not found for this app.":
  - `web/api/hasura/rotate-signer-key/index.ts:96`
  - `web/api/hasura/switch-to-self-managed/index.ts:112`
  - `web/api/hasura/toggle-rp-active/index.ts:93`
- Stops ~36/3d legitimate rejections ([5016f8f2](https://app.datadoghq.com/error-tracking/issue/5016f8f2-1fb9-11f1-8570-da7ad0900002), [cc67a924](https://app.datadoghq.com/error-tracking/issue/cc67a924-0745-11f1-9554-da7ad0900002)) from polluting Error Tracking.

## Why this is noise, not a bug
The guard (`if (!rp_registration || rp_registration.length === 0)`) fires only when the app has **no** RP-registration row at all — a user acting on an app that never registered an RP. The row is created by a separate `register-rp` flow, so there's no read-after-write race. All sites return HTTP **400** `rp_not_registered`, `handling: handled`, `is_crash: false`. Same class as PR #1914.

`errorHasuraQuery` defaults `logLevel` to `"error"`; these handlers omitted it, so each rejection emitted `logger.error`. Downgrading to `warn` keeps the audit log but drops it from Error Tracking. The shared `rp-registration-flows.ts` path already returns a tagged result without logging, so it needed no change.

## Follow-up (out of scope)
The co-occurring "Cannot toggle RP. Current status is 'pending'…" and "User does not have permission to toggle RP status." are the same class of legitimate 400 logged at error level — could be downgraded too in a follow-up.

## Test plan
- [ ] tsc + format (done locally, clean)
- [ ] Watch the two issues — expect them to stop appearing in Error Tracking (continue as warn logs)

### DIFF
--- a/web/api/hasura/rotate-signer-key/index.ts
+++ b/web/api/hasura/rotate-signer-key/index.ts
@@ -99,6 +99,7 @@ export const POST = async (req: NextRequest) => {
       detail: "RP registration not found for this app.",
       code: "rp_not_registered",
       app_id,
+      logLevel: "warn",
     });
   }
   const teamId = rp_registration[0].app.team_id;
--- a/web/api/hasura/switch-to-self-managed/index.ts
+++ b/web/api/hasura/switch-to-self-managed/index.ts
@@ -115,6 +115,7 @@ export const POST = async (req: NextRequest) => {
       detail: "RP registration not found for this app.",
       code: "rp_not_registered",
       app_id,
+      logLevel: "warn",
     });
   }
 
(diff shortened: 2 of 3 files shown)


Gr1dlock:
@codex review

bin-umar (approved):

PR #1943, opened 2026-06-09
chore(auth): migrate @auth0/nextjs-auth0 v3 → v4

Gr1dlock:
## Context

PR 1 of 2 for the Next.js 15 → 16 upgrade. `@auth0/nextjs-auth0@3.8.0` does not support Next 16 (peer range stops at `^15.2.3`) and relies on the removed `/edge` subpath, so it's the main blocker. This upgrades it to **v4.22.0** (peer range includes `^16.0.10`) **on Next.js 15**, so the rewrite lands and bakes before the framework bump (PR 2).

## What changed

- **`web/lib/auth0.ts`** (new) — configured `Auth0Client`. **All** auth routes (`login`/`logout`/`callback`/`profile`) are pinned under `/api/auth/*` (v4 defaults to `/auth/*`) so the Auth0 tenant's Allowed Callback/Logout URLs and the `lib/urls.ts` helpers are unchanged.
- **Dual-host redirects** — `appBaseUrl` is set to the allowed-host list (the `worldcoin.org` + `world.org` variants, via `lib/app-base-url.ts`) in **allow-list mode**, so v4 builds `redirect_uri`/logout URLs from the host the request came in on (matching v3's `getAppUrlFromRequest`) instead of a single static `APP_BASE_URL`.
- **Client profile route** — `useUser()` resolves its fetch URL from the build-time `NEXT_PUBLIC_PROFILE_ROUTE`; it's inlined to `/api/auth/profile` in every build path (Dockerfile `ARG` default, `.env` files, CI build + e2e jobs) so the client and the mounted server route agree.
- **`web/middleware.ts`** — `withMiddlewareAuthRequired` → `auth0.middleware()` + explicit `auth0.getSession()` gating. **CSP nonce + owner/admin role restrictions preserved verbatim.** Matcher: `/api/auth/*` + protected routes.
- **Route restructure** — removed the v3 `[auth0]` catch-all (v4 auto-mounts auth routes via middleware); added standalone `app/api/auth/login-callback` and `delete-account` route handlers.
- **API handlers** — `auth0.getSession()` / `auth0.updateSession()`; dropped the redundant `withApiAuthRequired` wrapper (each handler already guards the session). `user.hasura` session injection unchanged.
- **~16 page/layout/server files** — `getSession()` → `auth0.getSession()`.
- **Client** — `UserProvider` → `Auth0Provider`; `checkSession()` → `invalidate()`.
[...]

PR #1949, opened 2026-06-16
fix(auth): v4 staging regressions — server-only client bundle + render loop

Gr1dlock:
Follow-up to #1943 (Auth0 v3 → v4), found on **staging**: the browser console logs

> ⚠ WARNING: Not all required options were provided when creating an instance of Auth0Client … Missing: domain, clientId, secret, clientAuthentication

### Root cause
`web/lib/auth0.ts` constructs `new Auth0Client()` at **module scope**, and that module was being pulled into the **client bundle** and executed in the browser — where `AUTH0_DOMAIN`/`AUTH0_CLIENT_ID`/`AUTH0_SECRET` are server-only and `undefined`, hence the warning.

The leak path: `scenes/Portal/layout/index.tsx` is a Server Component that imports `auth0`, but it **also exported `colorAtom`**, which several **client** components import (`LoggedUserNav`, `Header`, `Profile/page`, `UserInfo/Icon`, `TeamProfile/Image`). Importing the atom dragged the whole module — including the `Auth0Client` construction — into the client graph. (v3 imported a `getSession` *function*, so importing executed nothing; v4 constructs at module scope.)

### Fix
- Move `colorAtom` into its own client-safe module `scenes/Portal/layout/color-atom.ts` (no `auth0` import) and repoint all 5 consumers.
- Add `import "server-only"` to `web/lib/auth0.ts` so any future client import **fails the build** instead of silently shipping the server client to the browser. (This is what surfaced the second relative-path consumer during verification.)

### Verification
`tsc` (0 errors), `next build` passes with the `server-only` guard (proves `@/lib/auth0` is no longer in any client bundle), `jest` unit+api 445/445.

### Out of scope (also seen in the same console, unrelated to this fix)
- CORS error to `world-id-server.com` — the World ID verification subsystem (sign-in-with-world-id / IDKit), not portal auth.
- `429 Too Many Requests` burst on `…/actions` — client-side Next `<Link>` prefetch (`net::ERR_ABORTED`) hitting a staging rate limiter; not introduced by the auth migration. Happy to dig into either separately.

### DIFF
--- a/web/components/LoggedUserNav/index.tsx
+++ b/web/components/LoggedUserNav/index.tsx
@@ -10,7 +10,7 @@ import { DOCS_URL } from "@/lib/constants";
 import { Auth0SessionUser } from "@/lib/types";
 import { urls } from "@/lib/urls";
 import { checkUserPermissions } from "@/lib/utils";
-import { colorAtom } from "@/scenes/Portal/layout";
+import { colorAtom } from "@/scenes/Portal/layout/color-atom";
 import { useMeQuery } from "@/scenes/common/me-query/client";
 import { useUser } from "@auth0/nextjs-auth0/client";
 import { useAtom } from "jotai";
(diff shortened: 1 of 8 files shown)


bin-umar (approved):

Gr1dlock:
@codex review

PR #1953, opened 2026-06-17
fix(auth): use static APP_BASE_URL to stop v4 InvalidConfigurationError (500s on staging)

Gr1dlock:
Found while checking staging Datadog logs after the Auth0 v4 migration (#1943).

### Regression
Error Tracking shows a new `InvalidConfigurationError` in `middleware.js`, **first-seen exactly at the v4 deploy** (2026-06-15T18:27) and **still firing** (35× on staging):

> APP_BASE_URL is not configured as a static string, and the APP_BASE_URL configuration does not contain a match for the current request origin.

A sample event: a request on the **internal host** `http://developer.staging-internal.worldcoin.org/api/auth/login` → `startInteractiveLogin → resolveAppBaseUrl` **throws → 500**.

### Cause
`appBaseUrl` was passed as a `[worldcoin.org, world.org]` allow-list **array**. v4's `resolveAppBaseUrl` throws `InvalidConfigurationError` for any request origin **not** in the array — which includes internal/health-check hosts. v3's `getAppUrlFromRequest` fell back to `NEXT_PUBLIC_APP_URL` and never threw, so this is a v4-migration regression. (This is the real downside of the array that the earlier "stop passing an array" review flagged — not that arrays are unsupported, but that they throw on non-listed origins.)

### Fix
Leave `appBaseUrl` unset so the SDK reads the single static `APP_BASE_URL` env (the canonical host). It never throws and builds callback/logout URLs against the canonical host — matching v3's effective behaviour. Removes the now-unused `getAllowedAppBaseUrls`; `siblingOrigin` stays for the middleware CSP allow-list.

Trade-off: login initiated on the sibling `world.org` host would now redirect through the canonical host instead of staying on `world.org`. That path requires the host to be in the Auth0 tenant's Allowed Callback URLs anyway, and avoiding 500s on every non-canonical origin is the right call.

### Verification
`tsc` 0, `prettier`, `next build` (server-only guard intact).

### DIFF
--- a/web/lib/app-base-url.ts
+++ b/web/lib/app-base-url.ts
@@ -1,6 +1,6 @@
 // The portal is served from both the worldcoin.org and world.org variants of the
-// same hostname. These helpers centralise the host handling so the CSP allow-list
-// (middleware) and the Auth0 v4 redirect host allow-list (lib/auth0) agree.
+// same hostname. `siblingOrigin` centralises that mapping for the middleware CSP
+// allow-list (so assets/connections from either origin are allowed).
 
 /**
  * Given one origin, return the sibling origin on the other domain variant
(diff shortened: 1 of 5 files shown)


Gr1dlock:
@codex review

Gr1dlock:
@codex review

bin-umar (approved):

PR #1962, opened 2026-06-23
Remove MiniKiosk from TryAction settings panel.

kchaw2005:
## Issue
We were hardcoding the signature in field in legacyKioskRequest, making the authenticator unable to trust the request. No clear solution as to how to make MiniKiosk functional since we need an actual RP signing key.

## Fix
Updated TryAction with an enableKiosk flag to show only try-it-out helper CodeBlocks when the app is NOT a mini app and nothing if it is.

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/ActionId/Settings/TryAction/index.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/ActionId/Settings/TryAction/index.tsx
@@ -1,91 +1,32 @@
 "use client";
 
-import { Button } from "@/components/Button";
-import { CodeIcon } from "@/components/Icons/CodeIcon";
-import { QRIcon } from "@/components/Icons/QRIcon";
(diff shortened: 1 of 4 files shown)


web/scenes/Portal/Teams/TeamId/Apps/AppId/Actions/ActionId/Settings/TryAction/index.tsx
```diff
@@ -23,13 +23,25 @@ type TryActionProps = {
...
   is_v4_action: boolean;
+  enableKiosk?: boolean;
 };
 
 export const TryAction = (props: TryActionProps) => {
-  const { action, is_v4_action } = props;
+  const { action, is_v4_action, enableKiosk = true } = props;
+  const isMiniApp = action.app.app_metadata.some(
+    ({ app_mode }) => app_mode === "mini-app",
+  );
```
Takaros999: why do we hide the codeblock if it's not a mini-app? we'd like this present for all apps

Takaros999 (approved):

PR #1964, opened 2026-06-24
fix(auth): send an absolute returnTo on logout (fixes Auth0 /v2/logout 400)

Gr1dlock:
## Problem
Clicking **Log out** on staging lands on Auth0's "something went wrong" page:
```
GET …auth0.com/v2/logout?returnTo=%2Flogin&client_id=… → 400 (Bad Request)
```
`urls.logout()` sent a **relative** `returnTo=/login`. Auth0's `/v2/logout` requires `returnTo` to be an **absolute** URL in the tenant's Allowed Logout URLs. In v3 the SDK absolutised the relative path against the app URL; the **v4 SDK does not**, so the bare `/login` is rejected — a latent v4-migration bug that surfaced now.

## Fix
Send the absolute `${NEXT_PUBLIC_APP_URL}/login`. `NEXT_PUBLIC_APP_URL` is build-inlined and available on both server and client — required because `urls.logout()` is used in server `redirect()` calls *and* client links, so `window.location` isn't an option. `<origin>/login` is already an Allowed Logout URL (v3 logout returned there).

> Dual-domain note: a user on the sibling `world.org` host is returned to the canonical `<NEXT_PUBLIC_APP_URL>/login` (cross-domain) on logout — acceptable (both `/login` URLs are allow-listed). A fully per-host returnTo would require the SDK to absolutise relatives, which it no longer does.

## Verification
tsc 0 · ESLint clean · prettier clean · 477/477 unit+api tests. (`delete-account` integration test references `urls.logout()` dynamically, so it stays consistent; the login-callback `.includes("/api/auth/logout")` check still matches.)

### DIFF
--- a/web/components/LoggedUserNav/index.tsx
+++ b/web/components/LoggedUserNav/index.tsx
@@ -228,7 +228,17 @@ export const LoggedUserNav = () => {
           <Dropdown.ListSeparator />
 
           <Dropdown.ListItem asChild>
-            <a href={urls.logout()} className="text-system-error-600">
+            <a
+              href={urls.logout()}
+              onClick={(e) => {
+                // Log out against the host the user is actually on (worldcoin.org
+                // vs world.org) so they land on that host's /login and aren't
+                // bounced back in by a stale session cookie on the other domain.
+                e.preventDefault();
+                window.location.assign(urls.logout(window.location.origin));
+              }}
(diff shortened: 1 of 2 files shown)


Gr1dlock:
@codex review

bin-umar (approved):

PR #1965, opened 2026-06-24
fix(portal): cross-tenant authz bypass — stop trusting client session memberships (#3687935, #3733522)

Gr1dlock:
## Summary
Fixes the authenticated **cross-tenant authorization bypass** reported in HackerOne **#3687935** (and its duplicate **#3733522**).

`/api/update-session` allowlisted `memberships` and merged the **client-supplied** array into the sealed JWE session after only checking `body.user.id === session.user.hasura.id` — no DB verification. Any authenticated user could forge `memberships` to claim membership/ownership of arbitrary teams. Three data paths then trusted that claim (or did no auth at all):

| Path | Before | Impact |
|------|--------|--------|
| `getTransactionData` | gated on cookie `memberships` only, then signed-fetched the internal payments backend by `appId` | **cross-tenant read of another tenant's payment/transaction data** |
| `getAppMetricsData` | **no authorization at all** | any authenticated user reads any app's metrics by `appId` (clean IDOR, no forgery needed) |
| `AppIdLayout` | gated on cookie `memberships`, fetched app by PK with no team filter | cross-tenant app config metadata |

The DB-backed `getIsUserAllowedToReadApp` (in `web/lib/permissions`) was already used by the *safe* paths (`getAccumulative*`, all mutations) — this PR makes the broken paths consistent with it.

## Changes
- **`api/update-session`**: now re-derives **all** Hasura session claims server-side from the database via the verified `session.user.hasura.id` (new `FetchUserForSession` query). The request body is no longer trusted, so client-supplied `memberships` can never reach the session. Errors return `401`/`500` with structured logs (no silent fallback).
- **`getAppMetricsData`**, **`getTransactionData`**, **`AppIdLayout`**: authorize via `getIsUserAllowedToReadApp(appId)` instead of cookie `memberships`. (Also fixes the `extractIdsFromPath(path, ["Teams"])` casing bug that left `team_id` undefined in metrics error logs.)
- Rewrote `tests/unit/update-session.test.ts`: asserts a forged-membership body is ignored, authz keys off the session id, and `401`/`500` error paths.

## Testing
- `npx tsc --noEmit` ✅
- `npx jest tests/unit/update-session.test.ts` ✅ (5/5)
- `prettier` ✅

## Rollout / compatibility
- `update-session` adds one indexed `user_by_pk` lookup per session refresh (the client already fetches the same row client-side; load increase is a single PK read). Behaviour is otherwise unchanged for legitimate users — session memberships now always reflect DB truth instead of stale/forged client data.
- No schema/migration changes. Independently revertible.


Gr1dlock:
@codex review

bin-umar (approved):

PR #1966, opened 2026-06-24
fix(reviews): canonicalize nullifier + atomic review counting (#3771446, #3703658)

Gr1dlock:
## Summary
Fixes two app-review integrity issues in `POST /api/v2/app/submit-app-review`:

- **#3771446 (review dedup bypass):** dedup — the `GetAppReview` lookup, the upsert, and the `app_reviews_nullifier_hash_key` UNIQUE constraint — keyed on the **raw** `nullifier_hash`. `verifyProof` canonicalizes the nullifier internally, so every hex re-encoding (`0xABC` / `abc` / `0x0abc` / leading-zero variants) verified the *same* proof but stored a *distinct* row → a single verified human could cast **unlimited** reviews and move `rating_sum`/`rating_count` arbitrarily.
- **#3703658 (counter TOCTOU):** the new-vs-edit decision (`rating_count += 1` vs `+= 0`) was made off a **non-transactional read**, so N concurrent identical submissions each incremented the cached counter, inflating the public star average / ranking inputs (`app-store.ts`).

## Changes
- New `canonicalizeNullifierHash()` helper (`api/helpers/verify.ts`): collapses case/prefix/leading-zero re-encodings to a **fixed-width `0x`+64 lowercase hex** value. This is **idempotent on the standard IDKit form already stored**, so re-encodings now collide on the existing UNIQUE constraint with no data backfill for the common case.
- Handler now canonicalizes `nullifier_hash` before every DB op, and replaces the read-then-decide upsert with an **atomic INSERT**:
  - insert succeeds → new review → `rating_count += 1`, `rating_sum += rating`.
  - uniqueness violation → edit → fetch prior rating, update it, `rating_sum += (new − old)`, `rating_count += 0`.
  
  The UNIQUE constraint serializes concurrent submissions, so the count increments **exactly once** regardless of concurrency. Non-uniqueness insert errors now fail closed (`500`, structured log) instead of being silently miscounted.
- Removed the now-unused `UpsertAppReview` mutation.

## Testing
- `npx tsc --noEmit` ✅
- `npx jest tests/api/v2/app/submit-app-review.test.ts` ✅ (4/4): new-review insert, **re-encoded-nullifier collapses to one canonical value**, unique-violation→edit (count unchanged), non-uniqueness error→500.
- `prettier` ✅

## Rollout / follow-up
- No schema/migration change; relies on the existing `UNIQUE(nullifier_hash)`. Independently revertible.
[...]

PR #1967, opened 2026-06-24
fix(verify): canonical nullifier storage to enforce max_verifications (#3771261)

Gr1dlock:
## Summary
Fixes HackerOne **#3771261**: a nullifier hex-encoding race that bypassed `max_verifications` on `POST /api/v2/verify/{app_id}`.

The write-side protections were split across two columns:
- the only `UNIQUE` constraint (`unique_nullifier_hash`) sits on the **raw** `nullifier_hash` text column;
- all dedupe/limit logic keys on the canonical `nullifier_hash_int` column, which has **no** unique index;
- `enforce_uses_limit` is a `BEFORE UPDATE FOR EACH ROW` trigger (per-row only).

So firing N concurrent verify requests that encode the same nullifier with different-but-equivalent hex (`0xabc` / `0x0abc` / `0xABC` / leading-zero variants) created N **distinct** rows (raw strings differ → no constraint hit), each independently passing the per-row `uses 0→1` trigger → **N verifications for a `max_verifications=1` action** (breaks Sybil resistance for airdrops, one-person-one-vote, faucets).

## Change
`api/v2/verify/index.ts` now stores the nullifier in **canonical fixed-width hex** (`canonicalizeNullifierHash`, added to `api/helpers/verify.ts`) in the atomic upsert. All re-encodings of the same nullifier now produce the **same** `nullifier_hash`, so they collide on `unique_nullifier_hash`: exactly one row is created and the existing per-row uses-limit trigger correctly enforces the cap (concurrent updates serialize on the row lock). This mirrors the **v4** path, which already enforces uniqueness on the canonical value.

The read-side check already used the canonical `nullifier_hash_int`, so it is unchanged.

## Testing
- `npx tsc --noEmit` ✅
- `npx jest tests/api/v2/verify.test.ts` ✅ (16/16) — added a test asserting a re-encoded (uppercased) nullifier is persisted as the canonical lowercase value.
- `prettier` ✅

## Rollout / follow-up
- No schema/migration change; relies on the existing `unique_nullifier_hash` constraint. Canonicalization is idempotent on the standard IDKit `0x`+64-hex form already stored. Independently revertible.
- **Defense-in-depth follow-up (separate, ops-coordinated):** add a `UNIQUE` index on `nullifier_hash_int` to match v4. This must be done with `CREATE UNIQUE INDEX CONCURRENTLY` *after* de-duplicating any existing sibling rows left by the bug, so it is intentionally **not** bundled here.
- Shares the `canonicalizeNullifierHash` helper with #1966 (submit-app-review); if that merges first, this branch's identical addition resolves trivially.


Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

bin-umar (approved):

Gr1dlock:
@copilot review

PR #1968, opened 2026-06-24
fix(actions): portal-authoritative external_nullifier (#3749623)

Gr1dlock:
## Summary
Fixes HackerOne **#3749623**: the `api_key` Hasura role can insert an **arbitrary `external_nullifier`** on an action, and the `generate_external_nullifier` event trigger deliberately **preserved** a pre-supplied value (early-returned unless it equalled `app_id`). This let a developer pin a non-canonical proof domain on their own action, bypassing the server-computed `hash(app_id, action)` invariant the first-party create paths enforce. (Triaged *difficult / low* — scoped to the attacker's own apps, no secret gained — but a real least-privilege/validation gap worth closing at the source.)

## Why fix it at the trigger (write side), not at verify
Recomputing the domain at **verify** time risks breaking verification for legacy/edge action types. Fixing it at the **trigger** is behavior-preserving for every legitimate flow and removes the tampered value before it can ever be used:
- **UI create-action** already let the trigger compute → unchanged.
- **v2 create-action** computes the same `hash(app_id, action)` server-side → the trigger now recomputes the identical value (idempotent overwrite).
- **Attacker raw `insert_action_one`** with an arbitrary value → trigger overwrites with the canonical value → **fixed**.
- **Legacy custom-nullifier apps** (`APPS_WITH_CUSTOM_EXTERNAL_NULLIFIER`, e.g. Grants `0xB16B00B5`) → preserved.

## Changes
- `api/_gen-external-nullifier`: replace "preserve any pre-supplied value" with "recompute & overwrite, except for `APPS_WITH_CUSTOM_EXTERNAL_NULLIFIER`". The `SetExternalNullifier` mutation runs as the service role, which overwrites (the old "will fail due to permissions" comment was inaccurate).
- Hoist `APPS_WITH_CUSTOM_EXTERNAL_NULLIFIER` to `lib/constants` and import it in both `precheck` (previously a local duplicate) and the trigger, so the legacy allowlist has a single source of truth.

## Testing
- `npx tsc --noEmit` ✅
- `npx jest tests/api/gen-external-nullifier.test.ts tests/api/v1/precheck.test.ts` ✅ (14/14)
- `prettier` ✅
- Per prior repo guidance, no new heavy-SDK-mock handler tests were added for this Hasura event-trigger endpoint; the change is behavior-preserving and the legacy allowlist is shared with precheck's existing semantics.

## Rollout
- No schema/migration change. The trigger is async (fires after insert), so there is a brief eventual-consistency window before a tampered value is overwritten — acceptable for this low-severity, own-app-scoped gap. Independently revertible.

### DIFF
--- a/web/api/_gen-external-nullifier/index.ts
+++ b/web/api/_gen-external-nullifier/index.ts
@@ -1,6 +1,7 @@
 import { errorResponse } from "@/api/helpers/errors";
 import { protectInternalEndpoint } from "@/api/helpers/utils";
 import { validateRequestSchema } from "@/api/helpers/validate-request-schema";
+import { APPS_WITH_CUSTOM_EXTERNAL_NULLIFIER } from "@/lib/constants";
 import { generateExternalNullifier } from "@/lib/hashing";
 import { logger } from "@/lib/logger";
 import { NextRequest, NextResponse } from "next/server";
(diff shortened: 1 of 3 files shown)


Takaros999 (approved):
👍

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

PR #1969, opened 2026-06-24
fix(verify): surface verified issuer_schema_id in v4 verify results (#3751156)

Gr1dlock:
## Summary
Fixes HackerOne **#3751156**: the World ID 4.0 verify endpoint copied the caller-supplied `responses[].identifier` verbatim into `results[].identifier` with **no binding** to the cryptographically-verified `issuer_schema_id` — and the result didn't expose `issuer_schema_id` at all. A relying party that authorized on the echoed `identifier` label could be misled (e.g. a device-level proof labeled `"orb"`).

(Triaged *difficult / low*: `issuer_schema_id` is the authoritative field and **is** verified on-chain; `identifier` is never used for any authorization decision in this codebase. Real impact requires a downstream RP misusing the echoed label. This is a hardening fix.)

## Change
Surface the **verified** `issuer_schema_id` in the success results of both verify paths:
- `uniqueness-proof/verify-v4.ts` → `UniquenessResult.issuer_schema_id`
- `session-proof/verify-util.ts` → `SessionResult.issuer_schema_id`

On success the on-chain proof is bound to `issuer_schema_id` (verification fails for any other value), so RPs can now authorize on the verified field instead of the caller-controlled `identifier`. The field is present **only on success** (where it's confirmed). `identifier` is preserved as a passthrough/correlation label for backwards compatibility — it is intentionally *not* removed or enum-constrained, since v4 treats it as a free-form caller tag (unlike v3's credential-level enum).

## Testing
- `npx tsc --noEmit` ✅
- `npx jest tests/api/v4/verify.test.ts tests/api/v4/session-proof.test.ts` ✅
- New `tests/api/v4/process-uniqueness-proof-v4.test.ts` ✅ — asserts the verified `issuer_schema_id` is surfaced on success and absent on failure (mocks only the on-chain RPC boundary).
- `prettier` ✅

## Rollout
Purely additive to the response (new optional field). No schema/migration change, no breaking change for existing integrations. Independently revertible.

### DIFF
--- a/web/api/v4/verify/session-proof/verify-util.ts
+++ b/web/api/v4/verify/session-proof/verify-util.ts
@@ -5,6 +5,10 @@ import { getSessionCommitment } from "@worldcoin/idkit-server";
 
 export interface SessionResult {
   identifier: string;
+  // The cryptographically-verified credential type. Present only on success,
+  // where the on-chain proof is bound to this value. Relying parties should
+  // authorize on `issuer_schema_id`, not the caller-supplied `identifier`.
+  issuer_schema_id?: string;
   sessionId: string;
   success: boolean;
   nullifier?: string;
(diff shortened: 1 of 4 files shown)


Takaros999:
this is the intended behaviour

PR #1971, opened 2026-06-25
feat(permissions): foundation — central team-permission policy

soamdesai-tfh:
## Foundation: central team-permission policy

Introduces the single source of truth for team permissions.

### What's here
- **`lib/team-permissions`** — `PERMISSION_RULES` (action → roles) and the helpers `roleCanPerformAction` / `userCanPerformAction` / `getTeamPermission`, plus the `useTeamPermission` hook
- **`lib/utils`** — `getUserTeamRole`
- Unit tests for the policy and the hook

Client gating built on these helpers is **UX only** — server-side enforcement (Hasura RLS) remains the real authorization boundary and is unchanged by this PR.

### Verification
- `tsc --noEmit`: clean
- `jest tests/unit/permissions tests/unit/use-team-permission`: 7 passed
- `prettier --check`: clean

---
Originally the first of a 3-PR stack (foundation → enforcement → disable-not-hide UX). The enforcement and disable-not-hide PRs are dropped in favor of the portal v3 rebuild, which consumes this policy directly.


web/lib/team-permissions/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  roles: [Role_Enum.Owner, Role_Enum.Admin],
+  message: OWNER_ADMIN_MESSAGE,
+};
+
+const OWNER_ONLY: PermissionRule = {
+  roles: [Role_Enum.Owner],
+  message: OWNER_ONLY_MESSAGE,
+};
+
+export const PERMISSION_RULES = {
```
Takaros999: this is not very TS idiomatic, and won't scale if add new roles

here's an example:
```
export const PERMISSION_REGISTRY: Record<Permission, Role[]>
```

where every permission maps to 1 role or more

web/lib/team-permissions/index.ts
Takaros999, line 1: I don't understand the "team permission" approach, isn't user/team specific?

we could rename it to a more generic `lib/permissions` and have a `useGetPermission` (or better naming)

web/lib/permissions/get-permission.ts
```diff
@@ -0,0 +1,54 @@
...
+ * always tracks the policy and scales when a role is added:
+ *   [Owner]                -> "Only Owners can perform this action."
+ *   [Owner, Admin]         -> "Only Owners and Admins can perform this action."
+ *   [Owner, Admin, Member] -> "Only Owners, Admins, and Members can perform this action."
+ */
+export const permissionMessage = (roles: Role_Enum[]): string =>
+  `Only ${roleList.format(roles.map((role) => ROLE_LABELS[role]))} can perform this action.`;
+
+export const roleCanPerformAction = (
+  role: Role_Enum | undefined,
```
Takaros999, line 34: The role should be strictly typed, having `undefined` here seems error-prone. 

I was thinking to the `Member` role but even that doesn't seem right; this should be explicitly defined

web/lib/permissions/use-get-permission.ts
```diff
@@ -5,11 +5,11 @@ import { useUser } from "@auth0/nextjs-auth0/client";
...
-  type Permission,
   type PermissionResult,
-} from "@/lib/permissions/policy";
+} from "@/lib/permissions/get-permission";
+import { type Permission } from "@/lib/permissions/policy";
 
-export type { PermissionResult } from "@/lib/permissions/policy";
+export type { PermissionResult } from "@/lib/permissions/get-permission";
 
 export const useGetPermission = (
```
Takaros999, line 14: this hook always returned you the result for a single permission, what if we need to learn about 2 or more permissions for a single page?

Would a `usePermissions` that returns a `{ [permission]: { allowed: boolean; message?: string; } }` make more sense?

Takaros999 (commented):
High level. This PR introduces a new permission registry, which means there are permissions set in 3 places:
- The new permission registry
- `lib/permissions/index.ts` which is used server-side to gate certain actions
- DB RLS policies that block the user from certain actions at the database level

I'm not sure we can end up with 1 source of truth but we could settle with 2:
- DB RLS policies we have right now
- Use the permission registry for both client and server side checks. We should document that this is "best-effort" on top of the DB policies

PR #1973, opened 2026-06-25
feat(permissions): disable-not-hide UX for restricted team actions

soamdesai-tfh:
## Disable-not-hide UX for restricted team actions

Replaces role-based **hiding** with the shared permission policy across the app. Restricted controls now render **disabled with an explanatory tooltip** instead of disappearing — so users can discover a capability exists and learn who can use it.

### Scope (49 files)
- **App configuration** — AppTopBar, BasicInformation, MiniApp config, and the AppStore form sections (gating folded into the `isEditable`/`canEditStore` prop)
- **World ID 4.0** — incl. RP action restrictions (`rp-action-restrictions`)
- **World ID actions** — list, settings, danger
- **Team** — members (row actions + `PermissionsDialog` now derives rows from the policy), API keys, navigation tabs
- Nav: `LoggedUserNav`, `AppSelector`

### Verification
- `tsc --noEmit`: clean
- `jest` (17 suites): 100 passed
- `prettier --check`: clean

> The local-dev auth/flag shim used while testing locally was intentionally **excluded** from all three PRs.

---
Stacked on #1972 (which is stacked on #1971) — merge in order. Largest layer of the 3-PR stack.


kchaw2005:
@codex review

PR #1975, opened 2026-06-25
(fix): stale /apps reroute on first time account->team->app creation flow

kchaw2005:
## Issue
(fix) first time account creators, after clicking through to createAppDialog, would be sent to a stale /teamid/apps/ because of refetch asking for app_id[0] despite a possible lagging first write

## Fix
return the created app_id from validateAndInsertAppServerSideV4 (the insert already had insert_app_one.id, it just wasn't passed back) and navigate straight to it instead of re-querying the app list and guessing apps[0] by newest created_at. since the id comes from the same write, there's no dependency on the read replica having caught up. also moved router.refresh() out from behind the latestApp?.id guard so it always fires, with a fallback to /<teamid>/apps so a missing id can never strand the user on a stale page.

### DIFF
--- a/web/scenes/Portal/layout/CreateAppDialog/index-v4.tsx
+++ b/web/scenes/Portal/layout/CreateAppDialog/index-v4.tsx
@@ -128,16 +128,20 @@ export const CreateAppDialogV4 = ({
         });
         return;
       }
-      const [refetched] = await refetchApps();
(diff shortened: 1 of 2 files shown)


kchaw2005:
@codex review

Takaros999 (approved):

PR #1982, opened 2026-06-29
chore: add warning for signing key rotation 

Takaros999:
Add a warning that signing key rotation might take 10 minutes to fully propagate.


kchaw2005 (approved):
LGTM

PR #1985, opened 2026-06-29
feat(portal-v3): first-slice shell

soamdesai-tfh:
First-slice shell for Developer Portal v3, switched by a local-only flag (`LOCAL_DEV_PORTAL_V3_ENABLED`). **Flag off → v2 unchanged.** First of a stacked series (A → B → C).

## What's here
- **Switching layer** — `isPortalV3Enabled` (server-only) + a pure, ordered route-mode classifier (`getPortalV3RouteMode` → typed modes, first-match rule table) + a sync `renderPortalScene`/`pickPortalComponent` chooser.
- **Design tokens** — additive `--v3-*` CSS vars + semantic Tailwind colors (light-only; 0 collisions with v2 utilities).
- **Shell** — `V3Shell` (server; user + teams resolved from the session) + presentational primitives (sidebar nav, team/app switchers, user popup, `ShellFrame`). Nav role-gates API Keys/Settings (Owner/Admin) via `checkUserPermissions`.
- **Structural mounting** — `(portal)` chooser + `PortalChromeGate`, `teams/[teamId]` mounts `V3Shell`, `apps/[appId]` access guard (drops chrome), minimal `/profile` X-chrome.
- **Proofs** — `/actions` renders the v2 body inside the v3 shell (compat); `/kiosk` is untouched (exempt).

### DIFF
--- a/web/app/(portal)/layout.tsx
+++ b/web/app/(portal)/layout.tsx
@@ -1,2 +1,5 @@
+import { renderPortalScene } from "@/lib/feature-flags/portal-v3/render-portal-scene";
 import { PortalLayout } from "@/scenes/Portal/layout";
-export default PortalLayout;
+import { PortalLayoutV3 } from "@/scenes/PortalV3/layout";
+
+export default renderPortalScene(PortalLayout, PortalLayoutV3);
--- a/web/app/(portal)/teams/[teamId]/apps/[appId]/layout.tsx
+++ b/web/app/(portal)/teams/[teamId]/apps/[appId]/layout.tsx
@@ -1,4 +1,6 @@
+import { pickPortalComponent } from "@/lib/feature-flags/portal-v3/render-portal-scene";
 import { AppIdLayout } from "@/scenes/Portal/Teams/TeamId/Apps/AppId/layout";
+import { AppIdLayoutV3 } from "@/scenes/PortalV3/Teams/TeamId/Apps/AppId/layout";
 import { ReactNode } from "react";
 import { AppLayoutRouteParams } from "./layout-params";
 
@@ -10,5 +12,6 @@ type Props = {
 export default async function Layout(props: Props) {
   const params = await props.params;
   const { children } = props;
-  return <AppIdLayout params={params}>{children}</AppIdLayout>;
+  const C = pickPortalComponent(AppIdLayout, AppIdLayoutV3);
+  return <C params={params}>{children}</C>;
 }
--- a/web/lib/feature-flags/portal-v3/flag.ts
+++ b/web/lib/feature-flags/portal-v3/flag.ts
@@ -1,0 +1,12 @@
+import "server-only";
+
+/**
+ * Local-only kill switch for Developer Portal v3.
+ *
+ * Gated on LOCAL_DEV_PORTAL_V3_ENABLED so v3 stays off in every deployed
+ * environment unless an engineer explicitly sets the var to the exact string
+ * "true" in their local shell. Any other value (unset, "1", "false", "TRUE")
+ * is treated as off — fail-safe to the v2 experience.
+ */
+export const isPortalV3Enabled = (): boolean =>
+  process.env.LOCAL_DEV_PORTAL_V3_ENABLED === "true";
--- a/web/lib/feature-flags/portal-v3/render-portal-scene.tsx
+++ b/web/lib/feature-flags/portal-v3/render-portal-scene.tsx
@@ -1,0 +1,35 @@
+import { createElement, type ComponentType, type JSX } from "react";
+import { isPortalV3Enabled } from "@/lib/feature-flags/portal-v3/flag";
+
+/**
+ * Pick the v3 component when the flag is on and a v3 counterpart exists,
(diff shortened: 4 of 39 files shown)


web/lib/feature-flags/portal-v3/flag.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Local-only kill switch for Developer Portal v3.
+ *
+ * Gated on LOCAL_DEV_PORTAL_V3_ENABLED so v3 stays off in every deployed
+ * environment unless an engineer explicitly sets the var to the exact string
+ * "true" in their local shell. Any other value (unset, "1", "false", "TRUE")
+ * is treated as off — fail-safe to the v2 experience.
+ */
+export const isPortalV3Enabled = (): boolean =>
```
Takaros999, line 11: This FF needs to be per team_id like we did for world id 4.0  by using the SSM parameter store

web/lib/feature-flags/portal-v3/render-portal-scene.tsx
```diff
@@ -0,0 +1,35 @@
...
+ */
+export function renderPortalScene<P extends object>(
+  V2: ComponentType<P>,
+  V3: ComponentType<P> | null,
+): (props: P) => JSX.Element {
+  const PortalScene = (props: P) =>
+    createElement(pickPortalComponent(V2, V3), props);
+  // Named so it shows up in React DevTools / stacks instead of "Anonymous"
+  // (also satisfies the react/display-name lint rule).
+  PortalScene.displayName = "PortalScene";
```
Takaros999, line 33: why do we need 2 helpers here? also not clear to me why `PortalScene.displayName = ...` just became relevant? what is the difference with what we did before?

web/scenes/PortalV3/layout/PortalChromeGate.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const PortalChromeGate = (props: { color: Color | null }) => {
+  const pathname = usePathname() ?? "";
+
+  // When the v3 shell owns this route, it supplies its own chrome — suppress
+  // the shared v2 Header. For kiosk/exempt + the /teams index, keep the v2
+  // Header (shared chrome for v2/exempt routes).
+  if (shouldRenderPortalV3Shell(pathname)) {
+    return null;
+  }
```
Takaros999, line 16: why do we need this? we could leave a page like kiosk as it is now, no need to make it "v3"

web/app/(portal)/teams/[teamId]/apps/[appId]/layout.tsx
```diff
@@ -10,5 +12,6 @@ type Props = {
 export default async function Layout(props: Props) {
   const params = await props.params;
   const { children } = props;
-  return <AppIdLayout params={params}>{children}</AppIdLayout>;
+  const C = pickPortalComponent(AppIdLayout, AppIdLayoutV3);
```
Comment (Takaros999) [inline web/app/(portal)/teams/[teamId]/apps/[appId]/layout.tsx:15]: nit. I would check if we can use `renderPortalScene` here, this Layout code here seems wrong in the first place

Takaros999 (commented):
* the tests seem very repetitive, not adding a lot of value. Testing the feature flag utilities and an integration test testing the routing should be enough 
* nit. on naming `Shell` i think we should move this under the `Scenes/PortalV3/layout` folder unless you plan to use these components somewhere else, in that case you could move them to `Scenes/PortalV3/common`

web/lib/feature-flags/portal-v3/render-portal-scene.tsx
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Build a scene component that defers the v2/v3 choice to render time.
+ * Used as the Style-A shim default export:
+ *   export default renderPortalScene(V2Page, V3Page);
+ * A compat route passes V3 = null so the v2 page renders inside the shell.
+ */
+export function renderPortalScene<P extends object>(
+  V2: ComponentType<P>,
+  V3: ComponentType<P> | null,
+): (props: P) => JSX.Element {
```
igorosip0v, line 28: We can use single object arg here, it will increase readability IMO. 

`renderPortalScene({ V2: AppLayout, V3:  AppLayoutV3})`

Not sure if the project prefers positional args here, so feel free to ignore if this goes against the local style.

web/app/(portal)/layout.tsx
```diff
@@ -1,2 +1,5 @@
+import { renderPortalScene } from "@/lib/feature-flags/portal-v3/render-portal-scene";
 import { PortalLayout } from "@/scenes/Portal/layout";
-export default PortalLayout;
+import { PortalLayoutV3 } from "@/scenes/PortalV3/layout";
+
+export default renderPortalScene(PortalLayout, PortalLayoutV3);
```
igorosip0v, line 5: WDYT of not adding V3 to the name? Import path is already telling that it's a v3 component. And for now to not to cause naming conflicts we can simply import with alias 

`import { PortalLayout as PortalLayoutV3 } from "@/scenes/PortalV3/layout";`

I think it could make future refactoring a bit easier.

PR #1990, opened 2026-06-30
feat(portal-v3): per-email activation flag (PR A of 3)

soamdesai-tfh:
Portal v3 activation from **per-team** to **per-user (email)**, so the v2/v3 choice can be made at the portal-layout level.

## What's here
- **Add `isPortalV3EnabledForEmail(email)`** — v3 is on when the global `LOCAL_DEV_PORTAL_V3_ENABLED` switch is on, or when the user's email is in the `PORTAL_V3_EMAILS` allow-list (comma-separated, case-insensitive). Synchronous env read — **no AWS/SSM dependency**.
- **Remove `isPortalV3EnabledForTeam`** (the per-team SSM flag). It was unused on `main`, and the per-team gating it backed is what forced the v3 shell below the v2 `(portal)` layout → the double-header.

## Tests
`portal-v3-flag.test.ts`: global switch, allow-list (case-insensitive/trimmed), non-listed, missing email, and global-forces-on. 6/6 pass; `tsc` + prettier clean.

## Stack
- **PR A (this)** — per-email flag
- PR B — v3 shell as a portal-level layout (dormant)
- PR C — activate per-email at `(portal)/layout` + thin team/app/profile layouts (the header fix)

### DIFF
--- a/web/lib/feature-flags/portal-v3/flag.ts
+++ b/web/lib/feature-flags/portal-v3/flag.ts
@@ -1,11 +1,11 @@
 import "server-only";
 
 /**
- * "true", v3 is turned on. Any other value (unset, "1", "false") is off. defaults to v2.
- * This also makes sure we don't accidentally globally turn for everyone in prod.
+ * Global switch: "true" turns v3 on for everyone in the current environment;
+ * any other value (unset, "1", "false") is off — fail-safe to v2. Leave it
+ * unset in deployed environments; use PORTAL_V3_EMAILS for controlled rollout.
  */
 export const isPortalV3Enabled = (): boolean =>
-  process.env.NODE_ENV !== "production" &&
   process.env.LOCAL_DEV_PORTAL_V3_ENABLED === "true";
 
 /**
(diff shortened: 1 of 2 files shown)


web/lib/feature-flags/portal-v3/flag.ts
```diff
@@ -1,30 +1,28 @@
...
 /**
- * Local-only kill switch for Developer Portal v3.
- *
- * Gated on LOCAL_DEV_PORTAL_V3_ENABLED so v3 stays off in every deployed
- * environment unless "true" in local shell.
+ * "true", v3 is turned on. Any other value (unset, "1", "false") is off. defaults to v2.
+ * This also makes sure we don't accidentally globally turn for everyone in prod.
  */
 export const isPortalV3Enabled = (): boolean =>
+  process.env.NODE_ENV !== "production" &&
```
Takaros999: I wouldn't recommend this. we might need to ship breaking UI we don't want to disrupt staging

Takaros999 (approved):

PR #2062, opened 2026-07-02
Simplify create app dialog to name-only

kchaw2005:
## What
- Remove Mini App toggle, App URL, and Category from the create app dialog.
- Create flow now only asks for an app name; mini app setup stays in Configuration.

## Why
Developer feedback: Users don't know their App URL on average on first app creation and the mini app toggle appears later anyways to configure that info. This makes the user only enter app name on createappdialog

## Test plan
- [ ] Open create app dialog from apps dropdown — only "App name" field shows
- [ ] Create an external app — lands on app dashboard successfully
- [ ] Confirm app is created as external (`app_mode: external`) with default integration URL
- [ ] Enable Mini App later in Configuration → toggle, URL, and store fields still work


web/graphql/graphql.ts
```diff
@@ -183,8 +183,6 @@ export type RegisterRpOutput = {
   operation_hash?: Maybe<Scalars["String"]["output"]>;
   rp_id: Scalars["String"]["output"];
   signer_address?: Maybe<Scalars["String"]["output"]>;
-  staging_operation_hash?: Maybe<Scalars["String"]["output"]>;
```
Takaros999, line 186: why this change? was it auto generated?

kchaw2005: these were just graphQL generations

Takaros999 (approved):

PR #2070, opened 2026-07-04
perf(portal-v3): faster app dashboard navigation

kchaw2005:
The app dashboard layout blocked navigation on two sequential GraphQL
queries (auth guard + FetchAppEnv) before its loading.tsx Suspense
boundary could paint, and the page then re-fetched FetchAppEnv a second
time. Collapse to a single cache()-deduped fetch owned by the page, and
stream the World ID sub-tabs so the layout no longer blocks on app-env.

- add loading.tsx skeleton for the app dashboard route
- page reads FetchAppEnv via fetchAppEnvCached (one request per nav)
- new AppWorldIdSubTabs server component renders the World ID sub-tabs
  from the same cached fetch, streamed via <Suspense>
- AppIdChrome takes a worldIdTabs slot instead of flag props
- page surfaces upstream FetchAppEnv failure as 500 (not masked)


kchaw2005:
@codex review

PR #2071, opened 2026-07-04
fix(portal) Make all portal pages remmeber AppID based on what's selected in AppDropdown

kchaw2005:
## What

Pages used useparams? to remember current appID. `Members`, `API Keys` etc. had **no app id** in the URL, so they would send back to dashboard /apps when clicking to /configuration from one of those pages in the sidebarnav

## Fix
export the lastAppID as a jotai atom from app dropdown, remember it in sidebarnav

## Before & After

https://github.com/user-attachments/assets/dfd036e4-e230-4a8a-92f3-5692cd9d66b8

https://github.com/user-attachments/assets/a46824de-7f84-47ac-b308-ecbdec46a08b


Takaros999 (changes requested):
i wasn't able to reproduce this lets talk offline

Takaros999 (approved):

PR #2076, opened 2026-07-07
fix: allow compound entity ids

Takaros999:
Apps with `app_staging_...` identifier always return 404 
Screenshot 2026-07-06 at 9 00 41 PM


kchaw2005 (approved):
lgtm seeded staging app and cross checked nice regex fix

PR #2078, opened 2026-07-07
feat(proxy): implement internal dashboard host handling and security …

igorosip0v:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

- Rewrite `/` → `/admin` on `INTERNAL_DASHBOARD_HOST`
- Host matching via normalized `x-forwarded-host` / `host`
- CSP and security headers on the rewritten response
- Add `INTERNAL_DASHBOARD_HOST` to `.env.example`
- Unit tests for dashboard host rewrite behavior

## Checklist


web/proxy.ts
```diff
@@ -159,16 +203,42 @@ const createSecurityHeadersResponse = (
...
+  // "/" has no product meaning of its own on the dashboard hostname, so
+  // treat it as the dashboard entry point there. Every other host keeps
+  // today's behavior of not running middleware on "/" at all (it wasn't in
+  // `config.matcher` before this route was added).
+  if (pathname === "/") {
+    return isInternalDashboardHost(request)
+      ? createDashboardRewriteResponse(request)
+      : NextResponse.next();
+  }
+
```
Takaros999, line 243:
```suggestion
  if (pathname === "/" && isInternalDashboardHost(request)) {
    return createDashboardRewriteResponse(request);
  }

```

Takaros999 (approved):

web/proxy.ts
igorosip0v: Since "/" is now in the middleware matcher, public "/" would also enter this proxy. 

The suggested change would let it continue through the Auth0 middleware path instead of preserving the previous "middleware did not run for /" behavior, so I kept an explicit NextResponse.next() fallback and only expanded the ternary into early returns.

PR #2084, opened 2026-07-08
design(landing): Port final draft of landing page from figma

kchaw2005:
## What and Why
Federico redesigned the landing page into something cleaner and more brand aligned. I've ported it over here from figma while preserving the intended animations on buttons and page load. Also ported the mobile version.

## New Images

Web:

https://github.com/user-attachments/assets/2b72fcdf-c886-4355-a6fd-6c43579c366c

Mobile:

https://github.com/user-attachments/assets/2f5a011e-02c3-461f-b2ee-0aefbcae14ce


Takaros999 (commented):
LGTM, we need to address 2 things:
- Fonts are loading client-side; this manifests as a jitter effect where the page loads with default font and then everything switches to "World Pro", you could look into https://nextjs.org/docs/app/getting-started/fonts
- The font weight and sizes are not matching figma. 
  - for the font weights, I think we're not properly loading all of them correctly 
  - for the font sizes - eg. the hero title - go one by one make sure we're aligned with figma


This screenshot is from the current design, you can clearly see font sizes and weights are off compared to figma

Image

kchaw2005:
@codex review

Takaros999 (approved):

PR #2090, opened 2026-07-13
fix(login-callback): consume team invites atomically (H1 #3857870)

Gr1dlock:
## Summary

Fixes **HackerOne #3857870**: the `GET /api/auth/login-callback` invite flow could turn a single-use team invite into **multiple memberships** under concurrent requests.

The old flow ran three separate, un-transactioned GraphQL round-trips:
1. fetch + validate the invite,
2. `insert_membership_one`,
3. `delete_invite_by_pk` (a null result was only logged, never rolled back).

With **no `UNIQUE(team_id, user_id)` constraint** on `membership`, two requests bearing the same `invite_id` could both pass step 1 and each insert a membership in step 2, so one invite produced several memberships. The World ID path also skips the invite-email match by design, widening who can consume a known `invite_id`.

## Fix

Consume the invite and create the membership in **one atomic transaction** via a new `accept_team_invite(_invite_id, _user_id)` Postgres function (tracked as a Hasura mutation for the `service` role), following the existing `merge_world_id_accounts` precedent:

```sql
DELETE FROM invite WHERE id = _invite_id RETURNING team_id INTO _team_id;
IF NOT FOUND THEN RETURN; END IF;               -- already consumed → no membership
RETURN QUERY INSERT INTO membership (...) VALUES (_team_id, _user_id, 'MEMBER') RETURNING *;
```

The `DELETE` on the invite PK is the **concurrency gate**: Postgres row-locks the invite, so concurrent callers serialize and only the one that actually removes the row creates a membership. A single-use invite therefore yields **exactly one** membership even under a burst, and a failed insert rolls back the delete, so the invite is never lost (no partial-commit hidden failure).

Because Hasura resolves a function's nested relationships from a pre-`INSERT` snapshot, the returned `user.memberships` omits the just-created row; the handler adds it back (dedup-guarded) so the session reflects the joined team.

## Rollout safety

- Migration is `CREATE OR REPLACE FUNCTION` only — **no table locks, no rewrite** — safe for rolling deploys.
- No data backfill; nothing to canary. Reversible via `down.sql` (`DROP FUNCTION`).
- The two now-unused operations (`insert-membership`, `delete-invite`) are removed.

## Tests

Full integration suite green (16 suites / 105 tests). New `login-callback` integration tests (run against real Hasura+Postgres):
- **concurrent burst**: 8 concurrent requests with one invite → exactly one membership, invite consumed once.
- **already-consumed**: reusing a consumed invite creates no second membership.

`tsc --noEmit` and prettier pass.

## Follow-up (out of scope)

`join-callback` (the new-user onboarding path) combines insert+delete into one mutation but doesn't gate the insert on the delete succeeding, so it retains a narrower version of this race for brand-new accounts. Worth a separate hardening pass.


Gr1dlock:
@codex review

Gr1dlock:
@codex review

bin-umar (approved):

PR #2091, opened 2026-07-13
Internal developer portal auth roles

igorosip0v:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

- Added a dedicated read-only Hasura role for the internal dashboard.
- Removed the privileged `service` role from dashboard data access.
- Added pluggable authentication providers, currently supporting Cloudflare Access and local development.
- Unified provider permissions through the provider-neutral `Read` access level.
- Added JWT prefiltering in Proxy, local authorization checks, and tests preventing the dashboard from receiving excessive permissions.

## Checklist


web/lib/admin-auth/providers/cloudflare-access.ts
```diff
@@ -56,11 +59,83 @@ const parseGroupNames = (claim: unknown): string[] => {
       return [];
     }
 
+    const id = (group as Record<string, unknown>).id;
     const name = (group as Record<string, unknown>).name;
-    return typeof name === "string" ? [name] : [];
+
+    return [
```
Takaros999: nit. For important files like this, it would be nice to make them easier to read, this bit takes a while to digest when in reality is really simple!

could this be simplified to:
`[id, name].filter((accessGroup): isString(accessGroup))` ? or something similar

web/lib/admin-auth/providers/cloudflare-access.ts
```diff
@@ -56,11 +59,83 @@ const parseGroupNames = (claim: unknown): string[] => {
...
+    return null;
+  }
+
+  const groups = new Set(providerGroups);
+  const readGroups = groupsByAccessLevel[DashboardAccessLevel.Read] ?? [];
+
+  return readGroups.some((group) => groups.has(group))
+    ? DashboardAccessLevel.Read
+    : null;
+};
```
Takaros999: suggestion. This would become much simpler if passed as env var: `CF_GROUP_TO_ACCESS_LEVEL` and map the CF/OKTA group directly to the appropriate dashboard access level

Takaros999 (approved):

web/lib/admin-auth/providers/cloudflare-access.ts
igorosip0v: Updated

web/lib/admin-auth/providers/cloudflare-access.ts
igorosip0v: Updated

PR #2092, opened 2026-07-13
feat(miniapps) redesign page to fit all components on one screen, add explanations for "permit2 tokens", etc.

kchaw2005:
## What
UI / UX improvements to the mini apps page. made all fields fit on one page, added explanations via on-hover tooltips for the permit2 tokens and contract entrypoints fields in case people don't know what they are, auto insert "https://" if dev wants to add URL to permitted domains,.

## Also in this diff (agent audit)
Things the summary above doesn't cover:

- **(fix) Mini Apps with the "External" store category could not send notifications** — both the send-notification API and the Notifications page conflated the store *category* with `app_mode`. Only `app_mode` gates now, with regression tests on both sides.
- **(fix) the Notifications page now targets the same metadata row the API resolves** (verified first, draft fallback), so `draft_id` can't desync from what actually sends. The manual version switcher is gone, and a loading skeleton replaces the flash of the wrong state.
- The CSV textareas on Permissions became a shared **EntryList**: one row per entry with copy + remove, truncated addresses, "show 10 more" batching for long lists, comma-separated paste, and whole-batch validation with case-insensitive duplicate detection at entry time instead of after autosave.
- The duplicate **Mini App / External selector is removed from Permissions** — app mode is owned by Configuration only; external apps see a "Permissions unavailable" notice here instead of a half-empty form.
- The **"Disable whitelist" checkbox is gone**: enforcement now derives from the list itself (empty = off), which makes the old "enforced but empty" invalid state unrepresentable and deletes the manual autosave error for it.
- The notification limit became a **slider card** (0 / 1 / 2 / Unlimited) in a sticky rail beside a rebuilt **Mini App preview card** (QR + copy link), replacing the bare QR column.
- The mode cards on Configuration switched to **container queries** so they wrap based on card width, not viewport width.
- **+545 lines of tests** across three new files plus updated API tests: external-category regressions, metadata precedence, EntryList batching/validation, allowlist derivation, slider values.

## New Look
Screenshot 2026-07-14 at 2 11 57 PM


kchaw2005:
@codex review

kchaw2005:
@codex review

kchaw2005:
@codex review

kchaw2005:
@codex review

Takaros999 (approved):

PR #2093, opened 2026-07-13
fix(portal): allow v3 members to view team settings

Takaros999:
This PR fixes the v3 team settings authorization regression.

- allows v3 team members to open the consolidated settings page while keeping writes owner-only
- preserves the v2 owner-only behavior and blocks users outside the team
- adds proxy regression coverage for both portal versions
- validates the change with focused Jest tests, formatting, and TypeScript checks


kchaw2005 (approved):
LGTM, tested owner and member views of the settings page and they displayed the intended behavior ^

PR #2100, opened 2026-07-14
feat(portal-v3): add World ID trend foundations

soamdesai-tfh:
## What this PR does

- Adds local-midnight weekly buckets and lifetime buckets for weekly and all-time analytics.
- Adds the shared period selector and interactive sparkline components used by the World ID overview and action-detail pages.
- Adds shared trend-query state handling for loading, error, retry, labels, and range text.
- Uses a lifetime-wide fallback when creation time is unavailable instead of labeling a weekly window as all time.
- Adds decimal-nullifier-to-hex formatting and compact relative timestamps for verification rows.
- Adds app-level and action-level GraphQL trend queries with generated Apollo hooks.
- Adds unit coverage for bucket calculation, the missing-creation-time fallback, and both formatting helpers.

## Stack

**1/4 (this PR)** → #2101 → #2102 → #2103

## Validation

- `pnpm format:check`
- `pnpm typecheck --noEmit`
- `pnpm lint`
- Utility unit tests
- Next.js production build on Tailwind 4

### DIFF
--- a/web/lib/day-buckets.ts
+++ b/web/lib/day-buckets.ts
@@ -1,0 +1,110 @@
+import { addDays, startOfDay } from "date-fns";
+
+/** Seven trailing-day buckets aligned to the viewer's local midnight. */
+export const weekBucketBounds = (now: Date): string[] => {
+  const todayMidnight = startOfDay(now);
+
+  const bounds: string[] = [];
+  for (let k = 0; k <= 7; k++) {
+    bounds.push(addDays(todayMidnight, k - 6).toISOString());
+  }
+  return bounds;
+};
+
+/** Seven equal-width buckets from creation through tomorrow's local midnight. */
+export const lifetimeBucketBounds = (createdAt: Date, now: Date): string[] => {
+  const end = addDays(startOfDay(now), 1).getTime();
+  const requestedStart = createdAt.getTime();
+  const start =
+    Number.isFinite(requestedStart) && requestedStart < end
+      ? requestedStart
+      : startOfDay(now).getTime();
+  const span = end - start;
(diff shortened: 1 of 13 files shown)


Takaros999 (changes requested):
im trying to write a PR review, i was thinking to respond with this:
  This seems wrong directly.

  1. read-time queries like this will kill the DB, we should have a pattern similar to `app_stats` where we roll up the queries
  2. I just realized that right now we don't include `nullifier_v4` into `app_stats` which means we're missing a lot of data

  moving forward, what if we:
  - create an action_stats table
```sql
  CREATE TABLE action_stats (
    action_id     TEXT NOT NULL,          -- action.id OR action_v4.id
    source        TEXT NOT NULL,          -- 'legacy' | 'v4'
    app_id        TEXT NOT NULL,          -- denormalized for app-level rollups
    date          DATE NOT NULL,
    verifications BIGINT NOT NULL DEFAULT 0,
    unique_users  BIGINT NOT NULL DEFAULT 0,  -- per (action, date) only
    PRIMARY KEY (action_id, source, date)
  );
  CREATE INDEX ON action_stats (app_id, date);
```
  - and use that to aggregate at the app level. and show to the user which verifications come from legacy vs new actions

  A faster win here, for the redesign, would be to show actions without their stats and keep the analytics part as we have it here, and follow up with the new tables mentioned above

desaisoam:
Closing because this implementation was superseded by #2114–#2116. The code is preserved for future reuse at [`archive/pr-2100`](https://github.com/desaisoam/developer-portal/tree/archive/pr-2100).

PR #2101, opened 2026-07-14
feat(portal-v3): add the World ID overview

soamdesai-tfh:
## What this PR does

- Adds the Portal v3 `/world-id` page and its overview GraphQL query.
- Shows unique-verification totals, this-week totals, and weekly/all-time trends in the hero card.
- Adds searchable action cards and the World ID 4.0 action-creation dialog.
- Keeps action cards on `/world-id-actions/[actionId]`; PR #2102 renders the new detail UI on that same route.
- Adds the World ID 4.0 tab with RP metadata, registration retry, signer-key reset, and self-managed migration controls.
- Adds RP setup for apps that have not registered an RP and keeps the World ID 3.0 legacy link when legacy actions exist.
- Reconciles RP status with the chain when the 4.0 tab mounts and refreshes the parent query if the stored status is stale.
- Keeps cached overview content visible when a refetch fails, catches retry failures, and adds tab/search accessibility attributes.
- Sets the action-search input to the same 40px height as the tab controls.

## Stack

#2100 → **2/4 (this PR)** → #2102 → #2103

## Validation

- `pnpm format:check`
- `pnpm exec tsc --noEmit`
- `pnpm lint`
- World ID route and utility unit tests
- Next.js production build on Tailwind 4

PR #2102, opened 2026-07-14
feat(portal-v3): add World ID action detail

soamdesai-tfh:
## What this PR does

- Reuses `/world-id-actions/[actionId]` for the new Portal v3 detail while keeping Portal v2 on its existing page.
- Removes the duplicate `/world-id/actions/[actionId]` route; old v3 Settings and Danger deep links render the same new detail content without changing Portal v2.
- Shows unique-verification totals, this-week totals, and weekly/all-time trends for an action.
- Polls the first verification-feed page every five seconds and keeps pagination stable with an ID tie-breaker.
- Keeps the stats and settings shell mounted while another feed page loads, so changing pages does not discard an in-progress settings edit.
- Preserves cached action content when polling or refetching fails and shows an unknown weekly value as `—` instead of a real zero.
- Schedules a trailing aggregate refresh when a feed update arrives inside the 60-second throttle window.
- Adds inline action settings and only renders deletion controls when the current user has server-backed update permission.
- Hides pagination for an empty feed and preserves the loading state during deletion navigation.
- Splits feed polling from aggregate GraphQL queries so five-second polls do not refetch trend statistics.

## Stack

#2100 → #2101 → **3/4 (this PR)** → #2103

## Validation

- `pnpm generate:graphql-types`
- `pnpm format:check`
- `pnpm exec tsc --noEmit`
- `pnpm lint`
- Action-detail route unit tests
- Next.js production build on Tailwind 4


desaisoam:
Closing because this implementation was superseded by #2114–#2116. The code is preserved for future reuse at [`archive/pr-2102`](https://github.com/desaisoam/developer-portal/tree/archive/pr-2102).

PR #2107, opened 2026-07-14
feat(portal, config): report autosave failures to telemetry

Gr1dlock:
## Context

While investigating why Holdstation Wallet got **413 Content Too Large** on *Submit for review*, the failure was hard to diagnose partly because the client **swallows the error**: a failed config autosave surfaces only as a local "Couldn't save" pill, and during the submit-for-review `flushAll()` as a generic "Some changes could not be saved" toast. Nothing is reported, so we had zero signal about *which* form failed or *why* (the actual cause was a WAF rule returning 413 for ~8 KB Server Action bodies — see world-id-deploy#545).

## Change

`useAutosaveWithStatus` now emits a `config_autosave_failed` PostHog event once per transition into the `error` state, tagged with `form_id` and the error `message`.

- Pure addition — no control-flow change; the existing status/pill behavior is untouched.
- Fires for real save failures (network / proxy / server errors) via the autosave error path; deduped per transition so retries of the same failure don't spam.
- Consistent with the existing `app_submit_review_attempted` PostHog events in this feature.

With this, the Holdstation case would have shown `config_autosave_failed { form_id: "advanced-setup", message: … }` instead of being invisible — aligned with the "no swallowed errors / no observability blind spots" principle.

## Tests

Added `use-autosave-with-status.test.tsx` (mirrors the existing `use-autosave` suite): reports once with form id + message on failure, silent on success, once per error transition.

> Note: I could not execute `jest`/`tsc` in this sandbox (the checkout's deps aren't resolvable by the local `npx`); relying on CI + `@codex review` for verification. The change was validated by inspection against the existing test patterns and the hook's null-safe context handling.

## Follow-up / related

- The user-facing blocker is fixed by **world-id-deploy#545** (WAF `BodySizeLimit` rule raised from 8 KB → 1 MB). This PR is the observability half.

### DIFF
--- a/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/Configuration/hook/use-autosave-with-status.ts
+++ b/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/Configuration/hook/use-autosave-with-status.ts
@@ -1,5 +1,6 @@
 "use client";
 
+import posthog from "posthog-js";
 import { useCallback, useEffect, useRef } from "react";
 import { FieldValues, UseFormReturn } from "react-hook-form";
 import { useSaveStatusActions } from "../SaveStatus";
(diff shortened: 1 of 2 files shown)


Gr1dlock:
@codex review

PR #2108, opened 2026-07-14
fix(portal): prevent mini apps from using the External category

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Mini apps could be saved with the "External" category, which silently hid them from the mini app store, because the mini-app toggle (`app_mode`) and the `category` field were fully decoupled with no validation tying them together. This PR blocks the External category for mini apps in the submit-for-review schema (both Portal and PortalV3, enforced client- and server-side), hides the External option from the category dropdown for mini apps, and defaults newly created mini apps to "Other" instead of "External". It also fixes a bug in the public `/api/v2/public/apps` endpoint where the highlights filter used `||` and only dropped apps that were external on both `category` and `app_mode` (unlike the top-apps filter's `&&`), so a mini app tagged "External" leaked into highlights. Added a regression test for the highlights filter and schema tests covering both portals.

## Checklist

- [x] I have self-reviewed this PR.
- [x] I have left comments in the code for clarity.
- [x] I have added necessary unit tests.
- [ ] I have updated the documentation as needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2122, opened 2026-07-16
fix(portal): coerce External category on remaining app_mode write paths

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Follow-up to #2108. That PR stopped a mini app from ending up under the store-hiding "External" category via the config toggle, autosave, creation, and submit-for-review, but two other `app_mode` write paths could still leave the invalid state in a draft (both were backstopped by the submit-for-review block, so neither could reach the store — this just makes the data consistent everywhere). This PR routes MCP `create_app` through the shared `resolveAppStoreCategory` helper so an explicit `category:"External"` on a mini app is coerced to "Other", and extends the `UpdateSetup` mutation (used by the Advanced / Mini App Permissions form) with the same conditional clear as `UpdateAppMode` — switching to mini-app now resets a lingering "External" category to "Other" in the same request. The invariant "mini app ⇒ category ≠ External" now holds at every write path.

## Checklist

### DIFF
--- a/web/api/mcp/index.ts
+++ b/web/api/mcp/index.ts
@@ -35,7 +35,10 @@ import {
   uploadAppImage,
 } from "@/api/helpers/app-image-storage";
 import { checkRateLimit } from "@/api/helpers/rate-limit";
-import { CategoryNameIterable } from "@/lib/categories";
+import {
+  CategoryNameIterable,
+  resolveAppStoreCategory,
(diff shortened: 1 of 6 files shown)


Gr1dlock:
@codex review

Takaros999 (approved):

PR #2128, opened 2026-07-20
portal(world-id sandbox): add sandbox testflight links to v2 header and v3 navbar (sidebar)

kchaw2005:
## What

Added a button on the sidebar that allows teams to register for **the world ID sandbox** for ios or android via link or QR code. 

For iOS
- user accesses public testflight link

For Android
- user must request access via valid google associated email first in order to be added to the Google Play invite list, only then will the link/QR be valid. User sees session email on request initiation, can correct it before sending, and a row will be created in the table `sandbox_access_request`

## Other additions
- Introduced the OpenFeature SDK with a hardcoded teams provider to set up future feature flags
- made this PR in world-id-deploy to intiate ssm parameter creation, which our created provider now reads from
- made a new table `sandbox_access_request` with fields `id,google_email,requested_by,team_id,status,created_at,processed_at` filled in from a mix of user input and existing session values

Screenshot 2026-07-20 at 7 31 53 PM
Screenshot 2026-07-20 at 7 31 48 PM
Screenshot 2026-07-20 at 7 31 43 PM


kchaw2005:
@codex review

kchaw2005:
@codex review

web/scenes/PortalV3/layout/index.tsx
```diff
@@ -12,8 +13,18 @@ export const PortalLayout = async (props: { children: ReactNode }) => {
     .filter((t): t is NonNullable<typeof t> => !!t?.id)
     .map((t) => ({ id: t.id, name: t.name ?? "Untitled team" }));
 
+  // Flags evaluate server-side only; the shell receives decisions, never lists.
+  const sandboxTeamIds = await getSandboxTeamIds(
```
igorosip0v: @kchaw2005 It would be clearer to expose feature flags through a single, discoverable interface with the following shape:

`featureFlags.<featureName>.<featureSpecificField>`

For example:

- `featureFlags.sandboxDistribution.getSandboxTeamIds()`
- `featureFlags.uiV4.isEnabled()`

This makes the source of each decision immediately obvious when reading the code and provides a consistent `featureFlags` entry point for discovering and using available flags.

kchaw2005: implemented the interface for this

kchaw2005:
@codex review

kchaw2005:
@codex review

kchaw2005:
@codex review

hasura/migrations/default/1784830339000_create_table_public_sandbox_access_request/up.sql
```diff
@@ -0,0 +1,13 @@
...
+        "google_email" varchar NOT NULL,
+        "requested_by" varchar NOT NULL,
+        "team_id" varchar NOT NULL,
+        "status" varchar NOT NULL DEFAULT 'pending',
+        "created_at" timestamptz NOT NULL DEFAULT now (),
+        "processed_at" timestamptz,
+        PRIMARY KEY ("id"),
+        FOREIGN KEY ("team_id") REFERENCES "public"."team" ("id") ON UPDATE restrict ON DELETE cascade,
+        CONSTRAINT "unique_sandbox_access_request_email_team" UNIQUE ("google_email", "team_id")
+    );
```
Takaros999, line 12: few comments:
- why do we need `team_id` 
- the `unique_sandbox_access_request_email_team` doesn't make sense, what is the goal here? Since this a very manual process i think it's fine if we have duplicate entries
- what is `requested_by` by here? should this be `user_id` ?

the `sandbox_request` here is too long compared to our other tables, what about `sbxreq` ? 
-  `"id" varchar NOT NULL DEFAULT gen_random_friendly_id ('sandbox_request'), `

web/scenes/PortalV3/layout/Shell/SandboxButton.tsx
```diff
@@ -0,0 +1,345 @@
...
+/** Static distribution links for the sandbox builds. Update here only. */
+const IOS_TESTFLIGHT_URL = "https://testflight.apple.com/join/VZEurhHe";
+// Google Play internal test: works only for Google accounts already on the
+// internal-tester allowlist (managed in Play Console, outside this repo).
+const ANDROID_URL: string | null =
+  "https://play.google.com/apps/internaltest/4701115249455610230";
+/** TestFlight's own App Store page, for testers who don't have it yet. */
+const TESTFLIGHT_APP_STORE_URL =
+  "https://apps.apple.com/app/testflight/id899247664";
+
```
Takaros999: these can't be hardcoded here, need to be passed fron .env or something, so we can change them easily

Takaros999 (commented):
I realized while reviewing this, some initial assumptions/requirements are a) making this PR harder b)  will accumulate tech debt. 

I'd like to make few suggestions:
- Drop the feature flag based on team ids. My initial idea was here, was that it would be nice to whitelist this for specific teams and their members. since we dont have a good FF mechanism in place this would mean when the same user switches team they would stop seeing the sandbox tab. (potentially keep a single boolean value that hides the sandbox tab from the UI) 
- This could be a separate PR, but we need to have an admin page to view and process all the sandbox requests, and "approve them"
- Apart from the table it would be nice to send an email to `sandbox.access@toolsforhumanity.com`, you can ask IT to make an email group with everyone from the developers team, murph and few android engineers

web/scenes/PortalV3/layout/Shell/SandboxButton.tsx
kchaw2005: made commit in world-id-deploy repo to add links to env variables at process time

hasura/migrations/default/1784830339000_create_table_public_sandbox_access_request/up.sql
kchaw2005: i guess we don't now that it's not a feature flag gated by team ID
we wanted to implmement upserts for google emails but we are now impementing upserts for user IDs
requested_by was the users portal email. we renamed it to `portal_email`
renamed the unique id to sbxreq

kchaw2005:
@codex review

kchaw2005:
@codex review

Takaros999 (approved):

PR #2136, opened 2026-07-21
fix(world-id): keep signer key setup open during lazy loading

soamdesai-tfh:
## Summary

- keep Configure Signer Key visible while the lazily loaded key screen initializes
- show a spinner in the existing Continue button, then swap directly to the requested key screen
- cover Generate New Key and Use Existing Key in both Portal implementations

## Root cause

#2132 changed the signer-key screens to bare `next/dynamic` imports. On a cold load, changing steps synchronously allowed the key-screen suspension to reach the dialog's outer Suspense boundary. That temporarily hid the dialog panel, and Headless UI closed the panel after it disappeared. Once the chunk was cached, the retry succeeded.

The step change now runs in a React transition. React keeps the already-rendered Configure screen mounted while the requested lazy component loads, preserving lazy loading without showing the white fallback or closing the dialog.

## Validation

- `npx jest tests/unit/world-id-key-step-loading.test.tsx --runInBand` — 4 tests passed
- `npx tsc --noEmit`
- `pnpm format:check`
- local simulation with a forced 3-second cold load confirmed Configure stays visible, the Continue spinner appears, and the next screen renders directly


kchaw2005 (approved):
lgtm

PR #2139, opened 2026-07-21
fix(config): Apollo 4 refetch -> straight query for showcase images on config page 

kchaw2005:
## What and Why
- replace refetch() with a call to the actual function because refetch() throws when a lazy query had never been executed yet. both callers in that page upload before calling get image, so the query its needed hasn't been executed

- remove extra getImage call after initial uploadImageVia.. cause it already checked with getImage within the method
-


kchaw2005:
@codex review

kchaw2005:
@codex review

soamdesai-tfh (approved):

PR #2140, opened 2026-07-22
fix(config): image upload survives mid-autosave form remounts — keep wizard step, persist metadata, show success toast

kchaw2005:
### Motivation
  Follow-up to #2125 → #2139. Fixing the Apollo 4 `refetch` crash let the image
  upload flow run further than it ever had, exposing a second latent bug: the
  AppStoreFormProvider` is keyed on `${appMetadata.id}-${viewMode}` (both
 portals), so the form can remount mid-upload when its own autosave/refetch
 changes that identity. `ImageUploadField` bailed on unmount *before*
persistence, so on staging an upload would: kick the user back to the first
wizard step, skip the metadata autosave + image refetch + success toast, and
leave the preview stale until a hard refresh — while S3 already had the file.

### Description
- `scenes/PortalV3/.../Configuration/page.tsx`: lift the wizard `activeStep`
  state out of `ConfigurationContent` into `AppProfilePage`, above the keyed
  provider boundary — a provider remount can no longer reset the user to the
  first step. (v3-only: v2 has no wizard.)
- `scenes/PortalV3/.../ImageForm/ImageUploadField.tsx` and the v2 copy
  (`scenes/Portal/...`): restructure `uploadAcceptedImage` so that once S3 has
  accepted the file, persistence (`onAutosave`, `onRefetchImages`) always runs
  — persistence is ownerless; only UI updates (`onChange`,
  `onUploadSuccess`) remain gated on the instance still being mounted. The
  refetch writes into the shared Apollo cache, so a remounted successor
  instance re-renders with the new image immediately, and the mutation's
  `onCompleted` success toast now fires reliably.
- v2 is included deliberately: it has the identical keyed provider
  (`page.tsx:105`) and identical bail, so the defect was reachable on the
  production path — same-defect-in-both-trees gets parity-fixed.

Known limitation (intentionally out of scope): these changes make the upload
flow *robust against* the remount rather than preventing it. Why the key
churns during a save is still undiagnosed — tracked separately; the fix there
is likely keying on `app_id + viewMode` instead of the metadata row id.

### Testing
- `npx jest tests/unit/use-image-upload.test.tsx tests/unit/pv3-config-redesign.test.tsx`
  — 18/18 pass (the upload suite pins the single-request flow and rejects on
  S3 refusal; any reintroduced post-upload request fails it by construction).
- `npx tsc --noEmit` clean, `pnpm format:check` clean.
- Manual repro on staging pending deploy: upload a showcase image → expect
  success toast, preview appears without refresh, wizard stays on the
  Store listing step.
  wizard step, skip the metadata autosave + image refetch + success toast, and
  leave the preview stale until a hard refresh — while S3 already had the file.

  ### Description
  - `scenes/PortalV3/.../Configuration/page.tsx`: lift the wizard `activeStep`
    state out of `ConfigurationContent` into `AppProfilePage`, above the keyed
    provider boundary — a provider remount can no longer reset the user to the
    first step. (v3-only: v2 has no wizard.)
  - `scenes/PortalV3/.../ImageForm/ImageUploadField.tsx` and the v2 copy
    (`scenes/Portal/...`): restructure `uploadAcceptedImage` so that once S3 has
    accepted the file, persistence (`onAutosave`, `onRefetchImages`) always runs
    — persistence is ownerless; only UI updates (`onChange`,
    `onUploadSuccess`) remain gated on the instance still being mounted. The
    refetch writes into the shared Apollo cache, so a remounted successor
    instance re-renders with the new image immediately, and the mutation's
    `onCompleted` success toast now fires reliably.
  - v2 is included deliberately: it has the identical keyed provider
    (`page.tsx:105`) and identical bail, so the defect was reachable on the
    production path — same-defect-in-both-trees gets parity-fixed.

  Known limitation (intentionally out of scope): these changes make the upload
  flow *robust against* the remount rather than preventing it. Why the key

### DIFF
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/ImageForm/ImageUploadField.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/ImageForm/ImageUploadField.tsx
@@ -155,8 +155,9 @@ export const ImageUploadField = (props: ImageUploadFieldProps) => {
 
         if (isMountedRef.current) {
           onChange(newUrls);
-          onUploadSuccess?.();
         }
+        // Parent toast / bookkeeping — must not be skipped on remount mid-upload.
+        onUploadSuccess?.();
       } catch (error) {
         console.error("error uploading image:", error);
 
--- a/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/ImageForm/ImageUploadField.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Apps/AppId/Configuration/AppStore/ImageForm/ImageUploadField.tsx
@@ -139,21 +139,24 @@ export const ImageUploadField = (props: ImageUploadFieldProps) => {
           isLocalized ? locale : undefined,
         );
 
-        // check if component is still mounted/valid before updating
-        if (!isMountedRef.current) {
-          return;
-        }
-
+        // S3 has the file from here on — the remaining steps are bookkeeping
+        // and must run even if this instance unmounts mid-flight (the form
+        // provider is keyed on metadata id + view mode and can remount during
+        // autosave). Persistence is ownerless; only UI updates are gated.
         const extractedPath =
           extractImagePathWithExtensionFromActualUrl(imageUrl);
         const newUrls =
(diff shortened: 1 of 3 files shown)


kchaw2005:
@codex review

soamdesai-tfh (approved):

PR #2141, opened 2026-07-22
fix(world-id): check on-chain signer before promoting RP to registered

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Fixes H1 #3869823. The v4 `rp-status` production reconciliation path promoted a managed RP to `registered` from on-chain `initialized && active` alone, without comparing the on-chain signer to the Portal's stored `signer_address` — a check the staging path in the same file already performed — so after a failed managed registration, whoever won the permissionless on-chain `register()` for that `rp_id` with their own signer could flip the row to `registered` and make `proof-context` serve the legitimate app's verified branding bound to an attacker-controlled OPRF signer (impersonation). This PR extracts the signer-trust logic into a shared `canTrustOnChainSigner` helper used by both the production and staging paths (so they can't silently diverge again), preserves the DB status and logs a warning on mismatch, and gates the DB writeback on the trust flag. Existing `rp-status` tests were updated to include the on-chain `signer` the real RPC always returns, and two tests were added: one asserting a foreign on-chain signer is not promoted (the takeover case), and one confirming self-managed RPs (no expected signer) still promote.

Note: this prevents future poisoning but does not auto-heal a row already at `registered` from a prior exploit (mismatch preserves the stored status, matching the staging path); the new warning log surfaces any such mismatches for follow-up.

## Checklist

### DIFF
--- a/web/api/helpers/rp-utils.ts
+++ b/web/api/helpers/rp-utils.ts
@@ -65,6 +65,38 @@ export function normalizeAddress(address: string): string {
   return `0x${address}`;
 }
 
+/**
+ * Whether an on-chain RP reading may be trusted as authoritative for the
+ * Portal's stored registration, judged by the signer.
+ *
+ * Returns true when there is no expected signer to compare against
+ * (self-managed mode — the developer owns the on-chain signer, so on-chain is
+ * authoritative by definition), or when the on-chain signer matches the
+ * Portal's expected `signer_address`.
+ *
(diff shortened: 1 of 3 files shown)


Gr1dlock:
@codex review

Gr1dlock:
@codex review

PR #2142, opened 2026-07-22
fix(portal): keep team lists fresh after create/delete without hard refresh

soamdesai-tfh:
## Summary

- revalidate `useMeQuery` (FetchMe) on mount so a newly created team shows up in the profile Teams list without a hard refresh
- `deleteTeamServerSide` rewrites the Auth0 session cookie before resolving; the dialog then `router.refresh()`es so the v3 sidebar team switcher drops the deleted team without a hard refresh
- post-delete navigation runs directly in `submit` (not in an effect), so it still happens if the session update unmounts the dialog mid-flight
- unit tests: refresh-and-close with teams remaining, last-team redirect without refresh, and navigation surviving a mid-delete unmount

## Root cause

Two separate staleness mechanisms, one per surface:

**Create → stale Teams table.** The browser Apollo client is a window-level singleton (`ApolloNextAppProvider` stores it on `window`), so the cache survives the client-side round trip through `/create-team` — and team creation is a REST route that never touches Apollo. With default `cache-first`, `FetchMe` answered from the pre-create cache and never hit the network. Now `cache-and-network` + `nextFetchPolicy: cache-first` fetches once per mount — the same approach the v2 nav dropdown already uses for this exact reason.

**Delete → stale sidebar switcher.** The v3 `TeamsDropdown` is server-rendered by the `(portal)` layout from the Auth0 session, and a layout never re-renders during client-side navigation inside its own route group — so after a delete the stale HTML persisted until a full reload. The session cookie is now rewritten inside the delete server action itself (`auth0.updateSession`, mirroring `create-team`), so when the action resolves the cookie is already correct and the dialog can `router.refresh()` deterministically — no racing client-side sync (a failed in-action rewrite is reported as `sessionUpdated: false` and the dialog awaits the independent `/api/update-session` fallback, which re-derives memberships from the DB, before refreshing — the refresh never runs against a knowingly stale cookie). Navigation happens in `submit` from the refetch result because the session update can flip `canWrite` on the team settings page and unmount the dialog — a navigation effect there may never run, stranding the user on the deleted team's URL.

## Validation

- `npx jest tests/unit/delete-team-dialog-refresh.test.tsx` — 4 tests pass (incl. unmount-mid-delete and failed-sync fallback ordering)
- `pnpm test:unit` (392) and `pnpm test:api` (494) pass
- `npx tsc --noEmit` clean, `pnpm format:check` clean
- manual E2E on a local stack: create team → appears in `/profile/teams` via client-side nav only; delete the *current* team from its own settings page → lands on `/profile/teams`, sidebar switcher updated immediately, no hard refresh, no stranding


web/scenes/PortalV3/common/DeleteTeamDialog/index.tsx
```diff
@@ -70,43 +70,39 @@ export const DeleteTeamDialog = (props: DeleteTeamDialogProps) => {
...
-    }
-  }, [team?.id, auth0User?.hasura?.id, refetch]);
 
-  useEffect(() => {
-    if (!deleteFinished || loading) {
-      return;
-    }
+      if (!result.sessionUpdated) {
+        // Fallback sync — the cookie must be correct before the refresh below
+        await fetch("/api/update-session", { method: "POST" }).catch(
```
igorosip0v: I think we should validate the fallback response here. `fetch` doesn't throw on `401` or `500`, and `.catch(() => null)` also hides network errors. In these cases we will still call `router.refresh()` with the old session.

Maybe we can do something like this:

```ts
const response = await fetch("/api/update-session", {
  method: "POST",
}).catch(() => null);

const data = response?.ok
  ? await response.json().catch(() => null)
  : null;

if (!data?.success) {
  toast.error("Team deleted, but session update failed");
  return;
}
```

igorosip0v (approved):
One comment, otherwise LGTM

PR #2144, opened 2026-07-22
fix(portal): deleting the last team no longer strands the user or kicks them to the landing page

soamdesai-tfh:
### Motivation
Deleting your only team from **Team settings** (sidebar) left the portal in a dead end: the success toast fired but you stayed on the deleted team's page, refreshing gave a bare **401 | Unauthorized**, and follow-up navigation logged you out onto the marketing landing page. The onboarding screen had the same failure in miniature: with no team, the header **X** was a logout link (mid-signup) or a history-back onto the logged-out landing page — and it flashed back in briefly right after submitting the create-team form.

### Description
- `scenes/{Portal,PortalV3}/common/DeleteTeamDialog`: the post-delete sequence (delete → refetch memberships → refresh session cookie → navigate) now runs entirely in the submit handler instead of a `useEffect`. The settings page unmounts the dialog the moment `canWrite` collapses after the session refresh, so the pending effect never ran — an async handler keeps executing after unmount. Zero teams left → `/create-team`; otherwise → `/profile/teams` as before.
- `scenes/{Portal,PortalV3}/Teams/TeamId/Apps/page`: the membership guard sent zero-membership users to `/api/auth/logout`; they now go to `/create-team` with the session intact. Foreign-team access still redirects to your own team; no session still logs out.
- `scenes/Onboarding/CreateTeam/layout/CloseButton`: the X only renders when the user has a team to return to, and latches hidden once the session resolves to no teams — the post-create session refresh lands before navigation finishes and would otherwise flash it back in (clickable, straight to the landing page).
- `proxy.ts`: a role-restricted team URL for a team the user has **no membership in at all** (just deleted, or someone else's) redirects to `/` — the root page routes to their first team or `/create-team` — instead of rewriting to the dead-end `/unauthorized`. A member with an insufficient role still gets the 401. Split into its own commit since it changes shared middleware behavior.

### Testing
- New suites: `create-team-close-button`, `apps-page-no-team-redirect` (both portal versions), `delete-team-dialog-navigation` (including a regression test that unmounts the dialog mid-flow); `proxy-auth-origin` updated for the new non-member redirect. Full unit run: 87 suites / 365 tests green; `npx tsc --noEmit` and `pnpm format:check` clean.
- Verified live against local Hasura: create team → delete from Team settings → lands on `/create-team` logged in; deleted-team URL redirects home instead of 401; a DOM observer across the create-team submit confirmed the X never reappears.

### DIFF
--- a/web/scenes/Portal/common/DeleteTeamDialog/index.tsx
+++ b/web/scenes/Portal/common/DeleteTeamDialog/index.tsx
@@ -13,7 +13,7 @@ import { useMeQuery } from "@/scenes/common/me-query/client";
 import { useUser } from "@auth0/nextjs-auth0/client";
 import { yupResolver } from "@hookform/resolvers/yup";
 import { usePathname, useRouter } from "next/navigation";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback } from "react";
 import { useForm } from "react-hook-form";
 import { toast } from "react-toastify";
 import * as yup from "yup";
@@ -42,7 +42,7 @@ export const DeleteTeamDialog = (props: DeleteTeamDialogProps) => {
   const router = useRouter();
   const path = usePathname();
   const { user: auth0User } = useUser() as Auth0SessionUser;
-  const [deleteFinished, setDeleteFinished] = useState(false);
+  const { invalidate } = useUser();
 
   const {
     register,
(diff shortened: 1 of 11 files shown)


web/scenes/Portal/common/DeleteTeamDialog/index.tsx
```diff
@@ -70,42 +70,46 @@ export const DeleteTeamDialog = (props: DeleteTeamDialogProps) => {
...
-  useEffect(() => {
-    if (!deleteFinished || loading) {
-      return;
-    }
+      const refetched = await refetch();
+      const membershipsCount = refetched?.data?.user_by_pk?.memberships?.length;
+
+      // Refresh session claims before navigating — server guards read memberships from the cookie.
+      try {
+        const res = await fetch("/api/update-session", { method: "POST" });
```
igorosip0v, line 82: I think we should check the response here. fetch doesn't throw on 401 or 500, so we still navigate with stale membership claims in the session. Should we return early when `res.ok` is false?

igorosip0v (approved):

PR #2146, opened 2026-07-23
feat(portal): structural loading skeletons across PortalV3 pages

soamdesai-tfh:
## What

Replaces the bare shimmer-block loading states on PortalV3 pages with structural skeletons: real chrome renders during load (tabs, headings, card frames, table headers, back links) and only the value slots shimmer. Covers World ID (actions grid + World ID tab), the four legacy action pages, MiniApp transactions, Sign-in client info, app Danger zone, Team apps grid, and Team API keys, plus a small shared `components/Skeletons` library (labelled form, card, table).

## Why this shape

Skeletons drift when they hand-copy real markup, so the ones here share their geometry with the real components instead of retyping it:

- Team app card and World ID action card skeletons import the frame/text class constants exported by the real `App`/`ActionCard` components, with line heights derived from the real `Typography` variants.
- The danger-zone skeleton renders the real card chrome (`DangerZoneCard`) with the app name/footer shimmered — card measures pixel-identical to the loaded one (768×185 desktop).
- Table skeletons render the real tables' actual header labels; the `CreateActionTile` shown while World ID loads is the same component the loaded grid renders.

## Verification

Loading branches were forced and compared against loaded pages in the browser at 1280/375 (World ID, API keys, Danger zone); data-gated pages (action details, transactions, client info) were verified at source level against their real tables' headers and copy. `tsc`, Prettier, and the unit/API suites pass on this branch standalone.

A follow-up PR rebuilds the Configuration page skeleton on shared layout shells; this PR is independent of it.


kchaw2005 (approved):
clicked through. lgtm.

PR #2149, opened 2026-07-23
feat(proxy): add x-current-path header for admin routing and update P…

igorosip0v:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

This PR prevents admin pages from periodically requesting Auth0 `/api/auth/profile` and receiving `401` responses. Admin pages use separate authentication, but the global PostHog provider previously mounted Auth0 `useUser()` for every route.

- Mounts Auth0 user identification only outside `/admin`.
- Keeps the PostHog provider available and resets its identity on admin pages.
- Passes the resolved route path through the proxy to handle dashboard-host root rewrites.
- Keeps existing admin authorization checks unchanged.
- Adds regression coverage for admin routes, root rewrites, client navigation, and proxy path forwarding.

Tested with:

- `pnpm exec jest tests/unit/root-posthog-provider.test.tsx tests/unit/proxy-auth-origin.test.ts --runInBand` — 29 tests passed.
- Prettier check for all changed files.

## Checklist

### DIFF
--- a/web/scenes/Root/layout/index.tsx
+++ b/web/scenes/Root/layout/index.tsx
@@ -7,9 +7,9 @@ import { IBM_Plex_Mono, Rubik } from "next/font/google";
 import localFont from "next/font/local";
 import { headers } from "next/headers";
 import { Suspense } from "react";
+import "react-image-crop/dist/ReactCrop.css";
 import { SkeletonTheme } from "react-loading-skeleton";
 import "react-loading-skeleton/dist/skeleton.css";
-import "react-image-crop/dist/ReactCrop.css";
 import { Slide, ToastContainer } from "react-toastify";
 
 const rubik = Rubik({
@@ -84,7 +84,7 @@ export const RootLayout = async ({
   // from `web/proxy.ts` to framework and page scripts.
   const requestHeaders = await headers();
   const currentPath = requestHeaders.get("x-current-path");
-  const isAdminRequest =
+  const disableUserIdentification =
     currentPath === "/admin" || currentPath?.startsWith("/admin/");
 
   return (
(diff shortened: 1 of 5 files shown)


web/scenes/Root/layout/index.tsx
```diff
@@ -82,7 +82,10 @@ export const RootLayout = async ({
 }>) => {
   // Force request-time rendering so Next can apply the per-request CSP nonce
   // from `web/proxy.ts` to framework and page scripts.
-  await headers();
+  const requestHeaders = await headers();
+  const currentPath = requestHeaders.get("x-current-path");
```
Takaros999, line 86: im curious if in nextjs using this header is the most reliable way for reading the path

web/scenes/Root/layout/index.tsx
```diff
@@ -95,7 +98,7 @@ export const RootLayout = async ({
         />
 
         <Auth0Provider>
-          <WithPostHogIdentifier>
+          <WithPostHogIdentifier isAdminRequest={isAdminRequest}>
```
Takaros999: i found the naming very confusing. Conceptually we're trying to disable the posthog provider for admin routes.

I feel like we should either have a `disable=boolean` prop in posthog provider or in this file hide conditionally based on the admin routes
eg. 
```
      <WithPostHogIdentifier disable={isAdminRequest}>
```

Takaros999 (approved):

web/scenes/Root/layout/index.tsx
igorosip0v: AFAIK it was an issue some time and still is. 

1. [Issue](https://github.com/vercel/next.js/discussions/90128#:~:text=Currently%2C%20Server%20Components%20do%20not%20have%20direct%20access%20to%20the%20current%20pathname)
2. Next.js docs [recommend](https://nextjs.org/docs/app/api-reference/file-conventions/proxy#:~:text=To%20pass%20information%20from%20Proxy%20to%20your%20application%2C%20use%20headers%2C%20cookies%2C%20rewrites%2C%20redirects%2C%20or%20the%20URL) passing information from proxy this way

web/scenes/Root/layout/index.tsx
igorosip0v: Updated

PR #2155, opened 2026-07-26
fix(portal): repair the MCP setup flow end to end

soamdesai-tfh:
A user on staging generated an MCP setup command, ran it, and nothing worked. This fixes why, plus three adjacent breaks found along the way.

## The root cause

**API keys are per-deployment; the portal hardcoded a production URL.** `MCP_ENDPOINT` was `https://developer.world.org/api/mcp` regardless of which portal you were on. A staging key resolved against staging's Hasura and is verified with an HMAC keyed on staging's `GENERAL_SECRET_KEY`, so it cannot authenticate against production. `authenticate()` runs before the method switch, so even `initialize` failed — the handshake died and no tool ever ran.

Confirmed empirically with the reporter's key: `developer.world.org` → `{"code":-32001,"message":"API key is not valid."}`; `staging-developer.world.org` → full handshake, 11 tools.

**The command itself was correct.** Verified against shipped mcp-remote 0.1.38 and codex-cli 0.144.5: the single-quoted key, the literal unexpanded `${WORLD_DEVELOPER_API_KEY}` in the header (mcp-remote substitutes it, not the shell), the missing space in `Authorization:Bearer`, and `--transport http-only` are all right. Only the hostname was wrong. Tests pin all four so nobody "fixes" them.

## What else was broken

**An existing-key user had no path to the command at all.** `ApiKeySecretFields` only mounts in post-mutation modals; the key row renders non-interactive `Reset to view` text, and the only route back was Reset key — Owner-only, in a `⋮` menu, and it invalidates the key you already deployed. The word "MCP" only appeared when you had zero keys.

**The CTA was a link.** `href={urls.teamSettings(...)}` — no key created, no dialog, no deep link. And the card lives only on the zero-apps page, removed by the redirect the moment any app exists, including one MCP creates.

**Nothing refetched.** Default `cache-first` on a window-scoped Apollo client, a shell that never remounts, and no invalidation signal anywhere. An MCP-created RP stayed invisible until a hard reload.

**A silent runaway poll, not previously noticed.** `GET /api/mcp` returned 200 with a handshake descriptor. The SDK only special-cases 405, so a 200 fell into a reconnect scheduler whose attempt counter is hardcoded to 0 — ~1 req/sec, forever, per connected client, no error emitted. Reproduced against a stub with the real SDK. mcp-remote bundles identical logic, so the command we hand out triggers it.

**The zero-apps page never learns the app exists.** The new in-place flow parks the user on `/teams/:id/apps` while an agent creates their first app out-of-band, but the redirect to that app runs only in the server component (`InitialApp`) — the page had no app query, no focus listener, and no `router.refresh()`, so the welcome state survived until a hard reload. Found in review.

## Commits

Reviewable independently, in landing order:

| | |
|---|---|
| `2a4e073c` | GET → 405, `Allow` header, JSON-RPC envelope guard |
| `27d56bad` | origin-derived endpoint + snippet dedup, drop `--scope project` |
| `df7832c8` | World ID freshness |
| `1477f0fa` | entry point, Owner gate, rotate confirmation, persistent section |
| `8d394209` | zero-apps page detects the MCP-created first app and hard-navigates |

## Notes for review

**The dedup is the point of commit 2.** The two `ApiKeySecretFields` copies were byte-identical. `fa3b8403` shows the failure mode: it fixed `getClaudeMcpCommand` in both trees and left `getCodexMcpCommand`, the next function in both hunks, untouched. Snippet generation now lives in one pure module, and a test asserts each component exports only `["ApiKeySecretFields"]`.

**`--scope project` removed.** `.mcp.json` is designed to be checked into version control, so the old command told users to commit a live secret.

**`cache-and-network` needs the `loading && !data` guard in the same commit.** Alone it makes every warm-cache mount paint a skeleton where `cache-first` painted content — strictly worse.

**Load characteristic:** Chrome fires both `visibilitychange` and `focus` on a tab return, so one return can put two requests on the wire. Apollo dedups the observable so the user sees one loading cycle. No throttle by design; a timestamp ref is the cheap mitigation if it shows up.

**The return check hard-navigates on purpose.** A soft `router.refresh()` relying on the server `redirect()` was rejected: its failure mode is the same silent staleness this fixes, and jsdom can't catch it — a mocked router passes forever. `CreateAppDialogV4` already uses `window.location.replace` after portal-native creation, and the hard nav re-renders the session-rendered shell for free. Unlike the WorldIdPage refetch, lazy executions get no Apollo observable dedup, so this one ships with a 1s timestamp throttle from the start. Checks are suppressed while the create-key modal is open — a navigation must not destroy the one-shot secret — and run once on close, the likeliest moment the MCP-created app already exists. A split-screen session that never blurs the tab still fires neither event; same accepted limitation as the WorldIdPage freshness work.

**Owner gate is seeded server-side.** `useUser` resolves via a client fetch with no SSR seed, so a client-only check would pop the primary CTA in after a round-trip.

**The persistent section never handles a user's key** — it renders a `YOUR_API_KEY` placeholder. An earlier draft had a paste-your-key field; that was removed deliberately, and a test asserts the section renders zero `input`/`textarea` elements.

## Deliberately out of scope

- **ChatGPT provider tab** — hands out `Authorization: Bearer …` as "connector config", but ChatGPT connectors support only OAuth or no-auth, with no field for a customer-supplied token. Product decision; unchanged here.
- **OAuth / `.well-known` metadata.** Don't ship a bare 401 without it — mcp-remote always attaches an auth provider, so it would pop a browser window and hang rather than failing fast. Today's `200` + `-32001` is the better of the two until the metadata exists.
- **CORS on `/api/mcp`** — every advertised client is a native/Node process; `Access-Control-Allow-Origin: *` on a bearer-authenticated endpoint would be a regression.
- **v2 `ClientPage`** has the same navigate-away CTA, deferred.
- **v2 twins of the rotate/persistent-section work** — PortalV3 is the shipping surface.

## Testing

**1051 unit + api tests pass; `tsc --noEmit` clean.** 69 new cases, ~2,000 test lines.

Verified live against a local server, not only mocks: `GET` → 405 with `Allow: POST, OPTIONS`, `null` body → `400`/`-32600`, full MCP handshake, `tools/list` → 11 tools, and `create_app` writing a real row.

Two mutation checks on the freshness work: reverting `loading && !data` reddens 4 cases; deleting the revalidation effect reddens a different 4. The PR4 suites were mutation-tested too — 12 independent mutations to `AppsPageClient`, all 12 killed. The return-path region likewise: guard inversion, throttle removal, dropped navigation, and cache-first each redden their targeted case, 4/4 killed.

The codex command test shells out to `/bin/sh` and proves the CLI form and the TOML config produce identical argv — the two forms that had already drifted once.

## Not verified

The `mcp-remote` + Codex client path was driven as raw JSON-RPC, not through the actual CLI. And the UI was not exercised in a signed-in browser — the freshness behaviour is covered by the mutation-tested suite, not a visual check.

### DIFF
--- a/web/api/mcp/index.ts
+++ b/web/api/mcp/index.ts
@@ -1670,6 +1670,20 @@ export const POST = async (req: NextRequest) => {
     );
   }
 
+  // Must sit ahead of the notification check: null/arrays/primitives parse
+  // cleanly, then either throw reading `.id` or get mistaken for notifications.
+  if (
+    typeof message !== "object" ||
+    message === null ||
+    Array.isArray(message)
+  ) {
+    return jsonRpcError(
+      null,
+      new McpError("Invalid JSON-RPC request.", -32600),
+      400,
+    );
+  }
+
   if (message.id === undefined) {
     return new NextResponse(null, { status: 202 });
   }
@@ -1693,17 +1707,20 @@ export const POST = async (req: NextRequest) => {
   }
 };
 
+// A 200 here reads as an SSE stream to the MCP SDK, which then reconnects
+// ~1/sec forever against an attempt counter it never increments.
 export const GET = async () =>
-  NextResponse.json({
-    name: "world-developer-portal",
-    transport: "streamable-http",
-    endpoint: "/api/mcp",
+  new NextResponse(null, {
+    status: 405,
+    headers: {
+      Allow: "POST, OPTIONS",
+    },
   });
 
 export const OPTIONS = async () =>
   new NextResponse(null, {
     status: 204,
     headers: {
-      Allow: "GET, POST, OPTIONS",
+      Allow: "POST, OPTIONS",
     },
   });
--- a/web/scenes/Portal/Teams/TeamId/Team/ApiKeys/page/ApiKeySecretFields/index.tsx
+++ b/web/scenes/Portal/Teams/TeamId/Team/ApiKeys/page/ApiKeySecretFields/index.tsx
@@ -5,36 +5,15 @@ import { CopyIcon } from "@/components/Icons/CopyIcon";
 import { ExternalLinkIcon } from "@/components/Icons/ExternalLinkIcon";
 import { LockIcon } from "@/components/Icons/LockIcon";
 import { TYPOGRAPHY, Typography } from "@/components/Typography";
+import {
+  getMcpEndpoint,
+  getProviderSnippets,
+  PROVIDERS,
+  type ProviderId,
+} from "@/scenes/common/Teams/TeamId/Team/ApiKeys/page/mcp-snippets";
 import clsx from "clsx";
 import { useMemo, useState } from "react";
 import { toast } from "react-toastify";
-
-const MCP_SERVER_NAME = "worldcoin-developer-portal";
-const MCP_ENDPOINT = "https://developer.world.org/api/mcp";
-const MCP_API_KEY_ENV_VAR = "WORLD_DEVELOPER_API_KEY";
-
-type ProviderId =
-  | "codex"
-  | "claude"
-  | "cursor"
-  | "windsurf"
-  | "chatgpt"
-  | "zed";
-
-type Provider = {
-  id: ProviderId;
-  name: string;
-  setupLabel: string;
-};
-
-type ProviderSnippet = {
-  command: string;
-  rawConfig: string;
-};
-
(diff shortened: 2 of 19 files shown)


kchaw2005:
@codex review

kchaw2005 (approved):
lgtm

PR #2157, opened 2026-07-27
feat(portal): enable v3 for tools for humanity emails

Takaros999:
This PR enables Portal v3 for all authenticated users whose email ends with `@toolsforhumanity.com`.

- Adds `toolsforhumanity.com` to the hardcoded `WHITELISTED_EMAIL_DOMAINS` list.
- Preserves the existing local global switch and `PORTAL_V3_EMAILS` per-address allowlist.
- Covers case-insensitive domain matching and rejects lookalike domains.
- Validates the change with focused Jest, TypeScript, formatting, and diff checks.


kchaw2005 (approved):
lgtm

PR #2158, opened 2026-07-27
fix(portal): refresh image previews after upload

kchaw2005:
## Summary

Stacked on #2145. This follow-up fixes image previews that can remain blank or stale after an upload even though the image was successfully persisted.

- Treat refetched unverified image paths as the preview source of truth after form remounts.
- Synchronize those paths back into the form state for validation and submission.
- Refresh expiring signed URLs with Apollo `cache-and-network` queries.
- Prevent browser caching for signed draft/unverified image responses while preserving normal caching for verified assets.
- Remove the old `window.location.reload()` second-upload workarounds from the V2 and V3 logo/content-card uploaders.
- Cover remount recovery, deletion, locale switching, post-S3 abort handling, and preview cache policy with focused tests.

## Root cause

The configuration form can remount while post-upload autosave/refetch work is still running. The new component instance can receive an empty or stale form value even though the upload and metadata write completed. Separately, draft images use stable S3 object keys, allowing a replacement image to reuse stale browser-cached bytes.

## Impact

After an upload or replacement, the preview should update without requiring a page refresh. Verified/public image cache behavior is unchanged.

## Validation

- `pnpm format:check`
- `npx tsc --noEmit`
- Focused image upload/preview and cache-command Jest tests
- `pnpm build`
- Current CI passed unit, API, integration, end-to-end, production build, and Docker build checks before the split


kchaw2005:
@codex review

PR #2161, opened 2026-07-28
feat(portal): Make the World logo interactable

SeanROlszewski:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

Extends the homepage World logomark pixel field (`BasePixelStrip`) from a static halftone SVG into an interactive piece.

https://github.com/user-attachments/assets/668eecff-3dfd-4e5c-98cc-e9df9a19c8bf

- **Pointer lens**: nearby halftone dots scale up and lighten toward a soft warm gray as the cursor approaches, instead of a flat opacity dim.
- **App icon reveal**: within a tight radius, real World App icons (fetched lazily from `world-id-assets.com`, zero upfront requests, and cached by URL so an icon reused across multiple cells only triggers one network request) develop in place of the dots on one continuous falloff - scale pops steeply near the cursor, while opacity/blur share a gentler falloff so neighbors fade/blur smoothly and z-order hand-offs between icons look continuous instead of snapping. Icons stay part of the grid permanently once revealed, settling to rest size (never re-blurring) when the cursor moves away.
- **Click ripple**: pressing sends a wave expanding outward through the dot grid, blurring any icon it sweeps through. Releasing sends a second wave that's the literal time-reverse of the press wave - continuing seamlessly from wherever the forward wave currently is (not resetting to the outer edge, which used to look like a second wave arriving from nowhere) and playing out much faster, reading as a snap back rather than a slow mirror.
- **Press spring + negative-space push**: clicking a revealed icon holds it near 3x its resting size (rather than shrinking away) so it stays the clearly most prominent thing on screen, while a radial push shoves its immediate neighbors out of the way so they don't crowd/cover it - it's fine for neighbors to look disorganized as a result. The pressed icon also shakes gently in place for as long as it's held.
- **Ambient pop-and-ring**: every so often, an idle dot pops up, scales, and spins a couple of full turns - a passive invitation to interact. Each pop also sends out a slow, quiet ring to its neighbors (a soft scale pulse plus a lightening toward the same warm gray the lens uses), reading as a gentle ripple radiating outward from the spinning tile.
- `icon-manifest.ts` provides 553 unique World App icons sourced from the public app-rankings API, mapped 1:1 to the pixel grid's 1,645 cells (each icon reused 2-3x, placements kept far enough apart that no two copies can ever appear in the same lens at once).
- An icon preprocessing/refresh pipeline (`scripts/normalize-app-icons.ts` + `.github/workflows/refresh-pixel-strip-icons.yml`) fetches, normalizes to a consistent WebP, optionally uploads to S3, and patches the manifest in place on a weekly schedule - without disturbing the cell<->app index mapping - opening a PR rather than pushing directly so a broken upstream response can't silently reach production.
- `app/pixel-probe/` is a dev-only QA route that mounts `BasePixelStrip` directly, bypassing the auth/env setup the real homepage needs, so this can be reviewed/tested without a full local environment. The real homepage shows the same component at `/` for logged-out visitors - no special setup needed there either, since that render path never touches auth/GraphQL and its one network call (network stats) falls back to static values if it fails.

All of the above is layered as direct DOM style writes driven by `requestAnimationFrame`, not React re-renders/state, since ~1,645 SVG elements re-rendering per pointermove would be far more expensive than mutating a handful of style properties per frame.

### Performance

Profiling (CDP traces with invalidation-reason attribution, rAF frame timing, JS self-time instrumentation, and CPU-throttled runs simulating mid/low-tier mobile) found the click/hold interaction dropping well below 60fps - driven by browser style-recalculation cost, not JS compute (`applyLens` itself only ever took 1-2ms at full speed):

- Merged three separate full-grid render passes (lens, ripple, icon-blur wave) into one, instead of each walking all 1,645 cells independently.
- Skip rewriting halftone dot styles - and the separate "settled icon" branch's styles - on frames where the pointer hasn't moved and no ripple is active, since both are pure functions of cursor position that don't need re-writing when nothing's actually changed. The settled-icon case in particular scaled with how much of the grid had already been explored, since revealed icons never un-reveal.
- Narrowed the click ripple's band width so fewer dots are touched per frame (the actual dominant cost, per a CDP trace breakdown of invalidation reasons).
- Skip redundant icon z-re-stacking (`appendChild`, which is a real DOM move even when the result is unchanged) when the stacking order hasn't actually changed frame to frame.
- Cache icon loads by URL instead of by cell index, so an icon reused across 2-3 cells shares one network request/decode instead of each cell fetching independently.
- The new ambient ring effect (see above) is built to stay cheap by construction despite pops now being frequent and concurrent: each pop's neighbor list is computed once via direct grid-lattice lookups, never a full-grid scan, and only the thin band of neighbors the wavefront is currently crossing gets written to on any given frame.

Median frame time during a held click/ripple is back to a steady 16.7ms (60fps) in local testing, down from routinely blowing past 20-33ms; idle ambient animation (pops + rings) holds the same flat 16.7ms with zero dropped frames even with several effects overlapping.

### Security

Pinned third-party GitHub Actions in the icon-refresh workflow to verified commit SHAs instead of mutable tags, and validate `app_id` against a strict identifier pattern before it's used in any file path or S3 key - addressing CodeQL/Semgrep findings from initial review (supply-chain risk and untrusted network data written to file, respectively).

Verified end-to-end against both the live `/pixel-probe` route and the real homepage with Playwright throughout: zero icon requests before hover, correct lazy-fetch and dedup counts, ripple direction/timing (forward growing outward, reverse collapsing inward, seamless continuity when reversed mid-flight), press scale/push/shake behavior, ambient pop/ring timing and decay, clean state reset after every effect settles, and frame-timing measurements (including CPU-throttled) before/after each performance change.

## Checklist


Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock (approved):

PR #2162, opened 2026-07-28
fix(portal): canonicalize nullifier_hash in OIDC authorize (H1 #3896406)

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

`POST /api/v1/oidc/authorize` passed the raw, client-supplied `nullifier_hash` into the nullifier lookup, the nullifier insert, the OIDC auth code, and the id_token `sub`/`email`, while `verifyProof` normalizes the same value internally (`decodeToHexString`) before verification — so case / `0x`-prefix / leading-zero re-encodings of one nullifier all passed proof verification but produced byte-distinct downstream identities, letting a single orb-verified human mint multiple stable OIDC subjects for the same app and create sibling rows in `public.nullifier` (RP-side Sybil). This PR canonicalizes the nullifier with the existing `canonicalizeNullifierHash` helper immediately after `verifyProof` succeeds and uses the canonical value for every identity-forming sink, exactly as the v2/verify handler already does. As a side effect it also closes a concurrency race: two re-encodings submitted with different proofs now collide on the byte-exact `unique_nullifier_hash` constraint instead of inserting sibling rows. The change is safe for existing data because `canonicalizeNullifierHash` is idempotent on the standard `0x`+64-hex form IDKit emits and already stores. Fixes HackerOne #3896406.

## Checklist


Gr1dlock:
@codex review

Sumit-Sabhaya:
Hi @Gr1dlock, just following up on the fix for H1 #3896406.

I noticed PR #2162 is still open and the latest CI run is showing failures in both the API Tests and Integration Tests. The PR also appears to still be awaiting the required code-owner review from @bin-umar.

Could you please take a look at the failing checks and let me know if anything is still needed from my side to help get the fix merged?

Thanks!
Sumit-sabhaya(Hackerone)

PR #2168, opened 2026-07-29
feat(rp-registration): add manager_key_dedicated field and related fu…

igorosip0v:
## PR Type

- [x] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

New managed RPs use a shared KMS manager key when `RP_REGISTRY_MANAGER_KMS_KEY_ID` is set, and the database now tracks whether an RP's key is dedicated to it, so the shared key is never scheduled for deletion.

- Adds `manager_key_dedicated` to `rp_registration`, backfilled to `true` for rows that already have a `manager_kms_key_id`.
- Adds `resolveManagerKey()`: reuses the key from `RP_REGISTRY_MANAGER_KMS_KEY_ID` (full ARN) when set, otherwise creates a dedicated per-RP key as before.
- Only calls `scheduleKeyDeletion()` for dedicated keys, in both the registration rollback and `switch-to-self-managed`.
- Fails managed registration before creating a key or submitting the transaction if the migration is not applied yet.
- Adds tests for the shared and dedicated key paths and the deletion guard.

No behavior change while the variable is unset. The key exists only in staging (worldcoin/world-id-deploy#552).

## Checklist


web/api/helpers/rp-registration-flows.ts
```diff
@@ -126,6 +127,27 @@ export async function submitManagedRpRegistration({
...
   // Slot is now claimed; any failure between here and the final DB write
   // must release it so retries don't bounce off `already_registered`.
+
+  // manager_key_dedicated is written together with the manager key at the end
+  // of this flow. If the migration adding it has not been applied yet, fail
+  // here rather than after the on-chain transaction has been submitted.
+  try {
+    await getVerifySchemaSdk(client).VerifyManagerKeySchema({
+      rp_id: rpIdString,
+    });
```
Takaros999: Im pretty sure the hasura migration runs before the web app deployment, I'm wondering if this case is needed.

web/api/helpers/kms-eth.ts
```diff
@@ -301,6 +312,38 @@ export async function createManagerKey(
   }
 }
 
+/**
+ * Resolves the manager key for a managed RP registration.
+ * Uses the shared key from RP_REGISTRY_MANAGER_KMS_KEY_ID when set;
+ * otherwise creates a dedicated per-RP key.
+ */
+export async function resolveManagerKey(
```
Takaros999: nit. If we create the shared kms key in both environments this would be simplified a lot. I also think we might be mixing the concerns between the 2 files (`kms-eth.ts` and `rp-registration-flow.ts`) 

I see your need about having a mechanism for smooth migration, what if we had a `ENABLE_SHARED_KEY_RP_REGISTRATION` or similar boolean flag, and use that in the `rp-registration-flows` to determine which key to use for the first rp registration

Takaros999 (commented):
lets discuss offline, but high level:
- `manager_key_dedicated` is not a very intuitive name, especially when we have `mode: managed | self-managed` column. Thoughts on `is_shared_manager_key`?

igorosip0v:
@Takaros999 

> manager_key_dedicated is not a very intuitive name… Thoughts on is_shared_manager_key?

Agree the name isn't great. I'd keep this polarity though: column default is `false`, and we only delete the key when the flag is `true`. That way a missing/`false` value never schedules deletion of the shared key. `is_shared_manager_key` inverts that, so I'd rather rename to `is_unique_manager_key`. WDYT?

web/api/helpers/kms-eth.ts
igorosip0v: > ENABLE_SHARED_KEY_RP_REGISTRATION or similar boolean flag

Plan is to test shared-key registration on staging only first, without touching prod deploys. So for now we only wire `RP_REGISTRY_MANAGER_KMS_KEY_ID` in staging; its presence is the switch. A separate boolean would matter if the key already existed in both envs before we were ready to use it.

igorosip0v: Actually, updated

Takaros999 (approved):

igorosip0v:
@codex review

PR #2171, opened 2026-07-30
Fix stale and broken portal links

soamdesai-tfh:
## Summary

- replace the broken homepage community destination and remove the obsolete Events footer link
- update Mini App notification docs, notification API reference, Discord, and default IDKit documentation links in both portal variants
- remove the expired Sign in with World ID deprecation banner whose announcement URL no longer exists
- rename the stale `DOCS_CLOUD_URL` constant to match its current IDKit destination

## Validation

- `pnpm format:check`
- `npx tsc --noEmit`
- `npx jest --no-cache tests/api tests/unit lib/compare-versions.unit.spec.ts lib/schema/schema.unit.spec.ts api/v2/minikit/send-notification/send-notification.schema.unit.spec.ts` — 155 suites / 1,100 tests passed
- confirmed all five replacement destinations return HTTP 200

## Test environment note

The unscoped `npx jest --no-cache` command also discovers integration and Playwright E2E suites. Locally, those service-dependent suites fail without Postgres/Hasura and when Playwright specs are invoked through Jest; the unit/API scope above is green.

### DIFF
--- a/web/components/ActionsHeader/index.tsx
+++ b/web/components/ActionsHeader/index.tsx
@@ -4,7 +4,7 @@ import { DecoratedButton } from "@/components/DecoratedButton";
 import { CaretIcon } from "@/components/Icons/CaretIcon";
 import { DocsIcon } from "@/components/Icons/DocsIcon";
 import { TYPOGRAPHY, Typography } from "@/components/Typography";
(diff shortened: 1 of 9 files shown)


kchaw2005 (approved):
clicked through all links, lgtm

PR #2177, opened 2026-07-30
fix(e2e): align create-team existing-user fixture

Takaros999:
This PR fixes the create-team e2e fixture after the endpoint began deriving existing-user state from Hasura.

- creates the Hasura test user with the same Auth0 ID used by the test session
- preserves generated Auth0 IDs for all existing helper callers
- keeps the existing-user `/teams/:id` and new-user `/teams/:id/apps/` branches distinct
- validates with API-test lint, formatting, typecheck, and `git diff --check`
- leaves the live focused e2e to CI because local execution lacks the CI `AUTH0_SECRET` and staging Hasura access

### DIFF
--- a/tests/api/helpers/hasura.ts
+++ b/tests/api/helpers/hasura.ts
@@ -254,14 +254,21 @@ const createTestUserIdentifiers = () => {
 };
 
 // Helper for creating test user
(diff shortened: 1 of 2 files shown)


kchaw2005 (approved):
lgtm

PR #2180, opened 2026-07-30
fix(portal): polish sidebar account menu

Takaros999:
This PR polishes the Portal V3 shadcn sidebar account experience.

- Shows `Anonymous user` for World ID-authenticated sessions without exposing nullifier-like names.
- Adds pointer cursors to interactive sidebar controls and switcher rows.
- Opens the shadcn profile menu directly above the account row and matches its width.
- Adds focused regression coverage; 32 tests, TypeScript, and formatting pass.


kchaw2005 (approved):
👍

PR #2182, opened 2026-07-30
test(ui): align cursor assertion with global rule

Takaros999:
This PR aligns the Portal V3 shell unit test with the global pointer-cursor rule introduced in #2174.

- Removes the obsolete component-level `cursor-pointer` assertion.
- Removes imports used only by that stale test.
- Keeps production pointer behavior owned by `styles/globals.css`.
- Validates the focused regression test, all 115 unit suites / 584 tests, formatting, and TypeScript.


kchaw2005 (approved):
lgtm

PR #2188, opened 2026-07-31
fix(portal): match app dropdown avatar colors

Takaros999:
This PR aligns fallback app avatar colors between the team apps page and the Portal V3 header dropdown.

- Reuses the shared app placeholder in the dropdown, seeded by the stable app ID.
- Keeps colors consistent in the dropdown trigger, dropdown rows, and team app cards, including after app renames.
- Adds focused unit coverage for deterministic color parity.
- Validated with the focused Jest suite, TypeScript, Prettier, and `git diff --check`.


kchaw2005 (approved):
fine

PR #2191, opened 2026-07-31
feat(analytics): World ID unique-verifications data layer (1/3)

soamdesai-tfh:
First of three stacked PRs: single-metric World ID analytics (Unique Verifications across legacy + v4 actions). Data layer.

- Rollup tables `action_legacy_stats_daily` / `action_v4_stats_daily` + read fn `world_id_analytics_app_daily()`
- `rollup_world_id_analytics(from_date, to_date)` — stateless, idempotent delete+recount of a UTC date window; no dates = trailing ~25h; upper bound clamped to `now − 5 min`; blocking advisory lock; parent-action pre-locks so a concurrent action deletion waits briefly instead of deadlocking
- Route `/_rollup-world-id-analytics` (fail-closed behind `WORLD_ID_ANALYTICS_ROLLUP_ENABLED`, Hasura cron `*/15`, webhook timeout 14 min): every tick rebuilds the trailing window, then advances the one-time history backfill itself — a Redis cursor (`world-id-analytics:rollup:processed-through`, 2-day TTL) records the last completed UTC day so finished chunks are never re-rolled, the first tick derives the start date from the earliest raw nullifier row, and chunking (10-day chunks, one transaction each, dedicated 120 s fetch policy) stops at a 12-minute budget and resumes next tick. No request body, no operator backfill calls. Once logs report `backfill.complete: true`, a follow-up PR deletes the catch-up half and keeps the trailing-window rollup only.

The `created_at` indexes on `nullifier` / `nullifier_v4` are a manual platform-team step (see Deployment) — no ops script is checked in.

Tests: fresh-stack 35/35 (window semantics, both-direction raw/rollup parity, index builds under live writes, cursor-driven backfill incl. halt-and-heal, real Hasura webhook delivery, end-to-end release runbook), route unit 11/11.

## Deployment

**Step 0 — pre-flight.** `HASURA_GRAPHQL_STRINGIFY_NUMERIC_TYPES=true` in prod Hasura; `WORLD_ID_ANALYTICS_ROLLUP_ENABLED` absent or `false` in developer-portal-deployment.

**Step 1 — build the indexes. ✅ Already done** (2026-08-07): run manually by the platform team against **staging and production** — prod builds took 1m17s (`nullifier`) and 0.5s (`nullifier_v4`), and both indexes are confirmed present via `pg_get_indexdef`. For reference / new environments:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS nullifier_created_at_idx
  ON public.nullifier (created_at);
CREATE INDEX CONCURRENTLY IF NOT EXISTS nullifier_v4_created_at_idx
  ON public.nullifier_v4 (created_at);
```

Run with plain `psql`/pgAdmin outside any transaction (`CONCURRENTLY` forbids transactions, including `DO` blocks). After a build, verify `indisvalid` — a canceled concurrent build leaves an `INVALID` index that `IF NOT EXISTS` would then skip; drop it concurrently and rerun.

**Step 2 — merge this PR.** The cron registers, but every tick returns `disabled` without touching the database, and nothing reads the new tables yet.

**Step 3 — set `WORLD_ID_ANALYTICS_ROLLUP_ENABLED=true`** in developer-portal-deployment. The cron backfills all history by itself: each 15-minute tick advances the cursor up to 12 minutes' worth of 10-day chunks, so the whole 2023→today history lands within the first few ticks. Done when the `"Rolled up World ID analytics"` log shows `backfill.complete: true`; from then on each tick maintains the trailing ~25h (plus a no-op cursor check) unattended.

**Step 4 — merge #2192 and #2193.**

**Step 5 — cleanup PR.** After `backfill.complete: true` has held in production, delete the catch-up half of the route (cursor, chunk loop, earliest-row query), leaving the trailing-window rollup — the intended steady state.

**Monitoring.** One Datadog alert: no `"Rolled up World ID analytics"` success log for >30 min (a dead cron emits nothing). Backfill chunk failures log as errors with their exact range and return 500, so the tick shows failed in cron history and retries next tick without skipping past the gap.

**Rollback / repair.** The flag is the kill switch — off stops all recurring work instantly while the UI keeps serving the last-rolled numbers. To re-verify or repair any historical range, delete the Redis cursor key: the next ticks re-walk the whole history (idempotent delete-and-recount). Reverting the PR is safe (additive DDL; `down.sql` exists). The out-of-band indexes remain if the rollout is abandoned; drop them concurrently only if their storage and write overhead are no longer wanted.


hasura/operations/world-id-analytics/create-nullifier-created-at-index.sql
```diff
@@ -0,0 +1,43 @@
+\set ON_ERROR_STOP on
+
+-- Run out of band before enabling the rollup. Hasura migrations are
+-- transactional, so these builds deliberately do not live in a migration.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS nullifier_created_at_idx
```
Takaros999, line 5: not needed, this will be executed manually

web/api/_rollup-world-id-analytics/index.ts
Takaros999, line 1: This script has a few problems:
1. You need to store success/failure for a given chunk otherwise they're going to be retried. you can use redis with a long TTL eg. 2 days
2. You expect from /end data to be passed into the endpoint's body. Why? The backfill should happen everytime the cron job runs without a manual operation. after it succeeds we'll just update the code and revert it to the intended state of triggering for the last date

hasura/metadata/cron_triggers.yaml
```diff
@@ -54,6 +54,20 @@
...
+- name: Rollup World ID analytics
+  webhook: '{{NEXT_API_URL}}/_rollup-world-id-analytics'
+  schedule: '*/15 * * * *'
+  include_in_metadata: true
+  payload: {}
+  retry_conf:
+    num_retries: 1
+    retry_interval_seconds: 10
+    timeout_seconds: 60
+    tolerance_seconds: 900
```
Takaros999, line 66: i would increase the timeout to 14m, so the cron job can make enough progress. if we only process rows 1 minute every 15 minutes this could hours/days.

PR #2194, opened 2026-07-31
fix(portal): show an unavailable notice for external app transactions

soamdesai-tfh:
## Problem

For external apps, Mini App → Transactions always rendered the red **"Failed to load transactions"** error state — the payments endpoint has nothing to return for an app that isn't a Mini App. Permissions and Notifications already handle this case with a grey "unavailable" notice, so Transactions was the odd one out.

| Before | After |
| --- | --- |
| Red "Failed to load transactions" + Try Again | Grey "Transactions unavailable / Transactions aren't available for external apps." |

## Changes

- **New server-side query** `GetAppMode` + `getIsExternalApp` helper. The Transactions page is a server component, so it can't reuse the Apollo queries the two client pages use, and no existing server query returns `app_mode` for an app id. The check looks at `app_mode` only — never `category`, which can be "External" on a real Mini App — and prefers verified metadata over the autosaved draft, matching the Notifications gate.
- **Both portal versions gated.** V3 is what the screenshots show; the legacy page has the identical failure and `pickPortalVersion` still routes to it.
- **Icon nudge.** The three notice icons sat 2.04px below the text's optical band (cap top → last baseline), measured in the browser at text-13/120%. Corrected with a new `noticeIconClassName`, alongside the existing optical helpers in `scenes/PortalV3/common/Icon`.

## Testing

- New `web/tests/unit/pv3-mini-app-transactions.test.tsx` covers four branches: external draft, verified-beats-draft, mini apps still fetch, and a failed lookup still falls through to the normal path.
- `pnpm test:unit` (612 passed), `npx tsc --noEmit`, `pnpm format:check`, `node scripts/check-optical-centering.mjs` — all clean.
- Verified against a local dev server with a real external app: notice renders, no payments request, and each icon re-measures at 0.04px off band center.


web/scenes/Portal/Teams/TeamId/Apps/AppId/MiniApp/Transactions/page/index.tsx
```diff
@@ -127,6 +151,16 @@ export const TransactionsPage = async (props: TransactionsPageProps) => {
   const { params } = props;
   const appId = params?.appId as `app_${string}`;
 
+  // External apps have no Mini App payments, so the payments endpoint only ever
+  // errors here — show the same unavailable notice as permissions/notifications.
+  if (await getIsExternalApp(appId)) {
```
igorosip0v: The Mini App path calls `getIsUserAllowedToReadApp` twice: once in `getIsExternalApp` and again in `getTransactionData`. Since `getTransactionData` is a standalone server action, its authorization check should remain. Could the page-level flow avoid the duplicate membership lookup without weakening that protection?

kchaw2005 (approved):
lgtm

PR #2195, opened 2026-07-31
fix(portal-v3): fall back to the empty state when an app logo 404s

soamdesai-tfh:
## Problem

App logos in the v3 configuration flow rendered their `alt` text instead of an image. In the wizard the text was clipped by the circular drop zone, so it read as a positioning bug; on the review step you also got the broken-image glyph.

The cause is a truthy URL for an image that does not exist. `hasura/seeds/default/1678146259008_appSeed.sql` writes `logo_img_url = 'logo_img.png'` into `app_metadata`, `get-unverified-images` sees a non-empty column and mints a URL for it, and nothing ever checks that the object is actually in the bucket. Every call site guarded with `logoUrl && `, which only asks whether the URL is a non-empty string — so the empty state was unreachable and the failed `` stayed in the DOM showing its alt text.

## Change

A shared `useImageFallback` hook that swaps to the caller's existing empty state on the image's `error` event:

```ts
isBroken: Boolean(url) && failedUrl === url
```

Keyed on the URL rather than a boolean, so a fresh upload gets another attempt without needing a reset effect. Wired into the three places that render an app logo — `LogoDropZone`, `ReviewStep`, `LivePreview`. Each adds an import, one hook call, and `onError`.

`alt` is now empty on these images. They sit next to the app name, so they are decorative, and an empty alt means nothing paints in the window between a failed load and the error event. The live-preview `` picked up a `data-testid`, since two existing tests queried it by alt text.

## Verification

- Reproduced end to end against a local dev server: the page requested `.../unverified/app_19c72734.../logo_img.png`, it failed, and the drop zone rendered the real "Drop an image, or Browse" state with no logo `` left in the DOM.
- Confirmed the images are client-rendered only (the server HTML has zero mentions of `logo_img`), so there is no pre-hydration window where the `error` event could fire before React attaches the handler.
- `pv3-logo-broken-image.test.tsx` covers both branches: error to empty state, and a new URL retrying after a failure.
- `npx tsc --noEmit` clean, Prettier clean, 18 tests pass across the two touched suites.

## Known limit

The hook remembers one failed URL, so a byte-identical URL is suppressed without a retry. If an object appears in the bucket during the same page session it stays empty until remount. A real upload avoids this, since the new URL comes back presigned with different query params.

## Noted but not addressed

- A wrong-team URL renders a fully interactive config editor that fails only on write, because the page loads by `appId` alone while `upload-image` validates app + team + membership.
- "Error uploading image" swallows the specific GraphQL error, which is only visible in the console.

### DIFF
--- a/web/scenes/PortalV3/common/useImageFallback/index.ts
+++ b/web/scenes/PortalV3/common/useImageFallback/index.ts
@@ -1,0 +1,19 @@
+"use client";
+
+import { useState } from "react";
+
+/**
+ * Guards against broken app images. A logo URL is truthy whenever the metadata
+ * row names a file, even if that object is missing from the bucket — the
+ * signed URL then 404s and the `<img>` paints its alt text over the layout.
+ * Callers render their empty state when `isBroken`, keyed on the URL so a
+ * fresh upload gets another chance.
+ */
+export const useImageFallback = (url?: string) => {
+  const [failedUrl, setFailedUrl] = useState<string>();
+
+  return {
+    isBroken: Boolean(url) && failedUrl === url,
+    onError: () => setFailedUrl(url),
+  };
+};
--- a/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/Configuration/LivePreview/LivePreview.tsx
+++ b/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/Configuration/LivePreview/LivePreview.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { getCDNImageUrl } from "@/lib/utils";
+import { useImageFallback } from "@/scenes/PortalV3/common/useImageFallback";
 import { useAtomValue } from "jotai";
 import { useWatch } from "react-hook-form";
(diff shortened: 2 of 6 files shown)


web/scenes/PortalV3/common/useImageFallback/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import { useState } from "react";
+
+/**
+ * Guards against broken app images. A logo URL is truthy whenever the metadata
+ * row names a file, even if that object is missing from the bucket — the
+ * signed URL then 404s and the `<img>` paints its alt text over the layout.
+ * Callers render their empty state when `isBroken`, keyed on the URL so a
+ * fresh upload gets another chance.
+ */
+export const useImageFallback = (url?: string) => {
```
igorosip0v, line 12: As an alternative, I can suggest implementing some `ImageWithFallback` component that will handle onError internally and apply some default fallback or accept it via `fallback` prop.

kchaw2005 (approved):
lgtm

PR #2196, opened 2026-07-31
feat(portal-v3): move the app danger zone into a configuration dropdown

soamdesai-tfh:
## What

Deleting an app was a top-level sidebar destination, which gave a destructive action the same weight as everyday navigation. It now lives as a collapsed **Danger zone** disclosure under **Advanced settings** in the configuration wizard's Basic information step, beside the other app-level settings.

## Changes

- **New `DangerZoneDisclosure`** — collapsed `<details>` row that expands to the existing delete card, matching the `<details>`/`<summary>` pattern already used by the action `SettingsCard`.
- **`SidebarNav`** — Danger zone item removed, along with the `FetchApps` query that existed only to gate it (one less client query per shell render). Configuration stays highlighted on the danger path.
- **`DangerZoneSection`** — dropped the "You'll be asked to confirm before anything is deleted." line and the red footer band; Delete app moved to the left at the card's 24px inset. `appId` relaxed to `string` since the wizard passes a plain string.
- **`DeleteModal`** — confirm by typing `Delete` rather than the full app name, matching [`DeleteTeamDialog`](https://github.com/worldcoin/developer-portal/blob/main/web/scenes/PortalV3/common/DeleteTeamDialog/index.tsx). Accepted case-insensitively so `DELETE` works too; the schema no longer depends on props, so it moved to module scope.

The standalone `/configuration/danger` route still renders — it is just no longer linked from V3 nav, since the old portal shares that route.

## Testing

- New `pv3-danger-zone-disclosure.test.tsx`: collapsed by default, and the owner-permission gate.
- `pv3-sidebar-section-nav.test.tsx` updated to assert the sidebar no longer carries a Danger zone link.
- Full unit suite: 568 passed / 124 suites. `tsc --noEmit` and `prettier --check` clean.
- Verified in a local dev server: collapsed row under Advanced settings, expands to the delete card, and the confirm dialog asks for `Delete`.

### DIFF
--- a/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/Configuration/Danger/DangerZoneDisclosure.tsx
+++ b/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/Configuration/Danger/DangerZoneDisclosure.tsx
@@ -1,0 +1,34 @@
+"use client";
+
+import { Icon } from "@/scenes/PortalV3/common/Icon";
+import { DangerZoneSection } from "./DangerZoneSection";
+
+/**
+ * Danger zone as a collapsed row inside the wizard's Advanced settings, so
+ * deleting an app takes a deliberate expand instead of a sidebar destination.
+ */
+export const DangerZoneDisclosure = (props: {
+  appId: string;
+  teamId: string;
+  appName?: string;
+}) => (
+  <details className="group">
+    <summary className="flex w-fit cursor-pointer items-center gap-2 select-none">
+      <Icon
+        name="chevron-down"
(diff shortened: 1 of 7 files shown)


Takaros999:
i'm not sure if this is the right approach, we had this discussion with  @kchaw2005 thoughts?

soamdesai-tfh:
> i'm not sure if this is the right approach, we had this discussion with @kchaw2005 thoughts?

After most recent configuration page changes it wouldn't make sense for it to no belong right under or in the advanced settings. They go hand in hand

Takaros999 (approved):

kchaw2005 (approved):
lgtm

PR #2201, opened 2026-07-31
fix(portal): persist latest team on switch

Takaros999:
This PR fixes latest-team persistence when switching teams from the Portal V3 dropdown.

- Moves latest-team recording from the cached portal shell into the dynamic `[teamId]` layout, so client-side team navigation records the selection.
- Preserves server-side membership validation and non-blocking Redis writes while leaving Portal V2 behavior unchanged.
- Adds route and membership-guard regression coverage.

Checks:
- `pnpm exec jest tests/unit --watchman=false --runInBand` (130 suites, 584 tests)
- `pnpm exec tsc --noEmit`
- `pnpm format:check`
- `git diff --check`


kchaw2005 (approved):
ok

PR #2202, opened 2026-07-31
fix(portal-v3): prevent image upload preview refresh

kchaw2005:
## Summary

Reworks Portal V3 image uploads around a shared optimistic transaction.

- Show a local blob preview immediately, then replace it with the signed S3 URL after metadata persistence succeeds.
- Patch React Hook Form and the locale-specific Apollo image cache locally instead of refetching images or triggering the whole-form autosave.
- Make the live preview read the selected locale's image cache entry, so localized showcase uploads update immediately.
- Add rollback, append, unmount, and locale-aware preview coverage.

## Root cause

Showcase uploads waited for a separate image-query refetch before updating the form and preview. That introduced loading transitions and left the form, image cache, global state, and preview briefly out of sync.

## Validation

- `npx jest tests/unit/use-app-image-upload.test.tsx tests/unit/pv3-image-upload-preview.test.tsx --runInBand`
- `npx tsc --noEmit`
- `pnpm format:check`


kchaw2005:
@codex review

kchaw2005:
@codex review

PR #2216, opened 2026-08-03
fix(world-id): unblock the actions grid while registration is pending

soamdesai-tfh:
## Problem

After registering a relying party, the Actions section renders a grid with no create tile (`canCreate` requires `status='registered'`, the row is `pending`), no actions, and no explanation — a blank page. Nothing on that page ever re-checks the status: the rp-status reconciler is mounted only by `RpSummary` on the Configuration section, so the grid stays blank until a window refocus. #2214 keeps the registration flow itself on Configuration, but any arrival at Actions during the pending window (sidebar item, bare `/world-id` re-entry) still hits the dead end.

Two adjacent cases render the same blank grid and never recover: a member (non-admin) viewing a registered app with zero actions, and a search query matching nothing.

## Change

- **Poll the reconciler from the layout** while an RP is pending and the user is off Configuration. Same endpoint, cadence (5s, skipped while hidden), timeout (4s), and hidden-tab behavior as `useRpRegistrationController`. When production status changes, refetch the overview — the grid flips to the normal Actions page with the create tile, no user action needed. The effect unmounts once the status leaves `pending`, so there is no steady-state polling.
- **Give the grid an empty state** instead of rendering nothing: "Finish registering your relying party to create actions." (owner, pending), "Ask a team owner or admin to create actions." (member — mirrors the copy Configuration already uses at `RegisterRpEmptyState`), "No actions match your search." (search miss). When the create tile is shown it remains the only empty state.

Registration flow, tab defaults, and sidebar behavior are unchanged from main.

## Failure handling

- Reconciliation fetch failures are swallowed and retried on the next tick; the page keeps the last known status (same posture as the existing controller).
- The poll is bounded: single in-flight guard via effect lifecycle, 4s abort, stops on status change, tab change, or unmount.

## Testing

- `tests/unit/pv3-world-id-layout-freshness.test.tsx`: pending on Actions triggers reconciliation and refetches on status change; no refetch while still pending; no layout poll on Configuration (RpSummary owns it there).
- `tests/unit/pv3-world-id-action-card.test.tsx`: empty-reason message, search-miss message, and create-tile-only when creation is allowed.
- 674/674 unit tests, `tsc`, Prettier clean.

Fixes this:
image

### DIFF
--- a/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/WorldId/layout/index.tsx
+++ b/web/scenes/PortalV3/Teams/TeamId/Apps/AppId/WorldId/layout/index.tsx
@@ -273,6 +273,41 @@ export const WorldIdLayout = (props: {
     };
   }, [props.appId, refetchOverview]);
 
+  const rpId = rp?.rp_id;
+  const rpServerStatus = rp?.status;
+  // The rp-status endpoint is what reconciles a pending registration (it reads
+  // on-chain state and syncs the DB row), but its usual host — RpSummary —
+  // only mounts on the Configuration section. Keep pending converging while
+  // the user sits on Actions so the grid unlocks without a manual refresh.
+  useEffect(() => {
+    if (!rpId || activeTab === WORLD_ID_TABS.Configuration) return;
+    if (effectiveRpStatus !== RpRegistrationStatus.Pending) return;
+
+    let cancelled = false;
+    const reconcile = async () => {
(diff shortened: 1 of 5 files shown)


Takaros999 (commented):
meta point: can we ignore rp registation compeltely? and perhaps show a info banner to that you need to setup your signing key?

Takaros999 (approved):

PR #2221, opened 2026-08-03
fix(world-id): stop trusting foreign on-chain RP registrations (H1 #3910854)

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Fixes H1 #3910854 — and with it the still-open SEC-2787 / H1 #3869823, whose fix in #2141 was closed unmerged, so `main` never received it. Because `rp_id` is `uint64(keccak256(app_id))` over a *public* `app_id` and on-chain `register()` is permissionless, zero-fee and first-come, anyone can claim any not-yet-migrated app's rp_id — and the Portal then treated that foreign registration as its own: both `/api/v4/rp-status` and the MCP `get_world_id_registration_status` tool promoted a managed RP to `registered` from on-chain `initialized && active` alone (a signer check each file's own *staging* branch already performed), so a squatter's row flipped to `registered` and `proof-context` served the legitimate app's verified branding bound to an attacker-controlled OPRF signer. Both paths now share a `canTrustOnChainSigner` helper, preserve the stored status on mismatch, gate the DB writeback on the trust flag, and fail a managed row left wedged `pending` by an untrusted registration once the UserOp validity window elapses so the retry path stays reachable instead of polling forever. Three supporting gaps in the same attack surface: managed registration now pre-reads the registry and returns `rp_id_taken` rather than spending a KMS manager key on a UserOp the contract rejects with `IdAlreadyInUse` (deliberately non-fatal on RPC failure — the security boundary is the reconciliation, not this convenience check); `rp-retry` compares the on-chain *manager* before submitting a rotation that always reverts yet still wrote `status: pending`, wedging the row and burning gas per click; and `register_rp` / `rotate_signer_key` reject the zero-address signer that `isAddress` lets through, via a shared `isZeroAddress` that also replaces the local copy in `switch-to-self-managed`.

Two things worth flagging for review: the signer-mismatch warning is Redis-throttled to one per rp_id per 10 minutes (fail-open) because rp-status caches pending statuses for 1s and would otherwise emit thousands of identical lines during a live takeover, and this change prevents future poisoning but does **not** auto-heal a row already at `registered` from a prior exploit (mismatch *preserves* the stored status, matching the staging path) — that needs an operational audit of managed rows whose on-chain signer differs from `signer_address`. Salting the rp_id derivation and the contract's missing caller binding / reclaim path are out of scope here (`world-id-protocol`).

## Checklist


Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

web/api/helpers/rp-utils.ts
```diff
@@ -65,6 +73,177 @@ export function normalizeAddress(address: string): string {
...
+   */
+  mode: unknown;
+  onChainManager: string;
+  onChainSigner: string;
+  expectedSigner: string | null | undefined;
+  /** Portal's manager address; `null` when it could not be resolved. */
+  expectedManager: string | null | undefined;
+}): OnChainTrust {
+  if (mode === "self_managed") {
+    return "trusted";
```
igorosip0v, line 156: @Gr1dlock  `self_managed` always returns `trusted`, so Portal accepts whoever registered the public `rp_id` first. Could we store the expected public `manager` and `signer` before creating or switching to self-managed, then use the same check as for managed RPs?

Takaros999 (approved):

PR #2222, opened 2026-08-03
fix(auth): require a same-origin POST to delete an account (H1 #3906522)

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

`/api/auth/delete-account` was mounted as **GET** (`app/api/auth/delete-account/route.ts`) and gated solely on the presence of an Auth0 session cookie before calling the irreversible `managementClient.users.delete()`; that cookie is `SameSite=Lax` (SDK v4 default at `abstract-session-store.js:14` — `lib/auth0.ts` sets no `session.cookie` override), and Lax cookies ride **cross-site top-level navigations**, so any attacker page could destroy a logged-in developer's Auth0 identity with a single link, `window.open` or meta-refresh (**HackerOne #3906522**) — the legitimate UI flow was itself a top-level GET navigation, so the attack was just the real flow triggered from someone else's page. `proxy.ts:370` returns `authRes` for all of `/api/auth/*` before any guard and no edge control fronts the host, so the handler is the only place this can be stopped.

The route is now **POST-only** (Lax never sends the cookie on a cross-site POST; GET returns 405, which also closes the reporter's first-party `/api/auth/login?returnTo=%2Fapi%2Fauth%2Fdelete-account` delivery variant since the post-login redirect lands as a GET) and is gated by a new `api/helpers/csrf.ts` guard that requires `Sec-Fetch-Site: same-origin`, falls back to an `Origin` allow-list, and **fails closed** when neither header is present — the guard is the layer that survives a future change to the cookie's `SameSite` attribute. It returns **204** instead of the old 307 to `/api/auth/logout`, because the caller is now a `fetch()` that would otherwise replay the POST against logout; both dialogs share a new `deleteAuth0Identity()` helper and drive the logout navigation themselves, so a failed Auth0 deletion surfaces as an error toast instead of rendering a raw JSON 500 at the user.

Blast radius of the original bug was bounded — only the Auth0 leg is deleted, the Hasura row survives (it is removed by a separate client-side mutation the attacker cannot reach) and `login-callback` re-links World ID / email-OTP users on next login, so permanent loss was confined to password-connection users plus everyone's MFA/passkey enrollments and `user_metadata`.

Verified against a running dev server, not just tests:

| request | before | after |
| --- | --- | --- |
| `GET` + `Sec-Fetch-Site: cross-site` (the exploit) | 307, identity deleted | **405** |
| `POST` + `Sec-Fetch-Site: cross-site` | — | **403** `cross_origin_request` |
| `POST` + `Origin: https://evil.example` | — | **403** |
| `POST` with no origin metadata | — | **403** (fail-closed) |
| `POST` + `Sec-Fetch-Site: same-origin` | — | **401** (reaches the auth gate; flow intact) |
| `POST` + `Origin: <APP_BASE_URL>` | — | **401** (Origin fallback works) |

Not addressed here, flagged rather than silently scoped in: the client deletes the Hasura row *before* the Auth0 identity, so a failure of the second call still leaves a partial state — this PR makes that failure visible (error toast) rather than making the pair atomic — and no step-up re-authentication was added, since that is a product decision and not required to close the CSRF.

## Checklist

### DIFF
--- a/web/api/delete-account/index.ts
+++ b/web/api/delete-account/index.ts
@@ -1,13 +1,32 @@
+import { isSameOriginRequest } from "@/api/helpers/csrf";
 import { errorResponse } from "@/api/helpers/errors";
-import { logger } from "@/lib/logger";
-import { urls } from "@/lib/urls";
 import { auth0 } from "@/lib/auth0";
+import { logger } from "@/lib/logger";
 import { ManagementClient } from "auth0";
-import { NextRequest } from "next/server";
-import { getAppUrlFromRequest } from "../helpers/utils";
+import { NextRequest, NextResponse } from "next/server";
 
+/**
+ * Irreversibly deletes the caller's Auth0 identity.
+ *
+ * Mounted on POST only, and rejects any request the browser reports as cross-site.
+ * The session cookie is `SameSite=Lax`, so a GET mount here would let any attacker
+ * page destroy a logged-in developer's identity with a single top-level navigation
+ * (`<a href>`, `window.open`, meta-refresh) — a valid session cookie alone does not
+ * prove the request came from our own UI. See `api/helpers/csrf.ts`.
+ *
+ * The caller deletes the Hasura user row first, POSTs here, then navigates to
+ * `/api/auth/logout` to clear the (still-valid, stateless) session cookie.
+ */
 export const deleteAccount = async (req: NextRequest) => {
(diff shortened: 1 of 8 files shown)


Gr1dlock:
@codex review

Takaros999 (approved):

PR #2237, opened 2026-08-05
fix(world-id): make a new rp_id unguessable before it is registered

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

Follow-up to H1 #3910854 (the trust half is #2221). `rp_id` is `uint64(keccak256(app_id))` over a **public** `app_id`, and on-chain `register()` is permissionless, zero-fee and first-come — so anyone can harvest app_ids from `/api/v2/public/apps`, compute the rp_id of every app that hasn't migrated to World ID 4.0 yet, and claim it. Gas-only, no authentication, and `registerMany` does the whole store in one transaction. The payoff is denying a competitor's (or our own) v4 onboarding, and there is no reclaim path in the contract.

This keys the derivation with a server-side secret — `HMAC(RP_ID_SALT, app_id)` — so an unregistered app's rp_id isn't computable by anyone who doesn't hold the salt.

**Why keyed-deterministic rather than a per-app random value.** The self-managed flow derives the id **twice, independently**: `getSelfManagedRegistrationInfo` shows the developer a number, the developer submits `register(uint64 rpId, ...)` from their own wallet, and then `register_rp` derives it again to insert the row. Those two derivations agree today only because the function is deterministic. A per-app random id breaks that, and getting it back needs a reservation protocol (reserve-then-display, adopt-on-complete), a dialog reorder in both portal trees, and a collision path — `ClaimRpRegistration`'s `on_conflict` targets the `app_id` unique constraint, not the `rp_id` PK, so a PK collision surfaces as a raw constraint error today. Keyed derivation buys the same unpredictability against the same attacker with none of that. It also sidesteps the open question about whether IDKit/World App assume determinism, because it stays deterministic.

Both schemes are unpredictable to someone without the secret; the tradeoff is that the salt is one shared secret rather than per-app entropy. Rotation is the mitigation and it's cheap — rotating only changes the ids **unregistered** apps will be assigned, since registered rows keep their stored value.

All three call sites that derive an id for a new registration now go through a single `resolveRpIdForNewRegistration`, so they can't drift. Drift between two copies of the same rule — prod vs staging — was the actual bug in the related H1, so this is deliberate.

**Compatibility.** `rp_id` is `rp_registration`'s primary key, immutable after insert (it isn't in the table's Hasura update permissions), and every authoritative consumer already reads the stored value instead of re-deriving — verified across `proof-context`, `v4/verify`, `rp-status`, `app-status`, `rp-retry`, `switch-to-self-managed`, `toggle-rp-active`, the MCP tools and the admin screens. So flipping the flag only affects registrations created after the flip, apps registered under either scheme keep working, and rollback is just unsetting the flag. The `rp_` + 16-lowercase-hex shape is preserved, which `parseRpId`, `isValidRpId` and the two literal `/^rp_[0-9a-f]{16}$/` admin regexes all require.

**Rollout.** Off by default behind `ENABLE_UNPREDICTABLE_RP_ID`. Enabling it without a usable salt **fails the registration** rather than falling back to the guessable derivation — a silent fallback would hand out exactly the ids the flag exists to protect.

Also: the World ID config screen no longer prints an unregistered app's future rp_id. That was publishing the one value an attacker needs, and under a keyed derivation a client component can't compute it anyway. It shows `—` until an id is actually assigned.

## Notes for review

- **Verify before enabling in production:** nothing in this repo derives rp_id client-side except the decorative `buildMockRpContext` (mock signature `0x00..1b`, never validated server-side — `v4/verify` overwrites with the DB value), but I could not inspect `@worldcoin/idkit` or World App from here. The one to check is the integrity bundle's `aud` (`api/v4/verify/integrity-bundle.ts:443`): if any client derives rp_id locally rather than reading it from `/api/v4/proof-context`, unpredictable ids would surface as an opaque 403. That's why this ships flag-off.
- Does **not** re-derive or migrate existing rows, by design.
- Defensively pre-registering the ids of apps that haven't migrated yet — the installed base this change can't help, since their app_ids are already public and predictable forever — is a separate PR.
- `CLAUDE.md` documents a `featureFlags` object in `web/lib/feature-flags/index.ts`; that file doesn't exist on main, so this follows the actual pattern in `feature-flags/portal-v3/flag.ts` (server-only module + env switch).

## Checklist

### DIFF
--- a/web/.env.example
+++ b/web/.env.example
@@ -101,11 +101,11 @@ ENABLE_UNPREDICTABLE_RP_ID=
 # SSM; treat it like a signing key. Rotating it only changes the ids that
 # unregistered apps will get, so rotation is safe.
 RP_ID_SALT=
-# The outgoing salt, for ONE drain window after a rotation. A self-managed
-# developer who was shown an id derived from the old salt has already registered
-# it on-chain; without this their registration is orphaned. Clear it once no
-# in-flight setup can still be using it.
-RP_ID_SALT_PREVIOUS=
+# NOTE on rotation: the outgoing salt is deliberately NOT kept as a fallback. An
+# attacker who learned a leaked salt could pre-register the old-salt id and have it
+# adopted, which is the squat this exists to prevent. So rotate during a quiet
+# window: a self-managed developer caught between the instructions screen and
+# completion will have their registration orphaned and need manual cleanup.
 
 # Verifier Contract
 VERIFIER_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000000
--- a/web/.env.example
+++ b/web/.env.example
@@ -90,6 +90,17 @@ RP_REGISTRY_SAFE_4337_MODULE_ADDRESS=0x75cf11467937ce3F2f357CE24ffc3DBF8fD5c226
 RP_REGISTRY_KMS_REGION=eu-west-1
 RP_REGISTRY_UPDATE_RP_TYPEHASH=0xe52c0a4424a7d2015a13748c3738793d6b1dd1774e4ee82516251372da305465
 RP_REGISTRY_DOMAIN_SEPARATOR=0x150a3a11bbb5ea6c4cfbd7cedc050d67bb55333eb8700bec241368de512607f1
+# When true, a NEW registration's rp_id is HMAC(RP_ID_SALT, app_id) instead of
+# the public uint64(keccak256(app_id)), so nobody can compute an unregistered
+# app's rp_id and claim it on-chain first. Already-registered rows keep their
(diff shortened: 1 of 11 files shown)


Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

PR #2238, opened 2026-08-05
feat(world-id): defensively claim unmigrated apps' on-chain rp_ids

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

> **Stacked on #2221** (base is `Gr1dlock/hanoi`, not `main`) because it extends that PR's on-chain collision guard. Review/merge #2221 first; this rebases onto `main` afterwards.

## Description

Salting new rp_ids (#2237) cannot protect the installed base. Every app that already exists has a public `app_id`, so its `uint64(keccak256(app_id))` rp_id is predictable **forever**, and on-chain `register()` is permissionless, zero-fee, first-come, with no reclaim path (H1 #3910854). For those apps the only defense left is to hold the id ourselves until the app is ready to use it — Takis's short-term suggestion in the thread.

`POST /api/_pre-register-rp-ids` claims the ids of a caller-supplied list of apps, registering each to the Portal's **shared** manager key with a placeholder signer.

**It creates no `rp_registration` rows.** On-chain state is the record of what we hold. Inventing rows would make the Portal assert that apps are registered when their owners never asked, and `proof-context` serves off exactly that row.

**Safeguards are the substance here,** since every claim spends L2 gas and pulls the registry's WLD fee from our Safe:

- `ENABLE_RP_ID_PRE_REGISTRATION` kill switch — off makes the endpoint inert regardless of arguments.
- **Dry-run by default.** Spending requires explicitly passing `dry_run: false`; a dry run reports exactly what a real run would submit.
- Hard 25-app ceiling per call. A larger backlog drains over repeated calls; "pre-register everything" is not one request away.
- Per-outcome counts in the response *and* the logs, so a run that skipped everything for an unexpected reason can't read as a successful sweep.
- Never claims on a failed on-chain read, and re-running is idempotent (an id already ours is skipped).
- A foreign-held id is logged at `error` — it's the single most important thing to surface, and the contract gives us no way to fix it.

## Adoption — what keeps a claimed app onboardable

Claiming an id would otherwise permanently lock the app out of registration. `submitManagedRpRegistration` already refused an rp_id taken on-chain (#2221); it now tells a squatter from our own claim by the **manager address** — the one role only our KMS key can sign for — and rotates the placeholder signer to the developer's real one instead of submitting a `register()` the contract would reject with `IdAlreadyInUse`. Adoption is pinned to the shared key regardless of `ENABLE_SHARED_KEY_RP_REGISTRATION`: a dedicated key would be a manager the contract has never seen, and every later `updateRp` would revert.

## Known gap, and why the endpoint takes an explicit app list

**Self-managed apps cannot adopt yet.** A self-managed developer registers from their own wallet, so a claimed id means their `register()` reverts. `register_rp` now fails that case loudly with `rp_id_taken` instead of inserting a row that polls `pending` forever — but the real fix is transferring the manager to the developer, and while `submitTransferManagerTransaction` exists, no flow drives it.

That's why this is an operator-driven, explicit-app-list tool rather than a blanket sweep over every unmigrated app. A blanket sweep would silently break self-managed onboarding for everyone it touched, and it's on-chain and irreversible. **Do not claim ids for apps expected to self-manage until the transfer flow lands** — that sequencing is a product call, not something this PR should decide.

## Notes for review

- I have **not run this against any environment.** Flag is off, and the first real run should be a dry run against a handful of apps.
- Reuses the existing generated `GetAppInfo` / `FetchRpRegistration` queries so no `.graphql` codegen (which needs a live Hasura) is required.
- `addressesEqual` is now exported from `rp-utils` so the ownership check can't drift into its own `.toLowerCase()` comparison that skips normalization.
- Worth deciding before a real run: which apps are worth claiming. Verified/high-traffic apps are the obvious targets; claiming the full long tail costs gas per app for ids nobody may ever want.

## ⚠️ Open findings — do not merge as-is

Eleven rounds of automated review produced 14 findings, **seven of them consecutively
in one place**: the rp_id ownership decision in `submitManagedRpRegistration`. That
decision is now ~105 lines over seven interacting inputs (production initialized /
staging initialized / staging readable / shared key configured / manager resolvable /
production claim in flight / staging claim in flight), and each round has found a
different unhandled cell of that matrix. Two are still open:

- **P2 — production read failure treated as "id is free."** If the on-chain read
  throws while pre-claims can exist and dedicated-key mode is on, the flow registers
  with a fresh dedicated manager. If the id was in fact pre-claimed, status and retry
  then compare the on-chain shared manager against the stored dedicated key and
  classify our own claim as foreign, leaving onboarding stuck until manual repair.
- **P3 — dry run does not consult the in-flight claim marker**, so it can report
  `would_claim` where the matching real run would report `skipped_claim_in_flight`.
  That weakens dry-run-as-preview, which is the safeguard operators rely on.

I stopped patching individual cells deliberately. The change that terminates this is
extracting the ownership decision into a **pure function over its inputs and
table-testing the combinations**, so correctness comes from a covered matrix rather
than from the next review round. That is a focused piece of work and it should
happen before this merges.

**It is also worth deciding whether this PR is wanted at all first.** #2237 already
makes every rp_id issued from now on unguessable; this PR only protects the existing
installed base, and only for apps an operator explicitly chooses to claim. If the
answer is "not broadly," most of the complexity above earns nothing. Self-managed
apps additionally cannot adopt a claimed id yet (see above), so claiming ids for
apps expected to self-manage is not safe until the manager-transfer flow lands.

## Checklist

### DIFF
--- a/web/.env.example
+++ b/web/.env.example
@@ -90,6 +90,16 @@ RP_REGISTRY_SAFE_4337_MODULE_ADDRESS=0x75cf11467937ce3F2f357CE24ffc3DBF8fD5c226
 RP_REGISTRY_KMS_REGION=eu-west-1
 RP_REGISTRY_UPDATE_RP_TYPEHASH=0xe52c0a4424a7d2015a13748c3738793d6b1dd1774e4ee82516251372da305465
 RP_REGISTRY_DOMAIN_SEPARATOR=0x150a3a11bbb5ea6c4cfbd7cedc050d67bb55333eb8700bec241368de512607f1
+# Kill switch for POST /api/_pre-register-rp-ids, which defensively claims the
+# on-chain rp_id of apps that haven't migrated yet so nobody can squat it. Every
+# claim spends L2 gas and the registry's WLD fee, so leave this unset unless a
+# run is actually intended. The endpoint dry-runs unless asked not to.
+ENABLE_RP_ID_PRE_REGISTRATION=
+# Placeholder signer recorded on a defensively claimed rp_id. Must be a valid
+# non-zero address that CANNOT sign — nothing should be able to verify against a
+# claimed-but-unadopted RP. Replaced with the developer's real signer when the
+# app registers for real.
+RP_ID_PRE_REGISTRATION_SIGNER=
 
 # Verifier Contract
 VERIFIER_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000000
--- a/web/api/_pre-register-rp-ids/index.ts
+++ b/web/api/_pre-register-rp-ids/index.ts
@@ -13,6 +13,7 @@ import {
   parseRpId,
 } from "@/api/helpers/rp-utils";
 import { getRpFromContract } from "@/api/helpers/temporal-rpc";
+import { USER_OP_MAX_VALIDITY_MS } from "@/api/helpers/user-operation";
 import { protectInternalEndpoint } from "@/api/helpers/utils";
 import { validateRequestSchema } from "@/api/helpers/validate-request-schema";
 import { logger } from "@/lib/logger";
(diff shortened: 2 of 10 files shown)


Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

PR #2243, opened 2026-08-06
feat(admin-dashboard): "reject" button + email template for sandbox testers 

kchaw2005:
## PR Type

- [X] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description
Creates a reject button for the users in the sandbox tab `/admin/sandbox-requests` for users that submitted a request with the wrong email. Currently the portal doesn't have a resubmit function for those requests, so sandbox rejections will **delete the users request row in the table `public.sandbox_access_request`**. this will open up the user's form for resubmission
## Checklist


kchaw2005:
@codex review

web/api/admin/sandbox-requests/[id]/reject/index.ts
```diff
@@ -0,0 +1,182 @@
...
+import { sendEmail } from "@/api/helpers/send-email";
+import { authenticateAdminRequest } from "@/lib/admin-auth";
+import { logger } from "@/lib/logger";
+import { NextRequest, NextResponse } from "next/server";
+import { getSdk } from "./graphql/delete-pending-sandbox-request.generated";
+
+const SANDBOX_REQUEST_ID_REGEX = /^sbxreq_[a-zA-Z0-9]+$/;
+const MAX_REASON_LENGTH = 1000;
+
+const isSandboxRequestId = (id: string) =>
```
Comment (Takaros999) [inline web/api/admin/sandbox-requests/[id]/reject/index.ts:10]: what do you gain out this?

Takaros999 (approved):

web/api/admin/sandbox-requests/[id]/reject/index.ts
Comment (kchaw2005) [reply web/api/admin/sandbox-requests/[id]/reject/index.ts:10]: One of those things you look over when reading the code and think "hmm, the llm wanted a max reason length. probably to incentivize concise reason reporting. sounds fine to me!" The regex is to make sure the request id is the right shape before it hits hasura.

PR #2249, opened 2026-08-10
refactor(portal): relocate v3-needed code out of the v2 tree (v2 removal, step 1/3)

soamdesai-tfh:
## What

First of three steps to remove the v2 portal (`web/scenes/Portal`) now that v3 is served to everyone. This PR is **pure relocation** — it moves the three things that live in the v2 tree but are needed by surviving code, so the follow-up flag-flattening and tree-deletion PRs touch nothing live.

1. **`InitialApp` GraphQL SDK** — `apps.gql` + `apps.generated.ts` move from the v2 Apps page into `scenes/PortalV3/Teams/TeamId/Team/page/graphql/server/` (the v3 Team page is its surviving consumer). Both importers and both jest mock paths repointed. Moved in one step because codegen enforces globally-unique operation names.
2. **MCP AppStore imports** — `api/mcp/index.ts` now imports its four AppStore modules from the PortalV3 twins. Diff-verified before repointing: `form-schema`, `field-schema`, `types`, `dataTransforms` byte-identical; `AppStoreFormTypes` adds only `LocalisationCacheRow`; `utils` adds only `CONTENT_CARD_IMAGE_NAME` handling.
3. **Public kiosk wrapper** — `scenes/Portal/Kiosk/index.tsx` moves to `scenes/PortalV3/Kiosk/index.tsx` byte-for-byte; the `/kiosk/[appId]/[actionId]` route repoints to it. Its relative `ActiveKiosk` import now resolves to the v3 copy.

## Verification

- `pnpm generate:graphql-types` idempotent at the new path (pure rename, zero regen diff)
- `tsc --noEmit` clean; unit suite 156/156 (763 tests), api suite 52/52 (669 tests); prettier clean
- Kiosk trees re-diffed at head: still identical except the two known files
- Live smoke test: `/kiosk/...` renders the QR end-to-end through the relocated wrapper + v3 `ActiveKiosk` (signed-in session)
- Found while testing: the **anonymous** kiosk has never fetched data — session-less `/api/v1/graphql` requests hit Hasura's `public` role, which has no permissions on any table, so `GetKioskAction` fails validation and the page renders header + blank. Pre-existing, unchanged by this PR; it means the signed-out risk for the upcoming flag-flattening step is cosmetic only.

## Next steps

Step 2 flattens the `portal-v3` flag (all 49 route shims + proxy/login-callback/Root, with their tests); step 3 deletes `web/scenes/Portal` and newly-dead shared code.


kchaw2005 (approved):
I'm thinking we wait until takis has routed everybody to v3 first then start cleaning up v2 files. CC https://github.com/worldcoin/developer-portal/pull/2245 @Takaros999

PR #2251, opened 2026-08-10
refactor(portal): remove the portal-v3 flag — v3 for everyone (v2 removal, step 2/3)

soamdesai-tfh:
## What

Step 2 of the v2 portal removal (step 1: #2249). The rollout is complete, so the `portal-v3` feature flag and every `pickPortalVersion` shim collapse to their v3 branch. After this PR, `web/scenes/Portal` is fully unrouted dead code — step 3 deletes it.

- **43 route files** under `app/(portal)` now render or redirect their v3 branch directly. The ~20 legacy-URL routes (`/teams`, `/teams/{id}/apps`, `/world-id-4-0`, `/actions`, `/world-id-actions`, `/api-keys`, `/danger`, `/profile/teams`, …) are kept as plain redirects with their exact previous destinations and query-forwarding — old bookmarks and deep links keep working. Five layouts that became identity wrappers (or re-exported v2) are deleted.
- **`lib/feature-flags/portal-v3` is deleted.** `LOCAL_DEV_PORTAL_V3_ENABLED` and `PORTAL_V3_EMAILS` are dead; removed from `.env.example`. Deleting them from deployment config is a separate cross-repo chore.
- The two proof-debugging routes (which served the v2 scene to everyone, with no shim) swap to the v3 scenes. The actions one is a complete implementation (verified); the SIWWI one is a placeholder in **both** trees — byte-identical `<h1>` stub, so no behavior change.

## Deliberate behavior changes (the point of this PR)

1. **Team settings authorization widens**: `proxy.ts` now allows Owner/Admin/Member on team-settings routes for everyone (previously Members needed the v3 flag). Regression-guarded in `proxy-auth-origin.test.ts`.
2. **Landing pages**: members land on `/dashboard` after login and on `/` (previously `/teams/{id}/apps` for non-flagged users); invite acceptances land on `/teams/{id}`.
3. **create-team returnTo** goes straight to `/teams/{id}` instead of chaining through the `/teams/{id}/apps/` redirect (root `tests/api` spec updated in the same PR — note that package has its own CI workflow).
[...]

PR #2253, opened 2026-08-10
chore(portal): delete the v2 portal tree and newly-dead shared code (v2 removal, step 3/3)

soamdesai-tfh:
## What

Final step of the v2 portal removal (steps 1–2: #2249, #2251). Stacked on #2251 — after step 2, nothing routes to `web/scenes/Portal`, so this is a pure deletion: **287 files, −30,516 lines, +42.**

- `web/scenes/Portal` deleted (230 files).
- 44 shared files whose last importer was the v2 tree: `LoggedUserNav`, `BottomBar`, `ActionDangerZone`, `Environment`, `InitialSteps`, 21 icons, `lib/errors.ts`, and 4 `scenes/common` GraphQL/server modules. Every candidate was re-verified against the current tree (grep for remaining importers) before deletion — the original dead-code analysis predated steps 1–2, which brought three previously-dead things back to life (`WorldTextLogo`, `OutgoingLink`, the `debugger` GraphQL pair — all kept).
- CLAUDE.md's feature-flags section rewritten to describe the real per-directory convention (the `featureFlags` registry object it described never existed); dead `additional-azure` color tokens removed.

## Deliberately NOT deleted

- `Section/`, `QuickAction/`, `ProgressBar/`, 3 icons, and 3 `scenes/common` modules — pinned by PortalV3 subtrees that are currently unreachable but look like in-flight work (the Configuration redesign with `AppStoreWizard`/`LivePreview` has dedicated unit tests). Deleting those subtrees is a product call, not v2 cleanup.
- 12 icons whose deadness predates the v2 removal — separate cleanup, kept out of this PR.

## Notable findings

- **PostHog**: `action_created` / `action_creation_failed` die with the v2 CreateActionModal — legitimately. v3 has no legacy-action creation flow (the LegacyActions tab is view-only with a deprecation notice), and instrumenting the V4 creation dialog with the legacy event names would corrupt the funnel's meaning. If dashboards track these events, they stop receiving data at this merge.
- **Zero-membership landing changed in v3** (pre-existing, documented here because the v2 test that covered it is deleted): users with no teams land on `/create-team` (was `/profile` in v2). The v3 Dashboard branch is covered by `dashboard-latest-team.test.ts`; the RootPage branch has no direct coverage.
- `action-schema.test.ts` was retargeted to the v3 `createUpdateActionSchema` (same webhook SSRF and deep-link validators); the v2 create schema has no v3 successor. The inline `createActionSchema` in `api/mcp` remains untested against any scene schema, as before.
- `use-autosave.test.tsx`: the v3 hook diverged only additively, so all 13 cases were retained and retargeted; the v3-only options were already covered by `autosave-self-persisting-fields.test.tsx`.

## Verification

`pnpm generate:graphql-types` diff-clean; no orphaned `.generated.ts` files; repo-wide grep for `scenes/Portal/` returns zero; `tsc --noEmit` clean; unit+api 1371/1371; prettier clean. Live smoke: signed-out home and the public kiosk render the v3 shell with no server or console errors.

E2e CI remains red everywhere (Qase reporter subscription, tracked separately).


kchaw2005 (commented):
nit: need to get rid of `teamApiKeys` in `urls.ts `(deprecated route)

```ts
  teamApiKeys: (params: { team_id: string }): string =>
    `/teams/${params.team_id}/api-keys`,
```

kchaw2005 (approved):
lgtm fix comment on #2251 first though

PR #2255, opened 2026-08-11
fix(ci): remove deprecated Qase integration that breaks e2e tests

soamdesai-tfh:
## Why

The Qase contract ended and was not renewed. Since then every CI run fails in the e2e job before a single test executes:

```
QaseError: Error on getting environments: Business plan subscription is required to use API.
    at TestOpsReporter.processError (.../qase-javascript-commons/dist/reporters/testops-reporter.js:654:20)
 ELIFECYCLE  Command failed with exit code 1.
```

CI sets `QASE_MODE=testops`, so the reporter phones home at startup, gets rejected, throws, and kills the Playwright process.

## What

Pure removal of the integration (140 deleted lines, 1 reformat):

- `web/playwright.config.ts` — drop the `playwright-qase-reporter` entry (list + html reporters stay)
- `web/tests/e2e/specs/app.spec.ts` — drop the `qase` import and `qase.id(2)` annotation
- `web/package.json` + `web/pnpm-lock.yaml` — remove the dependency (regenerated with pnpm 9, matching CI)
- `web/.env.test` — drop the `QASE_*` block
- `.github/workflows/ci-app-router.yml` — drop the two `QASE_*` env lines from the e2e job

## Verified

- `npx tsc --noEmit` clean
- `pnpm format:check` clean
- `npx playwright test --list` discovers the suite (config + spec load without the package)
- zero `qase` matches left repo-wide

## Follow-up (repo settings, not in this diff)

The Actions variable `QASE_MODE` and secret `QASE_TESTOPS_API_TOKEN` are now unused and can be deleted from repo settings whenever convenient.


kchaw2005 (approved):
gg

PR #2257, opened 2026-08-11
chore(portal): unlink Discord from the portal

soamdesai-tfh:
## Why

The World Discord server was compromised. Until it's safe again, the portal shouldn't be sending developers there.

## What

Removes the two places the portal links to Discord:

- **Help center menu** — the "Join our Discord" row under Community support
- **Onboarding footer** — the Discord entry in the social links row

Also drops `profile-menu-discord` from the user menu's icon preload list, so it stops fetching an icon nothing renders.

## Reversible on purpose

`DISCORD_URL` (with a comment noting why it's unlinked) and `profile-menu-discord.svg` both stay in the tree. Restoring the links means re-adding the `<HelpLink>` block, the footer entry, and the preload string — no constant or asset to resurrect.

## Verification

- `tsc --noEmit` clean, prettier clean
- Dev server on the onboarding page: zero Discord anchors in the DOM, social row reads X → Github → WhatsApp → Telegram → YouTube
- The Help center menu sits behind Auth0 and I wasn't signed in locally, so that change is verified by code review and type check only

PR #2258, opened 2026-08-11
chore(portal): drop the DISCORD_URL comment

soamdesai-tfh:
Follow-up to #2257, addressing @Takaros999's review comment on `web/lib/constants.ts` ("no need for this").

Removes the comment above `DISCORD_URL`. The constant and the `profile-menu-discord.svg` asset stay put — the reason they're kept is recorded in #2257 rather than in the source.

One deleted line; `tsc --noEmit` and prettier both clean.


rohseh303 (approved):
following discord unlink pr

PR #2262, opened 2026-08-17
fix(oidc): canonicalize proof for the authorize anti-replay lock

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

The OIDC authorize replay lock derived its Redis key from a hash of the raw `proof` request string, but `decodeProof` deliberately accepts several byte-distinct encodings that all collapse to the same proof (ABI blob in either hex case or with trailing padding, nested JSON, escaped JSON, decimal or zero-padded leaves, embedded whitespace) — so re-submitting an already-used proof under a different encoding produced a different key and the "This proof has already been used" check never fired, which I confirmed by decoding eight variants of one proof into eight distinct keys. This keys the lock off a new `canonicalizeProof` helper instead (decode, then `toBeHex(BigInt(leaf), 32)` per leaf), mirroring the existing `canonicalizeNullifierHash` fix for the sibling report on this same endpoint, and replaces the GET-then-SET with a single atomic `SET NX` so two concurrent submissions of one proof can no longer both observe an empty key. Canonicalizing also rejects undecodable proofs at the door instead of forwarding them to the sequencer, which the report separately flagged as missing `proof` format validation; no valid proof can be rejected by this, since a real proof's leaves are field elements well under 32 bytes and it has to be decodable to be verifiable at all.

One rollout note: the lock key format changes, so during the rolling deploy old- and new-format keys coexist and a proof used immediately before the deploy could be replayed once after it — a bounded one-time artifact of a 90-minute ephemeral key, with no migration worth writing for it. Fixes HackerOne #3944987.

## Checklist


Gr1dlock:
@codex review

Gr1dlock:
@codex review

web/api/v1/oidc/authorize/index.ts
```diff
@@ -234,9 +253,14 @@ export async function POST(req: NextRequest) {
...
     }
     const proofKey = `oidc:proof:${hashedProof}`;
-    const isProofReplayed = await redis.get(proofKey);
 
-    if (isProofReplayed) {
+    // Claim the proof early to prevent duplicate submissions while we process
+    // it. SET NX makes the check and the claim one atomic round trip — a
+    // GET-then-SET lets two concurrent submissions of the same proof both read
+    // an empty key and both proceed.
+    const proofLockAcquired = await redis.set(proofKey, "1", "EX", 5400, "NX");
```
soamdesai-tfh, line 261: nit: we store the fingerprint for 90 mins in redis but login flow rejects reference data that is older than 60 mins so stored data after 60 mins wouldn't matter

soamdesai-tfh (approved):

web/api/helpers/verify.ts
```diff
@@ -363,6 +363,27 @@ export const encodeNullifierForStorage = (nullifierHash: string): string => {
...
+ * escaped JSON, decimal or zero-padded leaves, embedded whitespace — so
+ * anything that identifies a proof (an anti-replay lock, a dedup key) must key
+ * off this value. Keying off the raw request string instead gives one distinct
+ * key per encoding, which defeats the check entirely.
+ *
+ * Throws when the proof cannot be decoded, or when a leaf is not a number or
+ * does not fit in 32 bytes — all cases the sequencer would reject anyway, so
+ * callers should surface them as an invalid-proof 400.
+ */
+export const canonicalizeProof = (proof: string): string =>
```
soamdesai-tfh: The sequencer accepts compressed proofs padded with four zero words and decompresses them before verification. The compressed and full versions of the same proof get different Redis keys. Switching representations therefore leaves a replay bypass.

PR #2263, opened 2026-08-17
fix: require explicit consent to accept a team invite (H1 #3943242)

Gr1dlock:
## PR Type

- [ ] Regular Task
- [x] Bug Fix
- [ ] QA Tests

## Description

`/api/auth/login-callback` consumed a single-use team invite on a **GET** gated only on the `SameSite=Lax` Auth0 session cookie, so one cross-site top-level navigation force-joined a signed-in developer into an attacker's team — and the `user` role's `self OR teammate` select filter then hands that team's owner the victim's row, including their email and stable `world_id_nullifier` ([HackerOne #3943242](https://hackerone.com/reports/3943242)). That handler cannot demand `Sec-Fetch-Site: same-origin` (it is legitimately reached as the tail of the Auth0 redirect chain, so the browser reports `cross-site` on the happy path too), and nothing carried in `returnTo` is trustworthy either since `/api/auth/login` is itself an attacker-drivable GET — an attacker can drive navigations but not clicks, so explicit consent is the only signal separating a user clicking their invite email from an attacker linking them. `login-callback` therefore no longer reads or consumes the invite and just forwards `invite_id` to `/join-callback` (which also removes the no-account force-logout that any bogus `invite_id` used to trigger), and `/join-callback` becomes a consent screen naming the team — it previously auto-POSTed from a mount effect, which left the endpoint's guard useless because our own page fired the request. `POST /api/join-callback` is now the single invite-consuming path: same-origin guarded via `api/helpers/csrf.ts`, extended to accept invites for users who already have a Hasura row, and sharing a new `api/helpers/find-hasura-user.ts` with `login-callback` so both resolve "does this identity have a user row?" identically; two adjacent hardenings ride along, dropping `auth0Id` from the `user` role's select columns (Hasura cannot vary columns per filter branch, so everything granted there is teammate-visible, and nothing in that path reads it) and stopping `accept_team_invite` from inserting a duplicate membership for someone already in the team.

**Trade-off accepted:** accepting an invite now costs one extra click; the one-click alternative (a signed `SameSite=Lax` intent cookie minted by the `/join` click) also holds but adds a silent-failure mode, since an invite opened in one browser and finished in another — or a TTL expiring mid-OTP — would drop the invite.

**Verified:** 132/132 integration (real Hasura+PG), 725/725 unit, 678/678 api, plus `tsc`, `prettier` and `eslint`; new regression tests assert the invite is *not* consumed by `login-callback` (email and World ID paths), that the endpoint rejects cross-site/same-site/no-signal/foreign-Origin requests without touching the invite, and that the consent screen posts nothing on mount.

## Checklist


Gr1dlock:
@codex review

Gr1dlock:
@codex review

PR #2268, opened 2026-08-20
fix(verify): report v3 environment mismatch

Takaros999:
Motivation: When calling the `/v4/verify` endpoint, it's very easy for RPs to use a production proof with `environment: staging` and vice versa. This results to an `invalid_merkle_root` error, even though the proof is valid. 

Solution: If we detect the RP used a valid proof but for the wrong environment, we return `environment_mismatch` with a detailed error message to help debug.


web/api/helpers/verify.ts
```diff
@@ -405,6 +407,54 @@ export const verifyProof = async (
...
+            {
+              method: "POST",
+              headers: {
+                "Content-Type": "application/json",
+              },
+              body,
+            },
+          );
+
+          if (oppositeResponse.ok && (await oppositeResponse.json()).valid) {
```
igorosip0v, line 436: Should we use `valid === true` here, same as the success path?

```suggestion
          if (oppositeResponse.ok && (await oppositeResponse.json()).valid === true) {

igorosip0v (approved):
lgtm, please take a look at the codex comment

PR #2275, opened 2026-08-25
fix(sandbox ios enrollment 1/4): tighten Android request permissions and admin display

kchaw2005:
## What

- Tighten Android sandbox request service permissions for safe insert-or-ignore behavior.
- Remove unnecessary user identity display from the Android admin request table.
- Add focused metadata assertions for the permission boundary.

## Why
Currently the android requests table shows user email instead of the unique user ID even when the fetch returns a value. Since portal_email is already a column in the table, this became redundant. the PR removes the user id field from the table entirely


kchaw2005:
@codex review

igorosip0v (approved):

PR #2292, opened 2026-08-31
feat(developer portal): store analytics in redis

kchaw2005:
## PR Type

- [X] Regular Task
- [ ] Bug Fix
- [ ] QA Tests

## Description

- Implements a redis store `redis-store.ts` that provisions versioned keys for the metadata of the newest table parsed from the s3 bucket and the analytics themselves, keyed by app id

Two types of keys
- selfie-check-analytics:v1:total:<snapshotUID>:<app_id>
- selfie-check-analytics:v1:daily:<snapshotUID>:<app_id>

`snapshotUID` is a hash of the source metadata of the newest s3 object, specifically the etag and unique key. this means that different csvs will always have different snapshotKeys in that part so we can tell if the data being shown is fresh or not

## Checklist

### DIFF
--- a/web/api/helpers/selfie-check-analytics/eligibility.ts
+++ b/web/api/helpers/selfie-check-analytics/eligibility.ts
@@ -1,97 +1,29 @@
 import "server-only";
 
-import { logger } from "@/lib/logger";
-
-const SELFIE_CHECK_ANALYTICS_APPS_PARAMETER =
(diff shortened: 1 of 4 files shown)


kchaw2005:
Superseded by #2293, which now contains the complete Redis publisher and runtime accessor migration in one PR.

PR #2307, opened 2026-09-04
fix(verify): bind v4 verification environment to the app (H1 #3989287)

Gr1dlock:
## Vulnerability (H1 #3989287, #3970836; earlier #3890996 / #3919444)

`POST /api/v4/verify/{app_id}` chose which World ID identity tree, verifier contract and integrity attestation service to trust from an `environment` field in the **request body**, and never reconciled it against the app:

- `web/api/v4/verify/request-schema.ts:207-210` accepts `production | staging | sandbox` from the client.
- `web/api/v4/verify/index.ts:146-149` collapses `sandbox` to `staging` and passes it straight through.
- `web/api/v4/verify/uniqueness-proof/handler.ts:75-77, 107-110` selects `VERIFIER_CONTRACT_ADDRESS_STAGING` (v4) or the staging sequencer via `is_staging` (v3) from that value; `session-proof/handler.ts:45-52` does the same for session proofs.
- `web/api/v4/verify/integrity-bundle.ts:32-44` additionally swaps the attestation issuer/JWKS to `attestation.worldcoin.dev` — so the same field also downgrades the device-integrity trust anchor that Self Check 4.0 depends on.

The endpoint is unauthenticated and internet-reachable, and the integration the portal itself documents forwards the wallet result verbatim (`scenes/.../TryAction/CodeBlock/index.tsx`: `body: JSON.stringify(result)`), so the value is controlled by whoever produced the proof. A staging identity is freely mintable from the simulator, so `environment: "staging"` returned `200 success=true` with a fresh, distinct nullifier against a production RP — defeating one-person-one-nullifier. The v2 endpoint derives the flag server-side from the app row (`api/v2/verify/index.ts:212`), which is the intended behaviour.

## Fix

Staging verification is no longer authorized by anything the request asserts on its own. Two independent conditions must hold, and the client controls neither:

1. **The app's own team opened a time-boxed staging window out of band.** New nullable `rp_registration.staging_verification_expires_at` (migration `1788480000000_add_staging_verification_window_to_rp_registration`, plus Hasura column permissions). Null, expired, or unparseable all mean production-only — the decision fails towards production.
2. **The request carries the token that opening the window issued.** New `rp_registration.staging_verification_token_hash` holds an HMAC (the existing `generateHashedSecret`/`verifyHashedSecret` pair used for API keys); the plaintext is shown once at issuance and sent back as the `x-staging-verification-token` header. It is not consumed per request — it stays valid for the lifetime of its window, and re-opening a window strands the previous token. Closing the window clears the hash, so the token dies with it.

The gate lives in `web/api/v4/verify/staging-access.ts` and is called from `api/v4/verify/index.ts` right after the environment is resolved, so it covers uniqueness proofs (v3 and v4), session proofs, and integrity-bundle verification before any of them trusts the value. Refusals are `403 environment_not_allowed`, with a detail naming the remedy.

New MCP tool `set_world_id_staging_verification { app_id, enabled }` opens a 24h window and issues the token, or closes the window immediately. MCP authenticates with a team API key and `requireApp` scopes it to that team's app, so only the developer can open a window — and the switch is nowhere near verification traffic.

**Why each half is needed.** The window alone is RP-wide: while a developer tested, anyone could send a simulator proof to that app's *production* endpoint, so the guarantee stayed bypassable exactly during the test session. The token alone is a standing credential, which a backend can end up attaching to production traffic while forwarding its user's `environment` — the confused deputy this endpoint started with (and the reason the first commit's API-key gate was dropped). Together, a staging verification has to be both currently sanctioned and individually authorized.

**Why not simply force production:** every World ID 4.0 RP belongs to a production app (`register_rp` rejects staging apps) and is deliberately duplicated onto the staging RP registry (`api/helpers/rp-registration-flows.ts:277-310`) so its developer can test against the simulator. That flow has to keep working; only the authority to invoke it moves out of the request.

## Compatibility impact

- Production traffic: unaffected — no `environment`, or `environment: "production"`, behaves exactly as before, with no authentication and no new headers.
- Existing sandbox/staging users: staging verification returns `403 environment_not_allowed` until the developer opens a window and adds the issued token as a header on their (server-side) verify calls, and again once the 24h window lapses. No portal UI ships here; a Configuration-page control over the same columns is the obvious follow-up.
- Docs (docs.world.org, separate repo) need the window + header described alongside the `environment` field.

## Rollout / rollback

- Migration is `ADD COLUMN ... NULL` for both columns with no defaults — metadata-only in Postgres, no table rewrite, no lock held on a hot table.
- Deploy order is safe either way: existing code ignores the columns, and new code treats a missing window or hash as closed. Rollback is a straight revert; the columns can be left in place (nothing else reads them) or dropped with the down migration.
- Impact is observable immediately: refusals go through `errorResponse`, which warns with `error.code:environment_not_allowed` and `app_id` — facet on it to see which integrations relied on unauthenticated staging. Window changes are audited via `logPortalEvent` as `portal_staging_verification_window` with `rp_id`, `enabled` and `expires_at`; the token itself is never logged.
- No new per-request I/O: both columns come from the RP registration the handler already fetches, and the token check is an in-process HMAC.

## Tests

`web/tests/api/v4/verify.test.ts` (handler-level, mocking only the I/O boundary — RP resolution, integrity bundle, proof handlers):

- refuses staging when the app has never opened a window (integrity bundle and proof handler never run)
- refuses sandbox once the window has expired, token or not
- refuses staging inside an open window with no token, with a token that does not match, and when no token was ever issued
- verifies a uniqueness proof, and a **session** proof, with an open window and its token
- refuses a staging session proof outside a window
- verifies production proofs whether or not a window is open
- `isStagingVerificationOpen` boundaries: never opened, cleared, unparseable, non-string, one second expired, and the exact expiry instant

Plus the pre-existing `sandbox → staging` normalization test, now run inside an open window.

`web/tests/api/mcp.test.ts`: walks `tools/list` and asserts every advertised tool dispatches through `tools/call`. The two dispatch switches in the MCP route are not exhaustiveness-checked, so a tool can compile while being unreachable — which is exactly what happened to `set_world_id_staging_verification` in the second commit here, and would have left the new gate refusing every staging proof with no way to open a window.

`web/tests/api/v4/verify/staging-access.test.ts` exercises the gate against the **real** HMAC helper (the handler suite mocks `verifyHashedSecret`, so it only proves the branch wiring): the issued token authorizes, while a token bound to another RP, a superseded token, a truncated token, and a valid token after the window closed are all refused.

`npx tsc --noEmit`, `npx jest tests/api` (870 passed, 1 skipped), and `pnpm format:check` are clean. The GraphQL schema/type regeneration was done against a local Hasura with this migration applied and reduced to the `rp_registration` types only, so `graphql.schema.json` carries no unrelated churn — the splice asserts that every difference outside the `rp_registration*` types is empty.

The MCP tests also assert the token is returned once, that only its HMAC reaches the database, and that closing a window writes `null` to both columns. They also cover the case where the window mutation matches no row, on both enable and disable — no token is handed out and no audit event is written.

## Where the open review threads stand

The automated review comments on earlier commits describe designs that were replaced rather than patched, so they read as unaddressed. For the reviewer's benefit:

| Thread on | Finding | Addressed by |
| --- | --- | --- |
| `245d92e` `api/v4/verify/index.ts` | An app API key on the verify call is a confused deputy for a backend that forwards its user's `environment` | `e8c29cf` — authorization moved to server state (the window); the API-key gate was removed entirely |
| `e8c29cf` `api/mcp/index.ts` | The new tool was advertised but missing from both dispatch switches, so no one could open a window | `2b6e5f3` — both switches, plus a test that walks `tools/list` and asserts every advertised tool dispatches |
| `2b6e5f3` `tests/api/mcp.test.ts` | That dispatch test calls each tool with `{}`, so it never reaches the handler's branches | `639090f` — enabled/disabled/no-registration branches asserted against the mutation |
| `639090f` `api/v4/verify/staging-access.ts` | An RP-wide window authorizes every caller for its duration | `943cd79` — a per-window one-time token is now required alongside the window |
| `943cd79` `api/mcp/index.ts` | The tool logged a success event and returned a one-time token even when `update_rp_registration_by_pk` matched no row — and on close it reported "staging is closed" while the window could still be open | `41e4319` — the mutation must return the RP it targeted before the event or the token is emitted; otherwise an explicit retryable error, with tests for both directions |

### Adversarial-review findings deliberately not actioned

| Finding | Why not |
| --- | --- |
| Make `verifyHashedSecret` constant-time | `web/api/helpers/utils.ts` is not part of this diff and backs 9 production auth paths (API keys, OIDC client secrets, MCP auth). The compared values are HMAC digests, and the caller controls only the preimage — it cannot steer the computed digest, so the byte-at-a-time oracle that makes `===` dangerous does not apply. Worth doing as its own hygiene change, not inside this fix. |
| Mixed-version rolling deploys leave old pods ungated | True of every security fix. The change is additive and forward-compatible in both directions (old code ignores the columns, new code treats absent state as closed), so the exposure window is just the deploy, and demanding a no-mixed-version cutover would conflict with the rolling-deploy requirement. |
| The Portal kiosk cannot supply the new header | Its `/api/v4/verify` branch is unreachable: both callers of `submitKioskProof` hardcode `is_v4_action=false` (`Settings/page/index.tsx:57`, `Kiosk/ActiveKiosk/index.tsx:123`). Even if reached, `getKioskEnvironment` only returns `staging` for `app_staging_*` ids, which `api/hasura/register-rp/index.ts:144-151` refuses to migrate to World ID 4.0. |
| Consume the staging token on first use, or bind it to a proof / nonce / session | Re-raised twice by the adversarial reviewer, and declined both times. The token is a window-scoped credential by design: it travels server-to-server on the integration's own verify calls and never reaches the party the gate defends against (whoever produced the proof), the window is a short self-healing timebox, and re-opening strands the old token. Per-proof binding or atomic consumption is a different security model for this endpoint — the fourth in this PR's history — and is an escalation for product, not a bounded fix. The accepted bound is now stated explicitly in `authorizeStagingVerification`'s doc comment rather than implied. |
| Add optimistic concurrency control to the window mutation | The handler does no read-modify-write on window state (`app-context.graphql` does not select the `staging_verification_*` columns), so there is no lost update to prevent. Every racing writer is the same authorized team, "opening another window replaces it" is documented in the tool's own response, and the window is a self-healing 24h timebox that still requires the matching token. Adding OCC would change the tool's contract — a design decision rather than a bounded fix. Re-raised as open/close reordering (a delayed open committing after a close returned success): same answer — both racers are the same authorized team, the reopened window still requires the token that open issued, and it lapses on its own. |


Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

Gr1dlock:
@codex review

igorosip0v (approved):

PR #2310, opened 2026-09-04
fix(verify): bind v4 verification to recognized credential issuers (H1 #3989060)

Gr1dlock:
## Vulnerability (H1 #3989060, #3938195)

`POST /api/v4/verify/{app_id|rp_id}` never bound a World ID 4.0 proof to the credential issuer a relying party expects.

- `web/api/v4/verify/request-schema.ts` accepted any integer `issuer_schema_id` (`yup.number().integer().required()`), with no allowlist. Only the value `11` (Self Check) had special handling, for the integrity-bundle gate.
- `web/api/v4/verify/uniqueness-proof/verify-v4.ts:30` and `web/api/v4/verify/session-proof/verify-util.ts:37` forwarded that value straight into the on-chain Verifier.
- The Verifier's only issuer check is `UnregisteredIssuerSchemaId` (`web/api/helpers/temporal-rpc.ts:112`), i.e. "some registered issuer signed this". Per the protocol spec, `CredentialSchemaIssuerRegistry` registration is open to anyone who pays the one-time fee, and "establishing trust between protocol parties" is explicitly listed as future work.
- The success body (`web/api/v4/verify/uniqueness-proof/handler.ts`) returned `results[].identifier` — a caller-chosen, unvalidated free-form string on v4 (v3 pins it to a credential-type enum) — and omitted `issuer_schema_id` entirely.

Net effect: an attacker registers their own credential schema, self-issues credentials to accounts they control, calls the endpoint with `identifier: "orb"`, and the portal answers `success: true` with a fresh nullifier. A third-party RP following the published docs has nothing in the response that distinguishes this from an Orb credential. The optional integrity bundle does not help: its digest covers only nonce and Self Check claims, and the v2-bundle requirement keys off `issuer_schema_id === 11`, which an attacker simply avoids. TFH's own consumer (app-backend-main) is protected because it compares `response.issuerSchemaId` against its own enum; external RPs cannot, because the field was not returned.

## Fix

1. **Server-side allowlist** (`web/api/v4/verify/issuer-schema.ts`). Protocol 4.0 requests whose `issuer_schema_id` is outside the recognized set are rejected with `400 unrecognized_credential_issuer` and `attribute: responses[i].issuer_schema_id`, before integrity verification and before any on-chain call. The recognized set is the TFH issuer schemas World App presents today: `1` (Proof of Personhood/Orb), `11` (Self Check 4.0), `128` (legacy PoH schema still held by unmigrated clients — see `PohMigrationProcessor.LEGACY_DEFAULT_ISSUER_SCHEMA_ID` in wld-android), `9303` (NFC document) and `9310` (mobile network credential); `1`/`9303`/`9310` match `app-backend-main/src/world-id/constants.ts`.
2. **Echo the issuer back.** `UniquenessResult` and `SessionResult` now carry `issuer_schema_id` on success and on failure, so an RP can bind a result to the issuer it asked for instead of trusting the caller-chosen `identifier`.
3. **Bound `identifier`** to 256 characters on the v4 and session schemas. It stays free-form (it is the RP's own request-item label), but it is no longer an unbounded string reflected into responses and logs.

## Compatibility / breaking-change analysis

- Legitimate World App traffic is unaffected: every issuer schema the client can present is in the default set, including the legacy `128` that clients still fall back to during PoH migration.
- Protocol 3.0 requests are untouched — they carry no issuer schema, and results for them have no `issuer_schema_id`.
- The new `results[].issuer_schema_id` field is additive; existing RP parsers ignore unknown fields.
- The only traffic that breaks is traffic presenting a credential from an issuer this API does not recognize — which is the vulnerability. Environments whose registry assigns different ids (e.g. a staging/QA registry, or the debug faux issuer) set the per-environment override rather than needing a code change.
- `identifier` values longer than 256 characters are now rejected. No known client sends one; the limit is generous for a request-item label.

## Rollout and kill switch

- `V4_VERIFY_ALLOWED_ISSUER_SCHEMA_IDS` and `V4_VERIFY_ALLOWED_ISSUER_SCHEMA_IDS_STAGING` (comma-separated ints) override the built-in set per verifier environment — staging and production registries are different contracts, so the same numeric id can name a different (issuer, schema) pair in each, and a staging override must not govern production traffic. Sandbox requests resolve to the staging list, matching the verifier they hit. A list that is empty or holds an invalid id is ignored **whole** with a warning and the built-in set stands, so a typo can neither widen the allowlist nor silently narrow it by dropping one entry.
- `V4_VERIFY_ISSUER_ALLOWLIST_ENFORCED=false` downgrades the check to log-only without a code rollback. Suggested rollout: deploy enforced, watch the rejection log for a day; if a legitimate issuer id turns out to be missing, flip the switch or extend the list and add the id in a follow-up.
- Every rejection (and every log-only pass-through) emits a structured `logger.warn` — `"Unrecognized credential issuer in v4 verify request"` with `rp_id`, `app_id`, `environment`, `enforced` and the offending `issuer_schema_ids` — so the rollout is observable in Datadog and the volume tells you immediately whether the default set is wrong. Rejections are expected to be rare, so a log rather than a metric is the right shape here.

## Tests

`web/tests/api/v4/verify.test.ts` — recognized issuer verifies; self-registered issuer rejected with the new code/attribute and no downstream handler call; session proofs rejected the same way; env override accepts a configured id and rejects a previously-recognized one; a staging override does not apply to production requests; an override holding an invalid id falls back to the built-in set instead of applying partially; kill switch lets an unrecognized id through; v3 requests unaffected.

`web/tests/api/v4/verify/uniqueness-proof/verify-v4.test.ts` (new) and `web/tests/api/v4/session-proof.test.ts` — `issuer_schema_id` is echoed on success, on verifier failure, and on malformed input, with the verifier mocked at the RPC boundary.

`web/tests/api/v4/request-schema.test.ts` — identifier accepted at 256 characters, rejected at 257.

`cd web && npx tsc --noEmit`, `pnpm format:check` and the full `tests/api` suite (843 passing) are green.

### DIFF
--- a/web/.env.example
+++ b/web/.env.example
@@ -108,10 +108,12 @@ RP_REGISTRY_DOMAIN_SEPARATOR=0x150a3a11bbb5ea6c4cfbd7cedc050d67bb55333eb8700bec2
 VERIFIER_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000000
 VERIFIER_CONTRACT_ADDRESS_STAGING=0x0000000000000000000000000000000000000000
 
-# Credential issuer schema ids /api/v4/verify accepts, comma separated. Unset
-# uses the built-in list in api/v4/verify/issuer-schema.ts. Set it only for an
-# environment whose registry assigns different ids.
+# Credential issuer schema ids /api/v4/verify accepts, comma separated, per
+# verifier environment (sandbox requests use the staging list). Unset uses the
+# built-in list in api/v4/verify/issuer-schema.ts; a list holding an invalid id
+# is ignored whole and the built-in list stands.
 V4_VERIFY_ALLOWED_ISSUER_SCHEMA_IDS=
+V4_VERIFY_ALLOWED_ISSUER_SCHEMA_IDS_STAGING=
 # Kill switch: "false" downgrades the check above to log-only.
 V4_VERIFY_ISSUER_ALLOWLIST_ENFORCED=
 
--- a/web/.env.example
+++ b/web/.env.example
@@ -108,6 +108,13 @@ RP_REGISTRY_DOMAIN_SEPARATOR=0x150a3a11bbb5ea6c4cfbd7cedc050d67bb55333eb8700bec2
 VERIFIER_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000000
 VERIFIER_CONTRACT_ADDRESS_STAGING=0x0000000000000000000000000000000000000000
 
+# Credential issuer schema ids /api/v4/verify accepts, comma separated. Unset
+# uses the built-in list in api/v4/verify/issuer-schema.ts. Set it only for an
+# environment whose registry assigns different ids.
+V4_VERIFY_ALLOWED_ISSUER_SCHEMA_IDS=
+# Kill switch: "false" downgrades the check above to log-only.
+V4_VERIFY_ISSUER_ALLOWLIST_ENFORCED=
+
 # Temporal RPC (alternative to Alchemy)
 TEMPORAL_RPC_URL=http://internal-temporal-jsonrpc-757737004.us-east-1.elb.amazonaws.com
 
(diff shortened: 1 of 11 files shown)


Takaros999:
Superseded by #2322, which keeps the fix in shared request validation: supported issuer/identifier pairs, both selfie and face for schema 11, and staging-only faux issuer 128. The replacement preserves the response schema. Schema 128 is a staging faux issuer, as confirmed in WalletKit and World App iOS.

PR #2323, opened 2026-09-08
chore(portal): updating selfie check analytics page

kchaw2005:
## Changes
Updates to the selfie check analytics page:
- moves all time and daily analytics to separate tabs
- changes the color of the Unknown OS value
- Capitalizations and grammar

image

image


vincent-assoun (approved):
not a fan of the multiple tabs but let's give it a try

PR #2326, opened 2026-09-08
fix(portal): restrict RP self-managed switch to team owners

soamdesai-tfh:
## What

`switch_to_self_managed` now requires the team **OWNER** role; previously any team **ADMIN** could trigger it. The World ID configuration page disables the "Switch to self-managed" button for non-owners and explains why.

## Why

Switching an RP to self-managed submits a one-step `updateRp` transaction that transfers the RP's on-chain manager to a caller-supplied address, then clears the portal-side KMS key and signer address. Unlike the other RP controls (`rotate_signer_key`, `toggle_rp_active`), this is not recoverable by the platform: once the manager transfers, only the new manager can act on the RP. That makes it owner-tier, consistent with the existing owner-only actions (app deletion, API key reset, team role changes).

`rotate_signer_key` and `toggle_rp_active` deliberately stay ADMIN+OWNER: both remain recoverable while the portal holds the manager key.

## How

- New `CheckUserIsOwnerInApp` query (`role: { _eq: OWNER }`) used by the handler. The shared `CheckUserInApp` (ADMIN+OWNER) is intentionally untouched — `rotate-signer-key` and `toggle-rp-active` still use it.
- Unauthorized responses keep the same `unauthorized` code; the detail message now states the owner requirement.
- `RpSummary` disables the switch button for non-owners via the existing client session check (`checkUserPermissions` with `Role_Enum.Owner`), with an inline note ("Only the team owner can switch this RP to self-managed.") wired via `aria-describedby`. This is display-only — the handler enforces OWNER. All other controls keep their existing admin-tier gating.

## Testing

- New handler test: a non-owner gets 400 `unauthorized` and no state is touched (no slot claim, no transaction, no DB write, no KMS scheduling).
- New UI test: when the owner check fails, the switch button is disabled with the note while "Rotate signer key" stays enabled, and the component asks for exactly `[Role_Enum.Owner]`.
- `cd web && npx tsc --noEmit` clean; full jest unit + api suites pass (202 suites / 1593 tests); prettier clean; GraphQL codegen produced only the new generated file.

## Follow-ups (not in this PR)

- The RP action handlers do not log the acting user id; actor attribution and an owner notification would improve traceability of destructive RP operations.
- The in-app permissions list does not mention the RP operations; worth adding rows once the tiers are settled.


Takaros999 (approved):
i think we can simplify this by a lot, we have hooks to check the user's permission that should be enough for the frontend layer and then it should a be simple check

PR #2329, opened 2026-09-08
fix(ci): gate integration tests on Hasura query readiness

soamdesai-tfh:
## What

Rewrites `web/tests/verifyHasuraIsUp.js` — the gate in `test:integration` — to wait for Hasura to be **query-ready** instead of declaring ready on the first `GET /healthz` 200.

## Why

The Integration Tests job has been failing intermittently (lately, consistently on cache-warm runners) with every test in the first suites erroring `TypeError: fetch failed / SocketError: other side closed`, then all later suites passing — e.g. on #2326 it failed 4 of 5 attempts across two heads.

Root cause: the test stack runs `hasura/graphql-engine:v2.47.0.cli-migrations-v3`, whose entrypoint applies migrations/metadata and then swaps engines. `/healthz` can return 200 while the port still drops connections moments later, so the gate opens ~2s early and the first jest suites (which start in under 2s) hit the swap window. The gate was a liveness probe used as a readiness probe.

Reproduced locally with a cold boot of `docker-compose-test.yaml`:

```
HEALTHZ_OK at +14s                                            <- old gate opens here
Waiting for Hasura to serve queries (attempt 6): fetch failed (other side closed)
Waiting for Hasura to serve queries (attempt 7): fetch failed (other side closed)
✅ Hasura is query-ready (attempt 8)
QUERY_READY at +16s                                           <- new gate opens here
```

Whether CI passed depended on how long `pnpm install` ran (cold caches gave Hasura time to finish), which is why reruns flip-flopped and unrelated PRs were hit.

## How

- Readiness = a real GraphQL query (`user(limit: 1)` with the test admin secret) succeeding **twice, 1.5s apart**, which requires applied migrations plus the final engine and bridges the engine swap.
- Bounded end to end: 120s overall cap (unchanged), 2s poll interval, 2s per-request abort via `AbortSignal.timeout`, one log line per attempt, exit 1 with an actionable message pointing at the Hasura container logs.
- Against an already-ready stack it passes on attempt 1 (adds only the 1.5s confirmation delay).

No production code paths change; this file is only invoked by `pnpm test:integration`.


kchaw2005 (approved):
cautious an LLM prompted me to make this same file when integration tests started being flaky. could be unnecessary but also could be a faster check so lgtm

PR #2332, opened 2026-09-08
fix(home): keep mobile header controls on one line

soamdesai-tfh:
prevent the mobile Sign in control from wrapping


soamdesai-tfh:
Superseded by #2333. This replacement uses the same head commit from an upstream branch, allowing the E2E workflow to access its trusted development secrets.

